PR #27, opened 2019-01-16
[DNM/WIP] bsdlib: Update BSD library for nrf9160.

gbakke:
Library version d7e14c9d6bb.

Signed-off-by: Glenn Ruben Bakke <glenn.ruben.bakke@nordicsemi.no>


lemrey:
@oxalistriangularis Unfortunately, this doesn't fix the issue we observed in the MQTT library. Basically, if a socket is used in blocking mode and there is no memory available in the socket to hold a message, the socket is expected to block. As I understand this socket implementation will return an error instead. The protocol library, since it's using the sockets in blocking mode, can not be expected to handle it as non-fatal error and retry. Even if we changed that in the protocol library, it would be difficult to justify it upstream (and probably we'd have the issue in other libraries). The only two solutions are: block or propagate the error and let the application code handle the retransmission.

@rlubos correct me if I am wrong :)

lemrey:
@rlubos this can be closed.

rlubos:
bsdlib got updated in a different PR.

PR #31, opened 2019-02-27
kconfig: Enforce that BT_LL_NRFXLIB depends on nRF52

SebastianBoe:
Add a dependency from BT_LL_NRFXLIB onto SOC_SERIES_NRF52X to prevent
users from creating invalid configurations. BT_LL_NRFXLIB only runs on
nRF52.


SebastianBoe:
@rlubos : Hi, would you mind reviewing/merging this trivial patch?

PR #32, opened 2019-03-04
nfc: Fix nfc platform clock assert

KAGA164:
Fix errors from use of two clock drivers
in the same time.

### DIFF
--- a/nfc/src/nfc_platform_zephyr.c
+++ b/nfc/src/nfc_platform_zephyr.c
@@ -3,8 +3,8 @@
  *
  * SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
  */
-
-#include <nrfx_clock.h>
+#include <clock_control.h>
+#include <device.h>
 #include <nrfx_nfct.h>
 #include <nrfx_timer.h>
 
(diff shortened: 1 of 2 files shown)


KAGA164:
@grochu @pdunaj please review.

nfc/src/nfc_platform_zephyr.c
```diff
@@ -14,11 +14,17 @@
...
 {
-	nrfx_clock_init(NULL);
-	nrfx_clock_enable();
-	nrfx_clock_hfclk_start();
+	int err;
+
+	clock = device_get_binding(DT_NORDIC_NRF_CLOCK_0_LABEL "_16M");
+	__ASSERT_NO_MSG(clock);
+
+	err = clock_control_on(clock, (void *)1);
```
grochu: What does this address point to: `(void *)1`?

grochu: What does this address point to: `(void *)1`?

KAGA164: It is strangely done in the clock driver. This pointer is cast to u32_t variable. If it is 1 then CPU is sleep until the 16MHZ clock settles in other case the cpu is not sleep and this is not blocking clock start.

grochu: Ok then I guess for now we have to use this driver in this weird way, until a proper Clock driver is added in Zephyr.

KAGA164:
@pizi-nordic Could you review and merge ?

PR #33, opened 2019-03-04
nfc: lib: Handle NFC T2T valid write command

KAGA164:
NFC read only Type 2 Tag stay in active state when
receive valid write command.


KAGA164:
@grochu @pizi-nordic Could you review and merge ?

PR #35, opened 2019-03-07
nfc: lib: Return T2T to RX after write command

KAGA164:
NFC read only Type 2 Tag should return to RX state
after receive valid write command.


KAGA164:
@grochu @pizi-nordic Please review and merge.

PR #39, opened 2019-03-29
Adding nrf_oberon v3.0.0

frkv:
This pull request adds the nrf_oberon version 3.0.0
It also removes nrf_oberon version 2.0.7 (and include files)

_Note that the occ-prefix has been replaced with ocrypto-prefix. 
Also note that the order of parameters in the sha256 finalization function is changed (output, context)=>(context, output)_


gbakke:
Noticed.

lemrey:
I'll write it here so that other reviewers are aware: if the existing version is replaced, that commit will break linking of any firmware using `bsdlib`, and that hurts a bit :)

As a side note, please format the commit message and add the `signed-off` string.

oyvindronningstad:
Can we merge this? :)

oyvindronningstad:
@carlescufi @tejlmand Ready to merge.

carlescufi:
@oyvindronningstad please note that you will need to update the `west.yml` in fw-nrfconnect-nrf for this to be pulled. I assume you already have a PR for that?

oyvindronningstad:
> @oyvindronningstad please note that you will need to update the `west.yml` in fw-nrfconnect-nrf for this to be pulled. I assume you already have a PR for that?

Will update west.yml as part of this PR: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/556

PR #44, opened 2019-04-08
Nrf cc310 mbedcrypto support

tejlmand:
First version of nrf cc310 mbedcrypto library support.

This PR contains the following:
- nrf_cc310_mbedcrypto libraries for cortex-m4 and cortex-m33 in various flavors.
- Kconfig integration to  NCS
- CMake integration to  NCS
- glue layer to support selection of multiple backends


oyvindronningstad:
What is empty.txt for?

nrf_security/Kconfig
```diff
@@ -0,0 +1,468 @@
...
+
+config CC310_BACKEND
+	bool "Use cc310 backend"
+	default y
+	depends on SOC_NRF52840
+	select NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0 if VANILLA_BACKEND
+	help
+	  Enable cc310 backend
+
+config VANILLA_BACKEND
```
oyvindronningstad: "vanilla" is perhaps too idiomatic. How about either of these in place of vanilla/VANILLA?

 - mbedtls
 - software or sw
 - standard

nrf_security/Kconfig
```diff
@@ -0,0 +1,468 @@
...
+	depends on SOC_FAMILY_NRF
+	select NRFXLIB_CRYPTO
+	help
+	  Help message
+
+if NORDIC_SECURITY_BACKEND
+
+comment "Backend Selection"
+
+config CC310_BACKEND
```
oyvindronningstad: Backend selection should be a `choice` if one and only one should be selected.

nrf_security/Kconfig
```diff
@@ -0,0 +1,468 @@
...
+	prompt "Use user provided mbedtls <config.h>"
+	depends on VANILLA_BACKEND
+
+if CUSTOM_MBEDTLS_CFG
+
+config CUSTOM_MBEDTLS_CFG_FILE
+	string "Custom mbed TLS configuration file"
+	default "custom-mbedtls-config.h"
+	help
+	  Use a custom mbed TLS configuration file. The nRF Connect SDK
```
oyvindronningstad: What is "nRF Connect SDK Security"?

nrf_security/Kconfig
```diff
@@ -0,0 +1,468 @@
...
+	  adjustments, this setting can be used to provide a custom mbed TLS
+	  configuration file instead.
+
+endif # CUSTOM_MBEDTLS_CFG
+
+comment "Configuration section"
+
+menuconfig AES_CONFIGURATION
+	bool
+	prompt "AES Cipher Mode configuration"
```
oyvindronningstad: The prompt should tell what is the difference between enabling and disabling this config.

nrf_security/Kconfig
```diff
@@ -0,0 +1,477 @@
...
+	bool
+	prompt "Backend selection"
+	depends on MBEDTLS_SHA1_C && NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
+	default CC310_MBEDTLS_SHA1_C
+
+	config CC310_MBEDTLS_SHA1_C
+	    bool "cc310" if CC310_BACKEND
+
+	config VANILLA_MBEDTLS_SHA1_C
+	    bool "mbedtls vanilla" if (VANILLA_BACKEND && NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0)
```
oyvindronningstad: Why is `NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0` checked here?

nrf_security/cmake/symbol_rename.cmake
```diff
@@ -0,0 +1,66 @@
+macro(keep_config_test_glue mbedtls_config backend_name)
```
oyvindronningstad: Add explanation of the file here like in kconfig_mbedtls_configure.cmake.

nrf_security/info.txt
```diff
@@ -0,0 +1,11 @@
+
```
oyvindronningstad: What does this file do?

nrf_security/Kconfig
```diff
@@ -0,0 +1,477 @@
...
+	  Enable cc310 backend
+
+config VANILLA_BACKEND
+	bool "Use mbedtls backend"
+	default y if !(SOC_NRF52840 || SOC_NRF9160)
+	help
+	  Enable mbedtls vanilla backend
+
+# Group 0 is used to indicate when AES and ECC should use choice menu
+config NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
```
oyvindronningstad: From what I've seen in the code base, this config is used for a lot more than the comment says. I suggest giving it a more accurate name, and a slightly longer comment :). (The comment could be `help` text BTW)

nrf_security/Kconfig
frkv: You can enable both, We will explain this in documentation update (coming later)

nrf_security/Kconfig
frkv: We don't have a better term, and it will suit upcoming architectural changes later on

oyvindronningstad: "nRF Connect SDK Security Subsystem"?
"nRF Connect SDK Security Module"?
"nRF Connect SDK Security System"?
"nrf_security"?

Hmm, removing the "The" at the beginning of the sentence would also make it better :).

nrf_security/info.txt
sigvartmh: looks like an MBEDTLS config file

nrf_security/Kconfig
frkv: We have not found a better term to describe the fact that the mbedtls is in an unaltered, de-facto state. This is the reason for the somewhat idiomatic term. 

And we wanted to shy away from SW as part of the CC310 functionality is based on running some software on top if only parts of the algorithm is HW accelerated.

nrf_security/Kconfig
frkv: Agreed on removing the "The".

nrf_security/include/mbedcrypto_glue/backend.h
```diff
@@ -0,0 +1,14 @@
+#ifndef BACKEND_H
```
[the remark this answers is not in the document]
frkv: Will fix

nrf_security/Kconfig
frkv: Will fix

nrf_security/cmake/symbol_rename.cmake
frkv: Will fix

nrf_security/Kconfig
oyvindronningstad: Fair enough. Take care that it isn't interpreted as a proper noun "Vanilla".

frkv: We will amend the name to include mbedtls for clarity

nrf_security/Kconfig
```diff
@@ -0,0 +1,477 @@
+menu "Nordic Security"
+
+config NORDIC_SECURITY_BACKEND
+	bool
+	prompt "Use Nordic provided security backend"
+	depends on SOC_FAMILY_NRF
+	select NRFXLIB_CRYPTO
+	help
+	  Help message
```
oyvindronningstad: Tell me more :).

Remember to point to documentation if options will not be explained fully in the help text. Users are very likely to happen upon this while exploring Menuconfig

nrf_security/Kconfig
```diff
@@ -0,0 +1,477 @@
...
+	depends on CC310_BACKEND || VANILLA_BACKEND
+	help
+	  Enable the DHM module.
+	  MBEDTLS_DHM_C setting in mbedtls config file.
+
+endif # !NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
+
+if NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
+
+menu "DHM   - Diffie-Hellman-Merkel modules support"
```
oyvindronningstad: Should be a menuconfig?

oyvindronningstad: IMO "modules support" is superfluous.

oyvindronningstad:
Manifest update: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/636

nrf_security/Kconfig
tejlmand: Correct, that's redundant.
will remove

nrf_security/info.txt
tejlmand: this file should have been removed.
Used internally during development.
Will remove.

nrf_security/Kconfig
tejlmand: Good point.

The comment should be updated.
For future updates to security features I prefer to keep name independent of actual selected backends, as it will allow for additional backends to be part of the choice group.
However, maybe the following name would be slightly better:
`NRF_CRYPTO_MULTIPLE_BACKEND_SELECTION_0`

If name starts to contain backend knowledge, like `NRF_CRYPTO_CC310_AND_VANILLA`, then the name must also change when adding more backends.

nrf_security/Kconfig
tejlmand: Classified :p 

Will add some more text.

nrf_security/Kconfig
tejlmand: Will check if `menuconfig` will give the desired behavior here.

Will remove `modules support`

tejlmand:
> What is empty.txt for?

stray file that has been added by mistake.
Thanks for spotting.

nrf_security/Kconfig
oyvindronningstad: The reason I'm asking :)

One backend

            *** Configuration section ***
    [*]     AES Cipher Mode configuration  --->
    [*]     AEAD  - Authenticated Encryption with Associated Data configuration  --->
    [*]     DHM   - Diffie-Hellman-Merkel modules support
    [*]     ECDH  - Elliptic Curve Diffie-Hellman library support
    [*]     ECDSA - Elliptic Curve Digital Signature Algorithm support
    [*]     RSA   - Rivest–Shamir–Adleman cryptosystem
    [*]     Hash functions configuration  --->

Both backends

            *** Configuration section ***
    [*]     AES Cipher Mode configuration  --->
    [*]     AEAD  - Authenticated Encryption with Associated Data configuration  --->
            DHM   - Diffie-Hellman-Merkel modules support  --->
            ECDH  - Elliptic Curve Diffie-Hellman library support  --->
            ECDSA - Elliptic Curve Digital Signature Algorithm support  --->
    [*]     RSA   - Rivest–Shamir–Adleman cryptosystem
                Backend selection (cc310)  --->
    [*]     Hash functions configuration  --->

Whereas I'm wondering if both could look like this:

            *** Configuration section ***
    [*]     AES   - Authenticated Encryption Standard  --->
    [*]     AEAD  - Authenticated Encryption with Associated Data  --->
    [*]     DHM   - Diffie-Hellman-Merkel  --->
    [*]     ECDH  - Elliptic Curve Diffie-Hellman  --->
    [*]     ECDSA - Elliptic Curve Digital Signature Algorithm  --->
    [*]     RSA   - Rivest–Shamir–Adleman cryptosystem  --->
    [*]     SHA   - Secure Hash Algorithm  --->

tejlmand: At first, I agree with you, that last example looks nicer, but what I dislike is that it provides a false indication that, as example:
```
[*]     ECDH  - Elliptic Curve Diffie-Hellman  --->
```
give the indication that `ECDH` is enabled, which may not be the case if both backends are un-selected in the sub-menu.

Whereas the current solution does not give such indication.
Therefore it is clearer to the user that he needs to enter the sub-menu to determine if any backend is enabled.

nrf_security/Kconfig
oyvindronningstad: I still don't really know what it means, but I'm guessing it means both backends are enabled? And if you had 3 backends, you would have a config for each combination of backends?

How about `NRF_CRYPTO_BACKEND_COMBINATION_0`

or `NRF_CRYPTO_BACKEND_COMBINATION` and it has a number value like 

    NRF_CRYPTO_BACKEND_COMBINATION = BACKEND_VANILLA + (BACKEND_CC310 << 1)

nrf_security/Kconfig
tejlmand: On the renamed menu text, I agree, text should be changed. Thanks.

oyvindronningstad: > ```
> [*]     ECDH  - Elliptic Curve Diffie-Hellman  --->
> ```
> 
> give the indication that `ECDH` is enabled, which may not be the case if both backends are un-selected in the sub-menu.
> 
> 
> Whereas the current solution does not give such indication.
> 
> Therefore it is clearer to the user that he needs to enter the sub-menu to determine if any backend is enabled.

I'd find it easier to understand if you force the selection via `depends on` (force off) or `select` (force on), so you would get 

```
        *** Configuration section ***
[*]     AES   - Authenticated Encryption Standard  --->
[*]     AEAD  - Authenticated Encryption with Associated Data  --->
[*]     DHM   - Diffie-Hellman-Merkel  --->
-*-     ECDH  - Elliptic Curve Diffie-Hellman  --->
[*]     ECDSA - Elliptic Curve Digital Signature Algorithm  --->
[*]     RSA   - Rivest–Shamir–Adleman cryptosystem  --->
[*]     SHA   - Secure Hash Algorithm  --->
```

`-*-` in menuconfig means "this has been forced on" and you can type `?` to see what is forcing it.

This patch will implement it for ECDH:

```
diff --git a/nrf_security/Kconfig b/nrf_security/Kconfig
index f9c6e69..a432992 100644
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -29,6 +29,7 @@ config VANILLA_BACKEND
 # Group 0 is used to indicate when AES and ECC should use choice menu
 config NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
        bool
+       select MBEDTLS_ECDH_C
 
 # List of additional symbols to indicate cipher key sizes.
 # Currently selected based backend selection.
@@ -316,9 +317,7 @@ endmenu
 
 endif # NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
 
-if !NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
-
-config MBEDTLS_ECDH_C
+menuconfig MBEDTLS_ECDH_C
        bool
        prompt "ECDH  - Elliptic Curve Diffie-Hellman library support"
        default y
@@ -327,12 +326,8 @@ config MBEDTLS_ECDH_C
          Enable the ECDH module.
          MBEDTLS_ECDH_C setting in mbedtls config file.
 
-endif # !NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
-
+if MBEDTLS_ECDH_C
 if NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
-
-menu "ECDH  - Elliptic Curve Diffie-Hellman library support"
-
 config GLUE_MBEDTLS_ECDH_C
        bool
 
@@ -349,9 +344,8 @@ config VANILLA_MBEDTLS_ECDH_C
        help
          Enable the ECDH module from mbedtls vanilla.
          MBEDTLS_ECDH_C setting in mbedtls config file.
-endmenu
-
 endif # NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
+endif # MBEDTLS_ECDH_C
 
 if !NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0
 ```

tejlmand: I know how kconfig works.
And what you outline here is not the desired behavior.

I think you misunderstand the purpose here.

A user can enable one or more back-end.
Depending on the back-end, different functionality might be available, such examples can:
- AES-x, e.g. AES-CBC, AES-CTR, etc. ciphers
- ECDH

as examples.

Now, AES has its own sub-menu because it contains many ciphers which can be en-/disabled individually, if the given backend supports the cipher.
Now, to make it easier for user who doesn't need anything from AES, that complete group can be disabled instead of disabling each cipher in thge sub-menu individually.

Now, ECDH, is its own, and as you noticed, if only one backend is selected, then it is a simple config, en-/disable setting.
If two (or more) backends are selected that supports ECDH, then it turns into a menu where the exact backend to use can be chosen, e.g. cc310 or mbedtls vanilla.
This is equivalent to what happens in AES under e.g. AES-CTR, AES-CBC, etc. where the same selection possibility applies if multiple backends are providing the same functionality (API wise, cause key size support may differs)

So your last example of selecting `MBEDTLS_ECDH_C` when both backends are selected is quite the opposite of what we want.

nrf_security/Kconfig
tejlmand: `NRF_CRYPTO_BACKEND_COMBINATION_0` is also fine.

`NRF_CRYPTO_BACKEND_COMBINATION = BACKEND_VANILLA + (BACKEND_CC310 << 1)` 
will not achieve what is already prepared for, as it will require more logic later in config selection as not all backends support the same, thus it still requires the logic we try to avoid with this `group` symbol.

tejlmand: > And if you had 3 backends, you would have a config for each combination of backends?

Not necessarily, the group is more a combination of supported features and the backends selected that support the particular feature set.
In this case the `NRF_CRYPTO_MULTIPLE_CHOICE_GROUP_0` it indicates that multiple backends has been selected which all support AES and ECC.

sigvartmh:
Is it intended to add the `.gitmodules` file in commit 9054d8000e00f6ed34bd3ddfd197cebf6db05560 ?

tejlmand:
> Is it intended to add the `.gitmodules` file in commit [9054d80](https://github.com/NordicPlayground/nrfxlib/commit/9054d8000e00f6ed34bd3ddfd197cebf6db05560) ?

Nope. Removed now.
Was dragged along by accident when moving from anopther repo.

PR #46, opened 2019-04-09
doc: ble_controller: Document known toolchain issue

SebastianBoe:
Certain toolchains are known to segfault when linking with the
ble_controller library.

It is not known exactly which set of toolchains segfault. It is
believed to affect all NCS revisions as of april 2019.

### DIFF
--- a/ble_controller/README.rst
+++ b/ble_controller/README.rst
@@ -28,6 +28,7 @@ Each library is distributed in soft-float, softfp-float, and hard-float builds.
   * The libraries are for evaluation purposes only.
   * The libraries are not fully functional.
   * The libraries are not built for performance.
+  * Some toolchains cause a segmentation fault when linking the libraries.
 
 
 .. toctree::


ble_controller/README.rst
```diff
@@ -28,6 +28,7 @@ Each library is distributed in soft-float, softfp-float, and hard-float builds.
   * The libraries are for evaluation purposes only.
   * The libraries are not fully functional.
   * The libraries are not built for performance.
+  * Certain toolchains are known to segfault when linking with the libraries.
```
ru-fu:
```suggestion
  * Some toolchains cause a segmentation fault when linking the libraries.
```

ru-fu: Not sure about "linking the libraries" - "linking in the libraries" or "linking to the libraries" might be better, depending on what you want to say.

SebastianBoe: good catch, "linking with the libraries" does indeed communicate the wrong thing.

PR #51, opened 2019-04-23
ble_controller: Release of ble_controller v0.2.0-1.prealpha

thomasstenersen:
Added:
* Added API for fetching build revision information
* Added ``ble_controller_rand_vector_get_blocking()``
  as a blocking call to get a vector of random bytes.
* Added API to get BLE controller build revision: ``ble_controller_build_revision_get()``
* Added separate ``ble_controller_init()`` API.

Bugfixes:
* Fixed issue in HCI control flow issue that severely limited BLE throughput.

Changes:
* Moved ``fault_handler`` and ``p_clk_cfg`` from ``ble_controller_enable()`` to
  ``ble_controller_init()``.
* Changed ``ble_controller_process_SWI5_IRQ()`` to be IRQ independent. The
  generic ``ble_controller_low_prio_tasks_process()`` is used instead and SWI5
  is no longer reserved.
* Aligned naming for ble_controller configuration names.
* Made minor changes to existing API
* Improved API documentation

Signed-off-by: Thomas Stenersen <thomas.stenersen@nordicsemi.no>


ble_controller/doc/pic/Architecture_With_RTOS.svg
```diff
@@ -0,0 +1,175 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
SebastianBoe: What's going on in this picture?

"BLE Controller Library" directly talks to "HW peripherals" doesn't it? It doesn't go through the application to do so.

thomasstenersen: It requires that the application forwards IRQs etc. to the library. For example: `void ble_controller_RADIO_IRQHandler(void);`.

SebastianBoe: Right, but the image is implying that "BLE Controller Library" does not directly interface with HW peripherals, this is wrong.

ble_controller/include/ble_controller.h
```diff
@@ -0,0 +1,348 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+#include <stdint.h>
+#include "nrf.h"
+#include "nrf_errno.h"
+
+
```
joerchan: I find it a bit weird that we changed the defaults that we have always used in SoftDevices.
I'm not saying we should change them now. but I think we should get them back before we do a proper release.
Is it OK to create a jira issue to revert these? to the s132 defaults.
Master link count should be 3, event length should be 3 * 1.25 ms.

ble_controller/doc/pic/Architecture_With_RTOS.svg
joerchan: Agree with sebastian here.
Add to a jira task for the next release to fix this?

ble_controller/include/ble_controller.h
thomasstenersen: We can take an internal discussion whether or not to revert anything. The default master link count=3 is valid -- just to show off that we can do more than 1. However, 3*1.25ms event length is not long enough when using coded phy.

I don't expect us to maintain full compatibility with what has been the defaults for the SoftDevice, much of this will be controlled through Kconfig anyway (when used in NCS).

ble_controller/doc/pic/Architecture_With_RTOS.svg
thomasstenersen: The figure has been updated internally and will be added to the next release :+1:

b-gent:
I thought you were requesting a doc review? I will be opening a separate PR with doc changes next week.

PR #57, opened 2019-05-03
libbsd: bsd lib is now an imported target

tejlmand:
Making liboberon to an imported target.

To ensure the ordering of linking when using both libbsd and liboberon, the bsd lib is now setting oberon as link library interface to ensure oberon will follow during final linking and thus symbols can be correctly resolved.

Signed-off-by: Torsten Rasmussen <torsten.rasmussen@nordicsemi.no>

### DIFF
--- a/bsdlib/CMakeLists.txt
+++ b/bsdlib/CMakeLists.txt
@@ -19,10 +19,12 @@ endif()
 # lib oberon are linked in because they are used by the bsd library
 # and must therefore be placed after the bsd library on the linker's
 # command line.
-zephyr_link_libraries(
-  ${BSD_LIB_PATH}/libbsd_nrf9160_xxaa.a
-  ${IMAGE}nrfxlib_crypto
-  c
-  )
+set(BSDLIB_TARGET ${IMAGE}bsd_nrf9160_xxaa)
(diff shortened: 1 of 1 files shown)


bsdlib/CMakeLists.txt
```diff
@@ -19,10 +19,11 @@ endif()
 # lib oberon are linked in because they are used by the bsd library
 # and must therefore be placed after the bsd library on the linker's
 # command line.
-zephyr_link_libraries(
-  ${BSD_LIB_PATH}/libbsd_nrf9160_xxaa.a
-  ${IMAGE}nrfxlib_crypto
-  c
-  )
+add_library(bsd_nrf9160_xxaa STATIC IMPORTED GLOBAL)
```
hakonfam: I don't think this is multi image proof.

My proposal is to do the following:

```
set(libname ${IMAGE}bsd_nrf9160_xxaa)
add_library(${libname} STATIC IMPORTED GLOBAL)
...
```

tejlmand: good point.

Let's make it better while we're at it :)

hakonfam:
@tejlmand  This should be ready to merge?

PR #60, opened 2019-06-03
cmake: added ${IMAGE} on all target handling to support multi image

tejlmand:
To support nrf security in SPM the CMake files controlling the
crypto and mbedtls libraries no have all target references prefixed
with ${PREFIX}.

### DIFF
--- a/nrf_security/cmake/symbol_rename.cmake
+++ b/nrf_security/cmake/symbol_rename.cmake
@@ -48,27 +48,27 @@ function(library_redefine_symbols backend template)
   configure_file(${template}
                  symbol_rename_${MBEDTLS_BACKEND_PREFIX}.txt)
   
-  set(BACKEND_RENAMED_LIBARY libmbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend.a)
+  set(BACKEND_RENAMED_LIBARY lib${IMAGE}mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend.a)
   add_custom_command(
     OUTPUT ${BACKEND_RENAMED_LIBARY}
     COMMAND ${CMAKE_OBJCOPY} 
             --redefine-syms
             ${CMAKE_CURRENT_BINARY_DIR}/symbol_rename_${MBEDTLS_BACKEND_PREFIX}.txt
-            $<TARGET_FILE:mbedcrypto_${MBEDTLS_BACKEND_PREFIX}>
+            $<TARGET_FILE:${IMAGE}mbedcrypto_${MBEDTLS_BACKEND_PREFIX}>
             ${BACKEND_RENAMED_LIBARY}
-    DEPENDS mbedcrypto_${MBEDTLS_BACKEND_PREFIX}
+    DEPENDS ${IMAGE}mbedcrypto_${MBEDTLS_BACKEND_PREFIX}
             ${CMAKE_CURRENT_BINARY_DIR}/symbol_rename_${MBEDTLS_BACKEND_PREFIX}.txt
   )
-  add_custom_target(${MBEDTLS_BACKEND_PREFIX}_backend_target
+  add_custom_target(${IMAGE}${MBEDTLS_BACKEND_PREFIX}_backend_target
                     DEPENDS ${BACKEND_RENAMED_LIBARY}
   )
-  add_library(mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend STATIC IMPORTED GLOBAL)
-  add_dependencies(mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend
-                   ${MBEDTLS_BACKEND_PREFIX}_backend_target)
-  set_target_properties(mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend
+  add_library(${IMAGE}mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend STATIC IMPORTED GLOBAL)
+  add_dependencies(${IMAGE}mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend
+                   ${IMAGE}${MBEDTLS_BACKEND_PREFIX}_backend_target)
+  set_target_properties(${IMAGE}mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend
                         PROPERTIES IMPORTED_LOCATION
                         "${CMAKE_CURRENT_BINARY_DIR}/${BACKEND_RENAMED_LIBARY}")
 
-  zephyr_append_cmake_library(mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend)
+  zephyr_append_cmake_library(${IMAGE}mbedcrypto_${MBEDTLS_BACKEND_PREFIX}_backend)
 endfunction()
 
--- a/nrf_security/src/mbedtls/CMakeLists.txt
+++ b/nrf_security/src/mbedtls/CMakeLists.txt
@@ -121,44 +121,56 @@ PREPEND(src_tls_replacement ${NRF_MBEDTLS_ROOT}/src/mbedtls/replacements/
 zephyr_include_directories(${ARM_MBEDTLS_ROOT}/include)
 zephyr_include_directories(${ARM_MBEDTLS_ROOT}/include/mbedtls)
 
-add_library(mbedcrypto_vanilla STATIC "")
-set(ZEPHYR_CURRENT_LIBRARY mbedcrypto_vanilla)
+# When using multiple backends this library should not be linked directly as
+# there might be name collisions with other libs, e.g. mbedcrypto_cc310
+# When multiple backends are enabled, unique symbols are guaranteed through
+# symbol renaming and glue layer.
+# Therefore add_library is used instead of zephyr_library
+add_library(${IMAGE}mbedcrypto_vanilla STATIC "")
+set(ZEPHYR_CURRENT_LIBRARY ${IMAGE}mbedcrypto_vanilla)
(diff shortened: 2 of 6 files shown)


nrf_security/src/mbedtls/CMakeLists.txt
```diff
@@ -121,43 +121,43 @@ PREPEND(src_tls_replacement ${NRF_MBEDTLS_ROOT}/src/mbedtls/replacements/
...
-add_library(mbedx509_vanilla STATIC "")
-set(ZEPHYR_CURRENT_LIBRARY mbedx509_vanilla)
+add_library(${IMAGE}mbedx509_vanilla STATIC "")
+set(ZEPHYR_CURRENT_LIBRARY ${IMAGE}mbedx509_vanilla)
 zephyr_library_sources(${src_x509})
 zephyr_library_include_directories(${common_includes})
 zephyr_library_compile_definitions(-DMBEDTLS_CONFIG_FILE=${MBEDTLS_CONFIGURATION})
-zephyr_library_link_libraries(zephyr_interface)
+zephyr_library_link_libraries(${IMAGE}zephyr_interface)
 if(TARGET mbedcrypto_cc310)
```
hakonfam: Should this be prefixed?

nrf_security/src/mbedtls/CMakeLists.txt
```diff
@@ -121,43 +121,43 @@ PREPEND(src_tls_replacement ${NRF_MBEDTLS_ROOT}/src/mbedtls/replacements/
 zephyr_include_directories(${ARM_MBEDTLS_ROOT}/include)
 zephyr_include_directories(${ARM_MBEDTLS_ROOT}/include/mbedtls)
 
-add_library(mbedcrypto_vanilla STATIC "")
-set(ZEPHYR_CURRENT_LIBRARY mbedcrypto_vanilla)
+add_library(${IMAGE}mbedcrypto_vanilla STATIC "")
```
hakonfam: Could you use `zephyr_library_named()` here? To avoid having to set the current library afterwards.

nrf_security/src/mbedtls/CMakeLists.txt
```diff
@@ -121,43 +121,43 @@ PREPEND(src_tls_replacement ${NRF_MBEDTLS_ROOT}/src/mbedtls/replacements/
...
 zephyr_library_include_directories(${common_includes})
-zephyr_library_link_libraries(zephyr_interface)
+zephyr_library_link_libraries(${IMAGE}zephyr_interface)
 zephyr_library_compile_definitions_ifdef(CONFIG_MBEDTLS_CIPHER_AES_256_ECB_C MBEDTLS_CIPHER_AES_256_ECB_C)
 zephyr_library_compile_definitions_ifdef(CONFIG_MBEDTLS_CIPHER_AES_256_CBC_C MBEDTLS_CIPHER_AES_256_CBC_C)
 zephyr_library_compile_definitions_ifdef(CONFIG_MBEDTLS_CIPHER_AES_256_CTR_C MBEDTLS_CIPHER_AES_256_CTR_C)
 zephyr_library_compile_definitions_ifdef(CONFIG_MBEDTLS_CIPHER_AES_256_CCM_C MBEDTLS_CIPHER_AES_256_CCM_C)
 if(TARGET mbedcrypto_cc310)
-  target_link_libraries(mbedcrypto_vanilla PRIVATE mbedcrypto_cc310)
+  target_link_libraries(${IMAGE}mbedcrypto_vanilla PRIVATE ${IMAGE}mbedcrypto_cc310)
```
hakonfam: Could you use `zephyr_link_libraries(${IMAGE}mbedcrypto_cc310)` here?

nrf_security/src/mbedtls/CMakeLists.txt
```diff
@@ -121,43 +121,43 @@ PREPEND(src_tls_replacement ${NRF_MBEDTLS_ROOT}/src/mbedtls/replacements/
...
+set(ZEPHYR_CURRENT_LIBRARY ${IMAGE}mbedcrypto_vanilla)
 zephyr_library_sources(${src_crypto} ${src_crypto_replacement})
 zephyr_library_include_directories(${common_includes})
-zephyr_library_link_libraries(zephyr_interface)
+zephyr_library_link_libraries(${IMAGE}zephyr_interface)
 zephyr_library_compile_definitions_ifdef(CONFIG_MBEDTLS_CIPHER_AES_256_ECB_C MBEDTLS_CIPHER_AES_256_ECB_C)
 zephyr_library_compile_definitions_ifdef(CONFIG_MBEDTLS_CIPHER_AES_256_CBC_C MBEDTLS_CIPHER_AES_256_CBC_C)
 zephyr_library_compile_definitions_ifdef(CONFIG_MBEDTLS_CIPHER_AES_256_CTR_C MBEDTLS_CIPHER_AES_256_CTR_C)
 zephyr_library_compile_definitions_ifdef(CONFIG_MBEDTLS_CIPHER_AES_256_CCM_C MBEDTLS_CIPHER_AES_256_CCM_C)
 if(TARGET mbedcrypto_cc310)
```
hakonfam: Does mbedcrypto_cc310 not support being built as multi image, or should this be prefixed?

Edit: I see it is prefixed other places, so I guess it should be so here as well.

nrf_security/src/mbedtls/CMakeLists.txt
```diff
@@ -121,43 +121,43 @@ PREPEND(src_tls_replacement ${NRF_MBEDTLS_ROOT}/src/mbedtls/replacements/
...
   zephyr_library_include_directories(
-    $<TARGET_PROPERTY:mbedcrypto_cc310,INTERFACE_INCLUDE_DIRECTORIES>
+    $<TARGET_PROPERTY:${IMAGE}mbedcrypto_cc310,INTERFACE_INCLUDE_DIRECTORIES>
   )
 endif()
 
-add_library(mbedtls_vanilla STATIC "")
-set(ZEPHYR_CURRENT_LIBRARY mbedtls_vanilla)
+add_library(${IMAGE}mbedtls_vanilla STATIC "")
```
hakonfam: Again, `zephyr_library_named(mbedtls_vanilla)` ? Or is there some reason this is not used?

nrf_security/src/mbedtls/CMakeLists.txt
tejlmand: yes, otherwise we'll end up getting zephyr version `autoconf.h` on some files in addition to, as example `spm/zephyr` version of `autoconf.h`

nrf_security/src/mbedtls/CMakeLists.txt
tejlmand: I remember at early days, there were some issues with that in relation to symbol renaming, but can't recall the exact nature of the problem, and thus, for this particular fix I consider it safest not to change too much.

But I agree, it is worth to investigate if this can be simplified, but I would suggest a separet commit and PR for that, to keep such changes independent of this PR.

tejlmand: Followup, not using `zephyr_library_named()` allows the lib not to go into `whole-archive` and thus only being linked if needed, in same fashion as when Zephyr links to externally mbedtls library.

nrf_security/src/mbedtls/CMakeLists.txt
tejlmand: Good catch.

Will fix.

nrf_security/cmake/symbol_rename.cmake
```diff
@@ -53,21 +53,21 @@ function(library_redefine_symbols backend template)
     COMMAND ${CMAKE_OBJCOPY} 
             --redefine-syms
             ${CMAKE_CURRENT_BINARY_DIR}/symbol_rename_${MBEDTLS_BACKEND_PREFIX}.txt
-            $<TARGET_FILE:mbedcrypto_${MBEDTLS_BACKEND_PREFIX}>
-            ${BACKEND_RENAMED_LIBARY}
-    DEPENDS mbedcrypto_${MBEDTLS_BACKEND_PREFIX}
+            $<TARGET_FILE:${IMAGE}mbedcrypto_${MBEDTLS_BACKEND_PREFIX}>
+            ${IMAGE}${BACKEND_RENAMED_LIBARY}
```
oyvindronningstad: Missing corresponding `${IMAGE}` in line 52

nrf_security/src/mbedtls/CMakeLists.txt
hakonfam: Ok, but what is the reason for needing to not be whole-archived here? 

I'm asking since the rest of the zephyr libraries are created in that fashion, what is the difference here?

nrf_security/src/mbedtls/CMakeLists.txt
tejlmand: no, cause that would cause it to be `interface` linking on zephyr_interface target and thus propagated to all targets.
mbedcrypto_cc310 is an imported target that should just be needed by vanilla target here.

nrf_security/src/mbedtls/CMakeLists.txt
tejlmand: as discussed elsewher, will add a comment

nrf_security/src/mbedtls/CMakeLists.txt
tejlmand: will do.

carlescufi:
@tejlmand can you rebase and address review comments?

carlescufi:
@hakonfam please re-review

nrf_security/cmake/symbol_rename.cmake
tejlmand: fixed. updated file was dangling in editor :(

PR #61, opened 2019-06-03
bsdlib: Moved bsd library before TOOLCHAIN_LIBS in CMakeLists.txt

osaether:
Added -lgcc at end of target_link_libraries in
CMakeLists.txt to fix undefined reference to
`__aeabi_uldivmod' when building samples that use bsdlib
and have CONFIG_DEBUG_OPTIMIZATIONS enabled.

Signed-off-by: Ole Sæther <ole.saether@nordicsemi.no>

### DIFF
--- a/bsdlib/CMakeLists.txt
+++ b/bsdlib/CMakeLists.txt
@@ -19,12 +19,8 @@ endif()
 # lib oberon are linked in because they are used by the bsd library
 # and must therefore be placed after the bsd library on the linker's
 # command line.
-set(BSDLIB_TARGET ${IMAGE}bsd_nrf9160_xxaa)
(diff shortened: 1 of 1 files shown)


osaether:
@lemrey  Please review

osaether:
> Please describe why the existing solution is not working ?

Becasue libgcc needs to come after libbsd_nrf9160_xxaa on the linker command line.

tejlmand:
> > Please describe why the existing solution is not working ?
> 
> Becasue libgcc needs to come after libbsd_nrf9160_xxaa on the linker command line.

But shouldn't that be fixed in upstream Zephyr, so that the `TOOLCHAIN_LIBS` are always linked last ?
@SebastianBoe comments ?

carlescufi:
@SebastianBoe can you please review?

PR #68, opened 2019-06-26
crypto: cc310_bl: Reword prompt

oyvindronningstad:
Not only nRF52840 has CC310.

Signed-off-by: Øyvind Rønningstad <oyvind.ronningstad@nordicsemi.no>


oyvindronningstad:
cc @frkv

PR #70, opened 2019-06-28
[DNM] bsdlib: Update BSD library for nrf9160

gbakke:
Library version 4b15e49f362.

Signed-off-by: Glenn Ruben Bakke <glenn.ruben.bakke@nordicsemi.no>


b-gent:
@gbakke  Does this mean that we should change BSD library version to 0.3.3 in NCS v1.0.0 release notes?

gbakke:
> Does this mean that we should change BSD library version to 0.3.3 in NCS v1.0.0 release notes?

@b-gent , yes, if we merge this and the corresponding manifest update in nrf-repo, we should do that.

PR #73, opened 2019-07-01
NFC: Add license

grochu:


grochu:
@rlubos could you review?

PR #74, opened 2019-07-01
nfc: add license

grochu:
Adding license reference for NFC libraries.

Signed-off-by: Michał Grochala <michal.grochala@nordicsemi.no>


grochu:
@rlubos could you review?

PR #83, opened 2019-09-13
Add Kconfig option CONFIG_SHORT_WCHAR_T

frkv:
-Added Kconfig option to indicate build with -fshort-wchar
-Updated nrfxlib_caculate_lib_path to also select short-wchar

### DIFF
--- a/common.cmake
+++ b/common.cmake
@@ -10,20 +10,29 @@ if(__NRFXLIB_COMMON_CMAKE__)
 endif()
 set(__NRFXLIB_COMMON_CMAKE__ TRUE)
 
+#
+# Build the path of the library taking in
+# -Arch type
+# -Float ABI
+# -wchar_t size
+#
 function(nrfxlib_calculate_lib_path lib_path)
-  if(CONFIG_FLOAT)
-	if(CONFIG_FP_HARDABI)
-      set(float_dir hard-float)
-	elseif(CONFIG_FP_SOFTABI)
-      set(float_dir softfp-float)
-	else()
-      assert(0 "Unreachable code")
-	endif()
-  else()
-	set(float_dir soft-float)
-  endif()
-
+  # Add Arch type
   assert(GCC_M_CPU "GCC_M_CPU must be set to find correct lib.")
-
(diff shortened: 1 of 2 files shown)


crypto/Kconfig
```diff
@@ -3,6 +3,13 @@ menu "Crypto libraries for nRF5x SOCs."
 config NRFXLIB_CRYPTO
 	bool
 
+config SHORT_WCHAR_T
+	bool "Compile with short wchar_t size (2 bytes)"
+	default n
```
SebastianBoe: default n is implied and should be omitted.

common.cmake
```diff
@@ -10,20 +10,32 @@ if(__NRFXLIB_COMMON_CMAKE__)
...
+  # Append floating ABI
   if(CONFIG_FLOAT)
 	if(CONFIG_FP_HARDABI)
       set(float_dir hard-float)
 	elseif(CONFIG_FP_SOFTABI)
       set(float_dir softfp-float)
 	else()
       assert(0 "Unreachable code")
-	endif()
+  endif()
```
SebastianBoe: indentation is wrong.

common.cmake
```diff
@@ -10,20 +10,32 @@ if(__NRFXLIB_COMMON_CMAKE__)
...
   else()
 	set(float_dir soft-float)
   endif()
-
-  assert(GCC_M_CPU "GCC_M_CPU must be set to find correct lib.")
-
-  set(${lib_path} lib/${GCC_M_CPU}/${float_dir} PARENT_SCOPE)
```
SebastianBoe: It is cleaner to do

`set(${lib_path} lib/${GCC_M_CPU}/${float_dir}${wchar_dir} PARENT_SCOPE)
`

Note that there is intentionally no / between float_dir and wchar_dir (because wchar_dir would include it when enabled).

Then we can still read how the path looks in general.

crypto/Kconfig
frkv: Fixed

common.cmake
frkv: Fixed

PR #95, opened 2019-11-04
nfc: Fix default IRQ priority when ZERO_LATENCY_IRQS are used

thomasstenersen:
Fixes NCSDK-3404 and a small documentation bug.

Signed-off-by: Thomas Stenersen <thomas.stenersen@nordicsemi.no>

### DIFF
--- a/nfc/Kconfig
+++ b/nfc/Kconfig
@@ -27,10 +27,11 @@ config NFCT_IRQ_PRIORITY
 	int
 	prompt "Interrupt priority"
 	range 0 6
+	default 5 if ZERO_LATENCY_IRQS
 	default 6
 	help
 		Sets NFC interrupt priority.
-		Levels are from 0 (highest priority) to 7 (lowest priority)
+		Levels are from 0 (highest priority) to 6 (lowest priority)
 
 choice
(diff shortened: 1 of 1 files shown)


thomasstenersen:
@grochu: This should fix the issue. FYI @qung @shantha-14.

joerchan:
@KAGA164 Could you check this?

thomasstenersen:
It doesn't really have anything to do with the low latency mode. It's just that when `CONFIG_ZERO_LATENCY_IRQS=y`, which the `BT_LL_NRFXLIB` sets, the number of IRQ levels for the kernel is reduced by one. When this config is set and the NFC lib defaults to 6, this assert will fail: https://github.com/zephyrproject-rtos/zephyr/blob/master/arch/arm/core/irq_manage.c#L93

grochu:
Alright I see. Perhaps we should change the way we set the NFCT IRQ priority in general (maybe use the default from DTS?), but that's a topic for another PR.

joerchan:
> Note that @KAGA164 is at the UPF right now and might not be able to review.

Yeah. I think he is not needed. We have enough eyes now.

PR #99, opened 2019-11-06
nrf_security: Replace problematic static assert

torsteingrindvik:
Some compile issues were found due to static assert usage.
Replace with #if defined(...) ... #endif blocks.

Signed-off-by: Torstein Grindvik <torstein.grindvik@nordicsemi.no>

### DIFF
--- a/nrf_security/src/backend/nrf5x/entropy_nrf5x.c
+++ b/nrf_security/src/backend/nrf5x/entropy_nrf5x.c
@@ -8,13 +8,6 @@
 #include <entropy.h>
 #include <mbedtls/entropy.h>
 
-#if !defined(CONFIG_ENTROPY_GENERATOR)
-#error "CONFIG_ENTROPY_GENERATOR is not enabled."
-#endif
-#if !defined(CONFIG_ENTROPY_HAS_DRIVER)
-#error "CONFIG_ENTROPY_HAS_DRIVER is not set."
-#endif
-
 int mbedtls_hardware_poll(void *data,
                           unsigned char *output,
                           size_t len,
--- a/nrf_security/src/backend/nrf5x/entropy_nrf5x.c
+++ b/nrf_security/src/backend/nrf5x/entropy_nrf5x.c
@@ -8,8 +8,12 @@
 #include <entropy.h>
 #include <mbedtls/entropy.h>
 
-static_assert(defined(CONFIG_ENTROPY_GENERATOR), "CONFIG_ENTROPY_GENERATOR is not enabled.");
-static_assert(defined(CONFIG_ENTROPY_HAS_DRIVER), "CONFIG_ENTROPY_HAS_DRIVER is not set.");
+#if !defined(CONFIG_ENTROPY_GENERATOR)
+#error "CONFIG_ENTROPY_GENERATOR is not enabled."
+#endif
+#if !defined(CONFIG_ENTROPY_HAS_DRIVER)
+#error "CONFIG_ENTROPY_HAS_DRIVER is not set."
+#endif
 
 int mbedtls_hardware_poll(void *data,
                           unsigned char *output,
--- a/nrf_security/src/mbedtls/CMakeLists.txt
+++ b/nrf_security/src/mbedtls/CMakeLists.txt
@@ -131,11 +131,12 @@ zephyr_library_sources_ifdef(CONFIG_MBEDTLS_ENABLE_HEAP
   ${NRF_SECURITY_ROOT}/src/mbedtls/mbedtls_heap.c
 )
 
-if(CONFIG_SOC_NRF52840 OR CONFIG_SOC_NRF9160)
-  zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/cc310/replacements/entropy.c)
-else()
+if (CONFIG_ENTROPY_NRF5_RNG)
   zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/nrf5x/entropy_nrf5x.c)
+elseif(CONFIG_ENTROPY_CC310)
+  zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/cc310/replacements/entropy.c)
 endif()
+
 zephyr_library_link_libraries(mbedtls_common)
 nrf_security_debug_list_target_files(mbedtls_base_vanilla)
 


nrf_security/src/backend/nrf5x/entropy_nrf5x.c
```diff
@@ -8,8 +8,12 @@
...
 #include <mbedtls/entropy.h>
 
-static_assert(defined(CONFIG_ENTROPY_GENERATOR), "CONFIG_ENTROPY_GENERATOR is not enabled.");
-static_assert(defined(CONFIG_ENTROPY_HAS_DRIVER), "CONFIG_ENTROPY_HAS_DRIVER is not set.");
+#if !defined(CONFIG_ENTROPY_GENERATOR)
+#error "CONFIG_ENTROPY_GENERATOR is not enabled."
+#endif
+#if !defined(CONFIG_ENTROPY_HAS_DRIVER)
+#error "CONFIG_ENTROPY_HAS_DRIVER is not set."
+#endif
```
tejlmand: This looks like an upside-down testing.

I would prefer that we update the Kconfig and CMakeLists.txt file so that this file is only chosen if to be included using:
`zephyr_library_sources_ifdef(... entropy_nrf5x.c)

torsteingrindvik: Changed in https://github.com/NordicPlayground/nrfxlib/pull/99/commits/f6f1b80f881e09748c507735fe9834f3eb173bc8

nrf_security/src/mbedtls/CMakeLists.txt
```diff
@@ -184,11 +184,14 @@ zephyr_library_sources_ifdef(VANILLA_ONLY_MBEDTLS_CHACHAPOLY_C
 zephyr_library_sources_ifdef(CONFIG_MBEDTLS_ENABLE_HEAP ${NRF_SECURITY_ROOT}/src/mbedtls/mbedtls_heap.c)
 zephyr_library_app_memory(k_mbedtls_partition)
 
-if(CONFIG_SOC_NRF52840 OR CONFIG_SOC_NRF9160)
-  zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/cc310/replacements/entropy.c)
-else()
-  zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/nrf5x/entropy_nrf5x.c)
+if(CONFIG_ENTROPY_GENERATOR)
```
torsteingrindvik: I don't know the build system very well so which `CONFIG_ENTROPY_*` should be here is guesswork for me :-)
This does compile for me on both an nrf52840 and an nrf52810.

Please let me know if some checks here should be reordered/added/removed.

torsteingrindvik:
@tejlmand ping

nrf_security/src/mbedtls/CMakeLists.txt
```diff
@@ -184,11 +184,14 @@ zephyr_library_sources_ifdef(VANILLA_ONLY_MBEDTLS_CHACHAPOLY_C
...
+  if(CONFIG_SOC_NRF52840 OR CONFIG_SOC_NRF9160)
+    zephyr_library_sources_ifdef(CONFIG_ENTROPY_HAS_DRIVER ${NRF_SECURITY_ROOT}/src/backend/cc310/replacements/entropy.c)
```
tejlmand, line 189: So in case I have:
```
CONFIG_ENTROPY_GENERATOR=y
CONFIG_SOC_NRF52840=y
CONFIG_ENTROPY_HAS_DRIVER=n
```
no file is added to the lib.
I assume that is intended ?
This is very different from earlier version, where a file would always be added to the lib.

torsteingrindvik: You're definitely right.

Looking at `entropy.c` its contents are guarded by `MBEDTLS_ENTROPY_C`.

And looking at `entropy_nrfx5.c` it will return an error code if `ENTROPY_NAME` isn't defined, which is defined when `ENTROPY_GENERATOR` is.

Perhaps it is more correct to revert the change in include logic, and simply have a similar guard in `entropy_nrf5x.c`?

tejlmand: > Perhaps it is more correct to revert the change in include logic, and simply have a similar guard in entropy_nrf5x.c?

we should try to avoid the user to do something foolish, like being able to include the file, and then later get a compilation error.

Then it is better to ensure correct dependency checking in Kconfig, and only include files of the right dependencies are met.

torsteingrindvik: Ok sounds reasonable.
I'm not familiar enough with the build scheme to know what the appropriate logic for including these files are though so I'll need help deciding what to check for the two files in question.

tejlmand: will take a closer look, and try to determine which of those two files, along with the CMakeLists.txt that is using the most appropriate solution, and then we could try to align them in this PR.
Will post comment when I have looked into it.

frkv: Proposal:

`if (CONFIG_ENTROPY_GENERATOR)
  if (CONFIG_ENTROPY_NRF5_RNG)
    zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/nrf5x/entropy_nrf5x.c)
  elseif(CONFIG_SOC_NRF52840 OR CONFIG_SOC_NRF9160)
    zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/cc310/replacements/entropy.c)
  endif()
endif()`

This allows for addition of CONFIG_SOC_NRF5340 in due time and lets you override and use nRF5 hardware entropy in nrF52840

tejlmand:
please rebase

torsteingrindvik:
> please rebase

done

nrf_security/src/mbedtls/CMakeLists.txt
tejlmand: Did some testing:
If disabling `CC310_ENTROPY` (which also disables `ENTROPY_HAS_DRIVER`), but keep `ENTROPY_GENERATOR=y` then you get the following error during linking:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:      184728 B     441856 B     41.81%
            SRAM:       73948 B       128 KB     56.42%
        IDT_LIST:         120 B         2 KB      5.86/opt/zephyr-sdk/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/9.2.0/../../../../arm-zephyr-eabi/bin/ld: modules/nrf/subsys/net/lib/aws_jobs/lib..__nrf__subsys__net__lib__aws_jobs.a(aws_jobs.c.obj): in function `publish':
/projects/github/ncs/nrf/subsys/net/lib/aws_jobs/src/aws_jobs.c:231: undefined reference to `sys_rand32_get'
collect2: error: ld returned 1 exit status
%
ninja: build stopped: subcommand failed.

```

tejlmand: seems to be the `AWS_JOBS` missing some deps. @sigvartmh this might not be an issue related to nrf_security, but instead to aws jobs. Could you look into this ?

sigvartmh: I guess the Kconfig for AWS Jobs is lacking some dependency on random. I'm not really sure what you want me to look into.

I would assume the Kconfig of something that uses this lacks

```
CONFIG_ENTROPY_GENERATOR=y
CONFIG_TEST_RANDOM_GENERATOR=y
```

To ensure that sys_rand32_get is linked.

tejlmand: Or we could simplify to:
```
if (CONFIG_ENTROPY_NRF5_RNG)
   zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/nrf5x/entropy_nrf5x.c)
elseif(CONFIG_ENTROPY_CC310)
    zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/cc310/replacements/entropy.c)
endif()
```
which will remove the need for testing against `CONFIG_ENTROPY_GENERATOR` cause both `ENTROPY_NRF5_RNG` and `ENTROPY_CC310` are depends on `ENTROPY_GENERATOR`.

And it still allows users to choose `ENTROPY_NRF5_RNG` over `CC310`, which happens when using the nordic LL.
Plus the fact that due to dependencies, you cannot select both `CONFIG_ENTROPY_NRF5_RNG` and `CONFIG_ENTROPY_CC310` at the same time, but you can select `CONFIG_ENTROPY_NRF5_RNG` and be a `CONFIG_SOC_NRF52840` at the same time.

tejlmand: > 
> 
> I guess the Kconfig for AWS Jobs is lacking some dependency on random. I'm not really sure what you want me to look into.
> 
> I would assume the Kconfig of something that uses this lacks
> 

@sigvartmh  As you authored the AWS_JOB Kconfig:
https://github.com/NordicPlayground/fw-nrfconnect-nrf/commit/ad719458e1620c5603e10a11b9912938977e92d0

I simply ask you to update the Kconfig file to ensure that users can only select `AWS_JOBS` when dependencies are enabled.

From my finding, it appears that `AWS_JOBS` has a dependency to entropy, so probably something like:
```
menuconfig AWS_JOBS
	bool "AWS Jobs library"
        depends ENTROPY_GENERATOR
```
here: https://github.com/NordicPlayground/fw-nrfconnect-nrf/blob/master/subsys/net/lib/aws_jobs/Kconfig#L6-L7

But you are probably the best candidate to know the exact dependency that must be added, to avoid users making an invalid configuration.

torsteingrindvik:
Rebased and used suggestion by @tejlmand .
Did a sanity check by building some samples with nrf52840, 9160, 52810.
Seems to work out nicely.

nrf_security/src/mbedtls/CMakeLists.txt
sigvartmh: I was a bit unsure what you wanted me todo if you want me to create a patch which adds this that's fine by me. What you suggested looks good in my mind but I guess it has to be added to the configuration of every sample which uses `aws_jobs`.

Do you want to add it in this PR or do you want me to open a separate one?

sigvartmh: Opened pr here https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/2189

torsteingrindvik:
This is now cherry-picked and squashed into https://github.com/nrfconnect/sdk-nrfxlib/pull/205

tejlmand:
> This is now cherry-picked and squashed into #205

@torsteingrindvik I assume this PR can be closed then.

torsteingrindvik:
@tejlmand yes it seems #205 will be merged, so closing.

PR #100, opened 2019-11-06
nrf_security: introduces mbedtls heap in nrf_security

tejlmand:
This commit introduces mbedtls_heap.c in nrf security and also add
globals in compiled libraries to userspace config using
zephyr_library_app_memory.

### DIFF
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -47,6 +47,30 @@ config MBEDTLS_TLS_LIBRARY
 	  Create the mbed SSL/TLS library in addition to the mbed crypto
 	  library.
 
+menu "mbed TLS memory configuration"
+
+config MBEDTLS_ENABLE_HEAP
+	bool "Enable global heap for mbed TLS"
+	help
+	  This option enables the mbedtls to use the heap. This setting must
+	  be global so that various applications and libraries in Zephyr do not
+	  try to do this themselves as there can be only one heap defined
+	  in mbedtls. If this is enabled, then the Zephyr will, during the device
+	  startup, initialize the heap automatically.
+
+config MBEDTLS_HEAP_SIZE
+	int "Heap size for mbed TLS"
+	default 512
+	depends on MBEDTLS_ENABLE_HEAP
+	help
+	  The mbedtls routines will use this heap if enabled.
+	  For streaming communication with arbitrary (HTTPS) servers on the
+	  Internet, 32KB + overheads (up to another 20KB) may be needed.
+	  Ensure to adjust the heap size according to the need of the
+	  application.
+
+endmenu
+
 comment "Backend Selection"
 
 config CC310_BACKEND
--- a/nrf_security/src/mbedtls/mbedtls_heap.c
+++ b/nrf_security/src/mbedtls/mbedtls_heap.c
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2019 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+ */
+
+#include <init.h>
+#include <app_memory/app_memdomain.h>
+
+#include "mbedtls/memory_buffer_alloc.h"
+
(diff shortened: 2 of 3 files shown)


nrf_security/Kconfig
```diff
@@ -47,6 +47,31 @@ config MBEDTLS_TLS_LIBRARY
...
+	  in mbedtls. If this is enabled, then the Zephyr will, during the device
+	  startup, initialize the heap automatically.
+
+config MBEDTLS_HEAP_SIZE
+	int "Heap size for mbed TLS"
+	default 512
+	depends on MBEDTLS_ENABLE_HEAP
+	help
+	  The mbedtls routines will use this heap if enabled.
+	  Default value for the heap size is not set as it depends on the
```
frkv: There is a default heap size, but then the help-text is that we don't set a default value..

nrf_security/Kconfig
```diff
@@ -47,6 +47,31 @@ config MBEDTLS_TLS_LIBRARY
...
+	bool "Enable global heap for mbed TLS"
+	help
+	  This option enables the mbedtls to use the heap. This setting must
+	  be global so that various applications and libraries in Zephyr do not
+	  try to do this themselves as there can be only one heap defined
+	  in mbedtls. If this is enabled, then the Zephyr will, during the device
+	  startup, initialize the heap automatically.
+
+config MBEDTLS_HEAP_SIZE
+	int "Heap size for mbed TLS"
```
frkv: In bytes?

nrf_security/Kconfig
```diff
@@ -47,6 +47,31 @@ config MBEDTLS_TLS_LIBRARY
...
+	default 512
+	depends on MBEDTLS_ENABLE_HEAP
+	help
+	  The mbedtls routines will use this heap if enabled.
+	  Default value for the heap size is not set as it depends on the
+	  application. For streaming communication with arbitrary (HTTPS)
+	  servers on the Internet, 32KB + overheads (up to another 20KB) may
+	  be needed. For some dedicated and specific usage of mbedtls API, the
```
frkv: Superfluous ", the"

nrf_security/src/mbedtls/mbedtls_heap.c
```diff
@@ -0,0 +1,30 @@
...
+static int mbedtls_heap_init(struct device *dev)
+{
+	ARG_UNUSED(dev);
+
+	mbedtls_memory_buffer_alloc_init(mbedtls_heap, sizeof(mbedtls_heap));
+
+	return 0;
+}
+
+/* Entropy cc310 is initialized with CONFIG_KERNEL_INIT_PRIORITY_DEVICE and the
```
frkv: hw_cc310 is the thing that must be initialized before

nrf_security/Kconfig
tejlmand: yep, this is a bit puzzling, but I decided to just use the upstream description for now

nrf_security/Kconfig
tejlmand: no, echo_server uses `80000` bytes afaik

nrf_security/src/mbedtls/mbedtls_heap.c
tejlmand: but entropy must be init before heap, correct ?
I prefer to only state first dependency here, and then we should consider to state the entropy -> hw dependency in the entropy driver.

frkv: entropy is only dependent on hw_cc310 (because it requires mutex-usage). So the chain is hw_cc310->heap or hw_cc310->entropy_cc310->heap (dependent of the presence of entropy in your configuraiton)

frkv: So entropy and heap could happen on same level, actually... They are not dependent on each other (since the entropy_cc310 driver only requires mutex-support, and the same is the thing for heap)

tejlmand: then I will change priority to match entropy priority as that will ensure mbedTLS is ready before app init prio.
No reason to keep mbed TLS heap on application init priority.

tejlmand:
@frkv comments addressed

PR #104, opened 2019-11-15
nfc: Remove nfc platform from nrfxlib

KAGA164:
nfc_platform moved to NCS.


KAGA164:
@kapi-no `select NRFX_NFCT` moved to NCS

nfc/Kconfig
```diff
@@ -4,58 +4,12 @@
...
-config NRFXLIB_NFC
-	bool
-	prompt "Enable NFC"
-	select NRFX_NFCT
-	help
-	  Enable NFC functionality
-
-if NRFXLIB_NFC
-
 comment "Common NFC module configuration"
```
anangl: This does not seem to be needed anymore.

KAGA164: Ok, removed.

nfc/Kconfig
```diff
@@ -4,58 +4,10 @@
...
-	  Enable NFC functionality
-
-if NRFXLIB_NFC
-
-comment "Common NFC module configuration"
-
-config NFC_T2T_LIB_ENABLED
+config NFC_T2T_NRFXLIB
 	bool
 	prompt "Enable NFC Type 2 Tag library"
```
anangl: On the occasion, you could also get rid of these `prompt`s here.
```
bool "Enable NFC Type 2 Tag library"
```

KAGA164: Ok changed.

PR #108, opened 2019-11-18
Kconfig: ble_controller: Update to upstream Kconfig support configs.

joerchan:
Add Kconfig support configuration symbols to BLE controller for so that
dependent configurations can be set from this. This enables features
in the host e.g. PHY update.


joerchan:
@rugeGerritsen Good point. Thank you.

PR #116, opened 2019-11-29
nrf_security: Fixing wchar_t support in generated libraries

frkv:
-Add -fshort-wchar for all libraries generated by nrf_security build

Ref: NCSDK-3820

### DIFF
--- a/nrf_security/src/mbedcrypto_glue/CMakeLists.txt
+++ b/nrf_security/src/mbedcrypto_glue/CMakeLists.txt
@@ -32,6 +32,8 @@ if(CONFIG_NRF_CRYPTO_GLUE_LIBRARY)
   #
   zephyr_library_named(mbedcrypto_glue)
 
+  zephyr_library_compile_options_ifdef(CONFIG_SHORT_WCHAR_T -fshort-wchar)
+
   #
   # Add glued files if enabled
   #
--- a/nrf_security/src/mbedcrypto_glue/vanilla/vanilla_glue.cmake
+++ b/nrf_security/src/mbedcrypto_glue/vanilla/vanilla_glue.cmake
@@ -29,6 +29,10 @@ nrf_security_debug("######### Creating vanilla noglue library #########")
 #
 add_library(${IMAGE}mbedcrypto_vanilla)
 
+if (CONFIG_SHORT_WCHAR_T)
+  target_compile_options(${IMAGE}mbedcrypto_vanilla PRIVATE -fshort-wchar)
+endif()
+
 #
 # Adding all standard compile/linker options (e.g. float ABI)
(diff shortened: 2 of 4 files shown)


SebastianBoe:
Intuitively, I would think we would want to use 

  zephyr_compile_options_ifdef(CONFIG_SHORT_WCHAR_T -fshort-wchar)

what is the rationale for adding it per-library? Does it not need to be applied to all source files?

frkv:
> Intuitively, I would think we would want to use
> 
> zephyr_compile_options_ifdef(CONFIG_SHORT_WCHAR_T -fshort-wchar)
> 
> what is the rationale for adding it per-library? Does it not need to be applied to all source files?

This is the fix. We tried zephyr_compile_options_ifdef on a broader scope, and it didn't work

nrf_security/src/mbedcrypto_glue/CMakeLists.txt
```diff
@@ -32,6 +32,8 @@ if(CONFIG_NRF_CRYPTO_GLUE_LIBRARY)
   #
   zephyr_library_named(mbedcrypto_glue)
 
+  zephyr_library_compile_options_ifdef(CONFIG_SHORT_WCHAR_T -fshort-wchar)
+
```
SebastianBoe: This should be replaced with

zephyr_compile_options_ifdef(CONFIG_SHORT_WCHAR_T -fshort-wchar)

The same for other zephyr_library_'s.

The zephyr_compile_options invocation should be placed in upstream zephyr. If there is a need to get it in early it can in addition be placed in ncs/nrf/CMakeLists.txt until the upstream patch has made it's way downstream and made the nrf patch redundant.

The same applies to config SHORT_WCHAR_T.

nrf_security/src/mbedcrypto_glue/vanilla/vanilla_glue.cmake
```diff
@@ -29,6 +29,10 @@ nrf_security_debug("######### Creating vanilla noglue library #########")
 #
 add_library(${IMAGE}mbedcrypto_vanilla)
 
+if (CONFIG_SHORT_WCHAR_T)
+  target_compile_options(${IMAGE}mbedcrypto_vanilla PRIVATE -fshort-wchar)
+endif()
```
SebastianBoe: Is there anything wrong with doing

target_link_libraries(mbedcrypto_vanilla PRIVATE zephyr_interface)

?

If there is a problem with this then we can at least use "zephyr_get_compile_options_for_lang".

Then we can get rid of this patch and the TOOLCHAIN_C_FLAGS patch below (after also applying the zephyr_compile_options patch mentioned above).

PR #117, opened 2019-11-29
nRF53 Series support for nRF BLE Controller

thomasstenersen:
This PR adds support for running the nRF BLE Controller on the nRF53 Series SoC. The controller may be run together with an application on the network core or standalone as in the `hci_rpmsg` sample in fw-nrfconnect-zephyr.

In order to access type definitions for the vendor specific HCI commands of the controller, the BT_LL_NRFXLIB_VS_INCLUDE symbol has been added and build system modified accordingly. This is currently only used in the LLPM sample in the corresponding PR to fw-nrfconnect-nrf.

The PR also contains some minor fixes for all the nRF BLE Controllers.


b-gent:
Please remember to always add a tech writer if you have RST files in your PR.

ble_controller/Kconfig
```diff
@@ -54,12 +62,13 @@ config BLE_CONTROLLER_S132
 	select BT_CTLR_ADV_EXT_SUPPORT
 
 config BLE_CONTROLLER_S140
-	depends on (SOC_NRF52811 || SOC_NRF52840)
+	depends on (SOC_NRF52811 || SOC_NRF52840 || SOC_SERIES_NRF53X)
```
sofu1: Why not SOC_NRF5340?

ble_controller/CHANGELOG.rst
```diff
@@ -5,6 +5,29 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+Added
```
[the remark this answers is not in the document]
thomasstenersen: Going forward, a new header will only be added for a new NCS tagged release. S.t. the list will contain the changes since the previous tagged version of NCS.

ble_controller/Kconfig
thomasstenersen: Even though there is only nRF5340 now, there should be no reason that the controller shouldn't work with any future series of nRF53 :) If there are special cases, we'll handle it then. The reason specific versions of the nRF52-series is used here is because of HW compability.

ble_controller/CHANGELOG.rst
```diff
@@ -5,6 +5,29 @@ Changelog
...
+=======
+
+* Version numbers have been removed from the libraries.
+
+Bugfixes
+========
+
+* Fixed an issue where an assert could occur when receiving a packet
+  with a CRC error after performing a data length procedure on Coded PHY.
+* Fixed an issue where assert occured when setting secondary PHY to 0
```
thomasstenersen: Two r's in occurred.

ble_controller/Kconfig
```diff
@@ -23,14 +37,15 @@ config BT_LL_NRFXLIB
 	select BT_CTLR_EXT_SCAN_FP_SUPPORT
 	select BT_CTLR_PHY_UPDATE_SUPPORT
 	select BT_CTLR_CHAN_SEL_2_SUPPORT
-	# This controller only supports nRF52, it does not support 51 or
+	# This controller only supports nRF52 and nRF53, it does not support 51 or
 	# 91 (91 doesn't even have the physical HW needed for BLE).
-	depends on SOC_SERIES_NRF52X
+	depends on (SOC_SERIES_NRF52X || SOC_NRF5340_CPUNET)
```
ioannisg: This needs to be changed to SOC_SERIES_NRF53X.
Not only it complies with the SOC_SERIES_NRF52X, but also, it complies with line 72.

thomasstenersen: You cannot run the controller on the app core. I should rather update this on line 72.

PR #123, opened 2019-12-17
cmake: add indirect support for -fshort-wchar

tejlmand:
This PR superseedes: PR #116

This PR will indirectly select the correct library when `-fshort-wchar` is added to Zephyr compiler options.

The reason for not exposing users directly to this flag, is because the current C libraries distributed with Zephyr SDK or Armgcc has not been compiled with `-fshort-wchar`.
The following linker failure is seen if using Zephyr distributed C libs:
`CONFIG_MINIMAL_LIBC=y`
```
warning: /opt/zephyr-sdk/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/8.3.0/thumb/v7e-m/nofp/libgcc.a(_udivmoddi4.o) uses 4-byte wchar_t yet the output is to use 2-byte wchar_t; use of wchar_t values across objects may fail
```

`CONFIG_NEWLIB_LIBC=y`
```
/opt/zephyr-sdk/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/8.3.0/../../../../arm-zephyr-eabi/bin/ld: warning: /opt/zephyr-sdk/arm-zephyr-eabi/arm-zephyr-eabi/lib/thumb/v7e-m/nofp/libc.a(lib_a-assert.o) uses 4-byte wchar_t yet the output is to use 2-byte wchar_t; use of wchar_t values across objects may fail
```

This means that if users want to use this flag they must provide their own external C lib compiled with `-fshort-wchar` and also specify: `CONFIG_EXTERNAL_LIBC`.

To select correct lib, set `CONFIG_COMPILER_OPT=-fshort-wchar`


tejlmand:
Part of linking command:
```
/opt/zephyr-sdk/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc 
....
 /projects/github/ncs/nrfxlib/crypto/nrf_cc310_mbedcrypto/lib/cortex-m4/soft-float/short-wchar/no-interrupts/libnrf_cc310_mbedcrypto_0.9.1.a
 ...
```
when `CONFIG_COMPILER_OPT=-fshort-wchar`

tejlmand:
Manifest updated: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1658

PR #126, opened 2019-12-19
ble_controller: Configurable DLE support

rugeGerritsen:
Also fixes a bug which would occur on nRF53 where the radio would stay
in the TX state longer than expected

Signed-off-by: Rubin Gerritsen <Rubin.Gerritsen@nordicsemi.no>


joerchan:
@rugeGerritsen We need a PR in -nrf repo. Which includes the version and enables DLE support in the HCI driver, right?

rugeGerritsen:
> @rugeGerritsen We need a PR in -nrf repo. Which includes the version and enables DLE support in the HCI driver, right?

Created https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1671

PR #128, opened 2020-01-09
mpsl: New version of MPSL, now also with nRF53 support

sofu1:
New version of MPSL, also adding soft-float library for nRF53.
This version of MPSL is API compatible with ble_controller without
MPSL linked in.
Also removes the _nrf52 postfix from the mpsl library name to be in
line with other nrfxlib libraries.

Signed-off-by: Solveig Fure <Solveig.Fure@nordicsemi.no>


mpsl/README.rst
```diff
@@ -1,4 +1,4 @@
-.. _mpsl:
+.. _mpsl_readme:
```
rugeGerritsen: Keep using `_mpsl` only to be consistent with other libraries?

mpsl/doc/public_mpsl.rst
```diff
@@ -0,0 +1,77 @@
+.. _external_mpsl:
```
rugeGerritsen: It would be nice to remove "public" from its name, as this does not give customers additional information. The same accounts for "_external"

mpsl/doc/public_mpsl.rst
```diff
@@ -0,0 +1,77 @@
...
+
+   It is up to the application to forward any clock related events to :cpp:func:`MPSL_IRQ_CLOCK_Handler` in lower priority.
+   Irrelevant events are ignored, so the application is free to forward all events for the POWER_CLOCK interrupt.
+
+
+ * ``low_prio_irq`` interrupt
+
+   Low priority work is signaled by MPSL by pending the IRQ specified in the ``low_prio_irq`` argument to :cpp:func:`mpsl_init`.
+   When this interrupt is triggered, :cpp:func:`mpsl_low_priority_process` should be called as soon as possible, at least within a couple of ms.
+   This issues timeslot signals like blocked or canceled, among other things.
```
rugeGerritsen: Remove the sentence about blocked/cancelled

mpsl/README.rst
sofu1: To align with those libraries that use _readme for files that are not README, but rather integration notes? I think that's strange.

mpsl/doc/public_mpsl.rst
sofu1: The reason is that the mpsl documentation build system separates between internal and public doc like this, and now this file is just copied without modifying it during deploy. We have a task for replacing the use of external with public. I can extend that task to also look into removing the whole prefix. And then we can fix it before next release?

mpsl/doc/public_mpsl.rst
sofu1: - [ ] Remove sentence

mpsl/CHANGELOG.rst
```diff
@@ -5,6 +5,22 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+Added
```
sofu1: I will add a MPSL 0.5.0-2.prealpha header here and do some minor fixes to the changes section

rugeGerritsen: Do we need the version header? We are not using the version info anywhere else

sofu1: We don't need it, but this is tracked as the 2.prealpha internally, so that's all. For the next release we should be able to follow the ble_controller release model I think.

mpsl/CHANGELOG.rst
```diff
@@ -5,6 +5,27 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+MPSL 0.5.0-2.prealpha
+*********************
+
+Added
+=====
+
+* Added library with preliminary support for the nRF5340 SoC. The feature set is
```
b-gent:
```suggestion
* Added a library version with preliminary support for the nRF5340 device.
  The feature set is the same as in the MPSL library for nRF52.
  The following library has been added:
```

mpsl/CHANGELOG.rst
```diff
@@ -5,6 +5,27 @@ Changelog
...
+  the same as in the nRF52 MPSL. The following library has been added:
```
b-gent: clear this line if the above suggestion is implemented

mpsl/CHANGELOG.rst
```diff
@@ -5,6 +5,27 @@ Changelog
...
+
+  * :file:`soft-float/libmpsl.a`
+
+Changes
+=======
+
+* Removed ``_nrf52`` from the MPSL library file names.
+* It is no longer allowed to call :cpp:func:`mpsl_init()` if MPSL is already initialized.
+* Clock configuration parameters for any stack that uses MPSL must be provided in :c:type:`mpsl_clock_lfclk_cfg_t` to :cpp:func:`mpsl_init()`,
```
b-gent:
```suggestion
* Clock configuration parameters for any stack that uses MPSL must be provided in :c:type:`mpsl_clock_lfclk_cfg_t` to :cpp:func:`mpsl_init()`.
  This now also involves clock accuracy.
```

mpsl/README.rst
b-gent: The highest level RST files for each library do not have ``_readme`` in their IDs so I would also incline towards leaving just ``_mpsl`` here.
Whether having the ``readme`` suffix for other docs like integration notes is correct - this is another discussion and I agree this could be improved.

mpsl/doc/public_mpsl.rst
b-gent: Let's aim to delete the prefix (both `public` or `external`) before the next release.

mpsl/doc/public_mpsl.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Low priority
+------------
+Low priority is used for background tasks that are not directly tied to the radio or scheduling.
+These tasks are designed in such a way that they can be interrupted by high priority code.
+The tasks are however not designed to be interrupted by other low priority tasks.
+Therefore, make sure that only one MPSL API function is called from the application at any time.
+
+ * All protocol stacks using MPSL must be synchronized (i.e. not called concurrently) to avoid concurrent calls to MPSL functions.
+ * Application must only call MPSL APIs from non-preemptible threads, or with interrupts disabled (e.g. during initialization)
```
b-gent:
```suggestion
 * Application must only call MPSL APIs from non-preemptible threads, or with interrupts disabled (e.g. during initialization).
```

PR #129, opened 2020-01-10
crypto: Automated ECP and MPI limb sizes

doki-nordic:
The ECP/MPI max sizes are calculated based on the asymmetric
cryptography algorithms and enabled ECC curves.

Signed-off-by: Dominik Kilian <Dominik.Kilian@nordicsemi.no>

### DIFF
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -845,25 +845,6 @@ config MBEDTLS_MPI_WINDOW_SIZE
 	  Note that reducing this value might have an impact on the performance.
 	  MBEDTLS_MPI_WINDOW_SIZE setting in mbed TLS config file.
 
-config MBEDTLS_MPI_MAX_SIZE
-	int "MPI - Multiple Precision Integers maximum size"
-	range 0 1024
-	default 1024
-	help
-	  Maximum number of bytes for usable Multiple Precision Integers (MPI) / Bignum.
-	  This will reduce the size of MPIs that can be used for calculation.
-	  Only reduce this value if it is ensured that the system won't need larger numbers.
-	  MBEDTLS_MPI_MAX_SIZE setting in mbed TLS config file.
-
-config MBEDTLS_ECP_MAX_BITS
-	int "ECP - Max bit size of Elliptic Curves"
-	range 0 521
-	default 521
-	help
-	  This setting controls the largest elliptic curve supported in the library.
-	  If only smaller curves are used, then this value can be reduced in order to save memory.
-	  MBEDTLS_ECP_MAX_BITS setting in mbed TLS config file.
-
 config MBEDTLS_ECP_WINDOW_SIZE
 	int "ECP - Elliptic Curve multiplication window size"
 	range 2 6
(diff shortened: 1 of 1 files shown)


nrf_security/Kconfig
```diff
@@ -914,6 +895,36 @@ config MBEDTLS_SSL_CIPHERSUITES
 
 endif # NRF_SECURITY_ADVANCED
 
+config MBEDTLS_MPI_MAX_SIZE
+	int
```
tejlmand: Defined without a prompt means user cannot change the value.

nrf_security/Kconfig
```diff
@@ -914,6 +895,36 @@ config MBEDTLS_SSL_CIPHERSUITES
...
+	default 1024 if MBEDTLS_RSA_C || MBEDTLS_DHM_C || CC310_MBEDTLS_DHM_C || VANILLA_MBEDTLS_DHM_C
+	default 66 if MBEDTLS_ECP_DP_SECP521R1_ENABLED
+	default 64 if MBEDTLS_ECP_DP_BP512R1_ENABLED
+	default 56 if MBEDTLS_ECP_DP_CURVE448_ENABLED
+	default 48 if MBEDTLS_ECP_DP_SECP384R1_ENABLED || MBEDTLS_ECP_DP_BP384R1_ENABLED
+	default 32 if MBEDTLS_ECP_DP_SECP256R1_ENABLED || MBEDTLS_ECP_DP_SECP256K1_ENABLED || MBEDTLS_ECP_DP_BP256R1_ENABLED || MBEDTLS_ECP_DP_CURVE25519_ENABLED
+	default 28 if MBEDTLS_ECP_DP_SECP224R1_ENABLED || MBEDTLS_ECP_DP_SECP224K1_ENABLED
+	default 24 if MBEDTLS_ECP_DP_SECP192R1_ENABLED || MBEDTLS_ECP_DP_SECP192K1_ENABLED
+	default 1024
```
tejlmand: Problem with default, is that it is exactly that, a default value when Kconfig first parses the prj.conf.

This has the following two implications:
1) If application does not enable NRF_SECURITY, then everything is off, and the default value will become `1024`.
if user thereafter enables NRF_SECURITY, vanilla backend, and then selects `MBEDTLS_ECP_DP_SECP192R1_ENABLED`, then the value is still 1024
2) In an app where NRF_SECURITY is enabled by default, vanilla backend is enabled, and `MBEDTLS_ECP_DP_SECP192R1_ENABLED=y`, then default value is 24.
If user then selects `MBEDTLS_ECP_DP_CURVE448_ENABLED=y` afterwards, then ` MBEDTLS_MPI_MAX_SIZE` is still 24, but now it should be 56.

Because of 1) and 2), I believe it is safer to currently keep existing behavior, unless a better approach can be found.

doki-nordic:
@frkv @tejlmand: Do you agree to close this PR without merging it?

frkv:
> @frkv @tejlmand: Do you agree to close this PR without merging it?

Sure. We can close this without merging

tejlmand:
> @frkv @tejlmand: Do you agree to close this PR without merging it?

agree

PR #130, opened 2020-01-13
Update license files

rugeGerritsen:
This aligns all copyright files and bumps them to 2020


mpsl/license.txt
```diff
@@ -0,0 +1,5 @@
+/*
+ * Copyright (c) 2018 - 2020 Nordic Semiconductor ASA
```
tejlmand: Just noted that this license states that it is copyrighted 2018 - 2020.
But the old license, which is now removed, states 2007 - 2019.

I assume, this means the code was originally copyrighted in 2007, and thus I assume the start year should still be 2007.

Or should the start year be the year we changes the license, cause in that case, it should probably just state 2020, as the years 2007 - 2019 was under old license.

2018 - 2020, at least seems incorrect to me.
(Personally I assume 2007 - 2020 to be most correct)

But I'm not a lawyer, so unsure of what is most correct.
Could you check up on this, and also explain why 2018 is start year :)

ble_controller/license.txt
```diff
@@ -0,0 +1,5 @@
+/*
+ * Copyright (c) 2018 - 2020 Nordic Semiconductor ASA
```
tejlmand: Same comment regarding the years as in mpsl/license.txt

mpsl/license.txt
carlescufi: I actually don't know either. I think the correct format is always `<start year> - <end year>` but I doubt that MPSL was introduced 13 years ago?

tejlmand: > but I doubt that MPSL was introduced 13 years ago?

Now that you mention it, I highly doubt that too for mpsl.

Hope that @rugeGerritsen will provide more details on the correct years for this, as well as for the ble controller. The ble controller could actually contain code that originates a long time back.

tejlmand: @rugeGerritsen any updates, been more than a week :wink:

PR #144, opened 2020-02-26
Update mpsl and ble controller

rugeGerritsen:
See changelogs for details


rugeGerritsen:
I had to update the library. Now it also marks the "Write TX Power" as supported when calling "Zephyr Read Supported commands"

PR #148, opened 2020-03-06
nrf_security: crypto improvements post Q1 release

frkv:
- Adds minor changes postponed after Q1 release
- Enabling HMAC_DRBG as a configurable option in nrf_security
- Allowing mbed TLS AES to be configured to use ROM tables to conserve RAM

### DIFF
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -147,17 +147,37 @@ config MBEDTLS_CIPHER_AES_256_OFB_C
 
 comment "Configuration section"
 
-config NRF_SECURITY_RNG
+menuconfig NRF_SECURITY_RNG
 	bool
 	prompt "Random Number Generator support"
 	default y
-	select MBEDTLS_AES_C
 	select MBEDTLS_SHA256_C
 	help
 	  The Random Number Generator support in nRF Security provides a
 	  Pseudorandom Number Generator, PRNG.
 	  The Pseudorandom Number Generator is seeded by a True Random Number
 	  Generator, TRNG, available in hardware.
+
+if NRF_SECURITY_RNG
+
+config MBEDTLS_CTR_DRBG_C
+	bool
+	prompt "PRNG using CTR_DRBG"
+	default y
+	select MBEDTLS_AES_C
+	help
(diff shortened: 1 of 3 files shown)


nrf_security/Kconfig
```diff
@@ -838,6 +838,14 @@ if NRF_SECURITY_ADVANCED
 
 comment "Advanced mbed TLS settings"
 
+config MBEDTLS_AES_ROM_TABLES
+	bool "Use AES ROM tables"
+	depends on VANILLA_MBEDTLS_AES_C || (MBEDTLS_VANILLA_BACKEND && MBEDTLS_AES_C)
+	help
+	  AES lookup tables will be placed in ROM instead of RAM
+	  Placing the AES lookup tables in ROM will perform slower but will reduce RAM usage.
```
tejlmand: Just noticed the mbedTLS config file is very specific regarding RAM savings / ROM usages.

```
 * Tradeoff: Using precomputed ROM tables reduces RAM usage by ~8kb
 * (or ~2kb if \c MBEDTLS_AES_FEWER_TABLES is used) and reduces the
 * initialization time before the first AES operation can be performed.
 * It comes at the cost of additional ~8kb ROM use (resp. ~2kb if \c
 * MBEDTLS_AES_FEWER_TABLES below is used), and potentially degraded
 * performance if ROM access is slower than RAM access.
```
so I think it would be a nice gesture to users to include a line as:
```suggestion
   Placing the AES lookup tables in ROM will perform slower but will reduce RAM usage.
   Using precomputed ROM tables reduces RAM usage by ~8kB with the cost of additional ~8kB ROM use.
```

nrf_security/Kconfig
```diff
@@ -147,18 +147,38 @@ config MBEDTLS_CIPHER_AES_256_OFB_C
 
 comment "Configuration section"
 
-config NRF_SECURITY_RNG
+menuconfig NRF_SECURITY_RNG
 	bool
 	prompt "Random Number Generator support"
 	default y
-	select MBEDTLS_AES_C
```
tejlmand: Why is this no longer needed ?

frkv: AES is not needed for HMAC_DRBG

PR #156, opened 2020-04-17
bsdlib: doc: Adding GNSS conceptual documentation

evenl:
Adding GNSS documentation

Manifest update: https://github.com/nrfconnect/sdk-nrf/pull/2313

### DIFF
--- a/bsdlib/doc/api.rst
+++ b/bsdlib/doc/api.rst
@@ -114,6 +114,8 @@ GNSS socket data frames
    :project: nrfxlib
    :members:
 
+.. _gnss_socket_agps_df:
+
 GNSS socket AGPS data frames
 ============================
 
--- a/bsdlib/doc/extensions.rst
+++ b/bsdlib/doc/extensions.rst
@@ -19,3 +19,4 @@ Extensions are available for:
 
    pdn_extension
    at_commands
+   gnss_extension
--- a/bsdlib/doc/gnss_extension.rst
+++ b/bsdlib/doc/gnss_extension.rst
@@ -1,0 +1,305 @@
+.. _gnss_extension:
+
+GNSS socket
+###########
+
+`Global navigation satellite system (GNSS)`_ socket is one of the socket types supported by the BSD library.
+This socket type is used to configure and fetch GPS position fix data from the GPS module, and to write `A-GPS`_ data to the GPS module.
+
+Creating a GNSS socket
+**********************
+
+The following code shows how to create a GNSS socket with the :c:type:`NRF_PROTO_GNSS` protocol family and the 
+:c:type:`NRF_AF_LOCAL` proprietary address family.
+
+.. code-block:: c
+
+   gnss_fd = nrf_socket(NRF_AF_LOCAL, NRF_SOCK_DGRAM, NRF_PROTO_GNSS);
+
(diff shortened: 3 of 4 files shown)


martinportmann:
Would it be possible to add some documentation to the following GPS options:
- NRF_SO_GNSS_POWER_SAVE_MODE  (What is the difference and tradeoff of the different modes)
- NRF_SO_GNSS_FIX_RETRY (What does this mean?)

Perhaps some limitations of the current implementation would also be interesting:
- no leap seconds reported at the moment (Firmware 1.1.1 with nrfconnect 1.2)

Thanks

evenl:
@martinportmann 
I think a more suitable place for the for the clarification of NRF_SO_GNSS_POWER_SAVE_MODE and NRF_SO_GNSS_FIX_RETRY is in the API documentation it self. But I agree, it's not clear now so that documentation should be updated.

evenl:
@martinportmann 
> no leap seconds reported at the moment (Firmware 1.1.1 with nrfconnect 1.2)
This ended up as a pretty long hunt for what you where talking about, but then we found this post on devzone. https://devzone.nordicsemi.com/f/nordic-q-a/56735/using-gps-and-lte. I assume this is you? You can track the validity of the leap second calculation through the NRF_GNSS_PVT_FLAG_LEAP_SECOND_VALID bit in the flags struct member. But you want the number of leap seconds like in the NMEA data?

martinportmann:
> @martinportmann
> I think a more suitable place for the for the clarification of NRF_SO_GNSS_POWER_SAVE_MODE and NRF_SO_GNSS_FIX_RETRY is in the API documentation it self. But I agree, it's not clear now so that documentation should be updated.



> @martinportmann
> 
> > no leap seconds reported at the moment (Firmware 1.1.1 with nrfconnect 1.2)
> 
> This ended up as a pretty long hunt for what you where talking about, but then we found this post on devzone. https://devzone.nordicsemi.com/f/nordic-q-a/56735/using-gps-and-lte. I assume this is you? You can track the validity of the leap second calculation through the NRF_GNSS_PVT_FLAG_LEAP_SECOND_VALID bit in the flags struct member. But you want the number of leap seconds like in the NMEA data?



> @martinportmann
> 
> > no leap seconds reported at the moment (Firmware 1.1.1 with nrfconnect 1.2)
> 
> This ended up as a pretty long hunt for what you where talking about, but then we found this post on devzone. https://devzone.nordicsemi.com/f/nordic-q-a/56735/using-gps-and-lte. I assume this is you? You can track the validity of the leap second calculation through the NRF_GNSS_PVT_FLAG_LEAP_SECOND_VALID bit in the flags struct member. But you want the number of leap seconds like in the NMEA data?

All these issues are not about documentation and do probably not belong here.

1. With the Firmware 1.1.1 I never did get the NRF_GNSS_PVT_FLAG_LEAP_SECOND_VALID.
2. I would like to know the actual leap second because I have some existing code that works internally with GPS week and TOW.

evenl:
> With the Firmware 1.1.1 I never did get the NRF_GNSS_PVT_FLAG_LEAP_SECOND_VALID.

Tech support can help you with this issue, just register a ticket on devzone.

b-gent:
See my comment here: https://github.com/NordicPlayground/nrfxlib/pull/163#issuecomment-625290416

PR #166, opened 2020-05-05
zboss: Add ZBOSS library

maciekfabia:
Add ZBOSS Zigbee stack version 3.3


zboss/Kconfig
```diff
@@ -0,0 +1,32 @@
...
+	bool "Libraries"
+	help
+	  Use ZBOSS compiled libraries while building Zigbee application
+
+config ZBOSS_LIBRARY_SOURCES
+	bool "Source code"
+	help
+	  Use ZBOSS source code while building Zigbee application
+
+endchoice
```
tejlmand, line 26: This should be adjusted based on the outcome of: `<insert link>`

zboss/Kconfig.sources
```diff
@@ -0,0 +1,313 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+# This file defines options, that are used to set configurable entries of zb_vendor.h and zb_config.h platform files.
+# It will be used only if ZBOSS source code is available.
```
tejlmand, line 7: Should be part of a Zephyr module.

zboss/Kconfig.sources
```diff
@@ -0,0 +1,313 @@
...
+
+menu "ZBOSS features"
+	visible if ZBOSS_LIBRARY_SOURCES
```
tejlmand, line 10: This only remove visibility in limited view in Kconfig.
If user toggles `show-all` in menuconfig, then those settings can still be viewed.

zboss/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+    )
+  # Trim a trailing whitespace.
+  string(STRIP ${ZOI_PLATFORM_SOURCE_DIR} ZOI_PLATFORM_SOURCE_DIR)
+
+  # Include ZBOSS build CMake file.
+  if(EXISTS "${ZOI_PLATFORM_SOURCE_DIR}/CMakeLists.txt")
+    include("${ZOI_PLATFORM_SOURCE_DIR}/CMakeLists.txt")
+  else()
+    message( FATAL_ERROR "Unable to find ZOI platform repository. Please checkout repositories from ZOI manifest file to recover. Exiting." )
+  endif()
```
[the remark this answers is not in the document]
wbober: @tejlmand Only if they are members of ZBOSS Open Initative. What we'd like to achieve is to have the possibility to use the same NCS source for either library or source ZBOSS builds. This is simpler for users and simpler for us (maintenance, CI, etc.). 

I'll let @tomchy comment on the modules feature but as far as I remember we didn't see an elegant way to pull in the ZBOSS sources conditionally, i.e., give the user option to use ZBOSS source and its Kconfig when available. Do you think what you propose will help?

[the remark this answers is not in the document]
tomchy: @tejlmand as you may anticipated, we did not find the solution to make both - conditional `Kconfg` and `CMakeLists.txt` includes. The Zephyr modules feature allows us to do that (the ZOI main manifest repository may be a module), we have missed that.

This leaves us with two remaining issues:
- We still have to provide Kconfig options, that were available during stack compilation, so the common set of headers will select correct ZBOSS defines (please take a look at the `{nrfxlib_zboss}/include/osif/zb_vendor.h` file). This is the reason behind using `visible if ZBOSS_LIBRARY_SOURCES` instead of a simple `if` inside `Kconfig.sources` file. My proposition to tackle this issue is to export relevant Kconfig defines to a separate file as an artifact of ZBOSS build command and add the following line inside nrfxlib to pull it in:
```
toolchain_cc_imacros(${CMAKE_CURRENT_SOURCE_DIR}/include/osif/libzboss_config.h)
```
- We still have to conditionally remove CMakeLists commands under `if (CONFIG_ZBOSS_LIBRARY_PRECOMPILED)`. If we go by the path of removing all ZBOSS-source-related things from nrfxlib, the `Kconfig` file becomes empty and we need to add something to conditionally exclude lines from `CMakeLists.txt`, without Kconfig options. My proposition here is to use the path of the module as the switch:
```
if ((NOT DEFINED ${ZEPHYR_ZOI_MODULE_DIR}) OR
    (CONFIG_ZBOSS_LIBRARY_PRECOMPILED))
```

Additional note: using the Kconfig inside both nrfxlib and defining ZOI module will make the configuration counter-intuitive, with the following structure:
```
Zephyr Kernel Configuration
|
+ Modules. --->
|  |
|  + ZBOSS stack configuration.  --->
|  |
|  + Nordic nrfxlib -->
|  |   |
|  |   + ZBOSS stack configuration.  --->
|  |
|  + Nordic nRF Connect  --->
|     |
|     + Enable Zigbee stack  --->
```
This is the main reason, why I'm voting for making the nrfxlib's `Kconfig` file empty if the source-related options are removed from nrfxlib.

What do you think about my proposals?

zboss/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+zephyr_include_directories(include/ha)
+zephyr_include_directories(include/se)
+zephyr_include_directories(include/osif)
+zephyr_include_directories(include/addons)
+
+nrfxlib_calculate_lib_path(lib_path)
+set(ZBOSS_LIB_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${lib_path})
+if(NOT EXISTS ${ZBOSS_LIB_PATH})
+  message(WARNING "This combination of SoC and floating point ABI is not supported by the ZBOSS lib."
+                  "(${ZBOSS_LIB_PATH} doesn't exist.)")
```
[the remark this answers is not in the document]
tomchy: @tejlmand Do you see any major issues if this is not caught by the Kconfig?

Adding the architecture via `depends on` under the main `ZIGBEE` config will block anyone from compiling the stack for a different architecture/floating point configuration.
If we remove `CONFIG_ZBOSS_LIBRARY_PRECOMPILED` Kconfig option, is there anything we can place under `depends on` to exclude this limitation from source code builds?

zboss/Kconfig.sources
tomchy: The reasoning behind this is explained in: https://github.com/NordicPlayground/nrfxlib/pull/166#discussion_r420679677

zboss/CMakeLists.txt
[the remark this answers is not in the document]
tejlmand: Regarding:
> We still have to provide Kconfig options, that were available during stack compilation, so the common set of headers will select correct ZBOSS defines (please take a look at the {nrfxlib_zboss}/include/osif/zb_vendor.h file). This is the reason behind using visible if ZBOSS_LIBRARY_SOURCES instead of a simple if inside Kconfig.sources file. My proposition to tackle this issue is to export relevant Kconfig defines to a separate file as an artifact of ZBOSS build command and add the following line inside nrfxlib to pull it in:

This appears very similar to a mbedTLS config file.
You could generated this file in same fashion as is done for mbedTLS, based on the Kconfig settings when you compile from source, and provide a static conf file that is part of the nrfxlib.
(i'm under the impression that this header file is static for users using the precompiled lib, correct ?)

zboss/Kconfig
```diff
@@ -0,0 +1,32 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+
+menu "ZBOSS stack configuration."
```
tejlmand, line 7: This is a followup to comment: https://github.com/NordicPlayground/nrfxlib/pull/166/commits/e563e205dc5150a57198b7ba3172246cac1f6cfb#r420684509

Feel free to move the depends to other Kconfig, in case you find it more appropriate to be located elsewhere.

But looking at lib folder, I only find: 
`zboss/lib/cortex-m4/hard-float`
so I would assume something like:
```suggestion
menu "ZBOSS stack configuration."
        depends on CPU_CORTEX_M4 && FPU
```
will prevent the user from selecting Zboss if not using a Cortex-M4, or when using soft-float.

zboss/CMakeLists.txt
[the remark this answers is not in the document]
tejlmand: Placed a comment / suggestion here: https://github.com/NordicPlayground/nrfxlib/pull/166/commits/e563e205dc5150a57198b7ba3172246cac1f6cfb#r421036374

zboss/Kconfig
maciekfabia: thanks for this suggestion,
I will narrow `depends on` to specific SoCs that are currently supported by zigbee: SOC_NRF52840 and (possibly) SOC_NRF52833.
I think Kconfig of zigbee subsys is a good place to add this dependency at `menuconfig ZIGBEE`. It will impact ZBOSS as well because all its options are enclosed with `if ZIGBEE`

Zigbee subsystem pull request:
https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/2208

maciekfabia: I coundn't find `FPU`. Did you mean `CPU_HAS_FPU`?

zboss/CMakeLists.txt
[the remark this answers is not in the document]
maciekfabia: It's certainly a good idea to add `depends on` for specific SoCs.
My follow-up to the suggestion: https://github.com/NordicPlayground/nrfxlib/pull/166#discussion_r421036374

zboss/Kconfig
tejlmand: because I was looking upstream, the right name in our fork is `FLOAT`.
NCS Fork: `FLOAT`
https://github.com/NordicPlayground/fw-nrfconnect-zephyr/blob/master/arch/Kconfig#L509

Zephyr upstream: `FPU`
https://github.com/zephyrproject-rtos/zephyr/blob/master/arch/Kconfig#L525

zboss/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+
+include(${NRFXLIB_DIR}/common.cmake)
+
+#old_sdk: ToDo - Add include path for radio using Kconfig. Now isn't working on NCS master
```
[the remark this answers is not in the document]
rlubos: Please verify if this is actually needed - I've just done a quick test and it seems nRF radio driver header inclusion works correctly on current NCS master.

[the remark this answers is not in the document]
tomchy: Confirmed, fixed, ToDo removed.

zboss/CMakeLists.txt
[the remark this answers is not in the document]
tomchy: `Kconfig` pruned from source-code related options, `zb_vendor.h` left as-is, `libzboss_config.h` generated by the `config_file` command.

zboss/Kconfig
tomchy: Moved to NCS platform repository.

zboss/Kconfig.sources
tomchy: Moved to NCS platform repository.

zboss/Kconfig.sources
tomchy: Changed to regular `if` in NCS platform repository.

zboss/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+
+
+if ((NOT DEFINED ZEPHYR_PLATFORM_NCS_MODULE_DIR) OR
+    (CONFIG_ZBOSS_LIBRARY_PRECOMPILED))
+
+  zephyr_include_directories(include)
+  zephyr_include_directories(include/zcl)
+  zephyr_include_directories(include/ha)
+  zephyr_include_directories(include/se)
+  zephyr_include_directories(include/osif)
```
tejlmand, line 32: This should be in dedicated zephyr library.

As this is also needed in `subsys/zigbee/` because of include directories, a Zephyr interface library can be used to linking to.
For inspiration, look here:
https://github.com/NordicPlayground/fw-nrfconnect-zephyr/blob/master/lib/posix/CMakeLists.txt
https://github.com/zephyrproject-rtos/mbedtls/blob/master/CMakeLists.txt

zboss/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+  nrfxlib_calculate_lib_path(lib_path)
+  set(ZBOSS_LIB_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${lib_path})
+  if(NOT EXISTS ${ZBOSS_LIB_PATH})
+    message(WARNING "This combination of SoC and floating point ABI is not supported by the ZBOSS lib."
+                    "(${ZBOSS_LIB_PATH} doesn't exist.)")
+  endif()
+
+  if (CONFIG_ZIGBEE_ROLE_COORDINATOR OR CONFIG_ZIGBEE_ROLE_ROUTER)
+    zephyr_link_libraries(${ZBOSS_LIB_PATH}/libzboss.a)
+    toolchain_cc_imacros(${CMAKE_CURRENT_SOURCE_DIR}/include/osif/libzboss_config.h)
```
tejlmand, line 43: Why do we want this as a imacro getting included everywhere ?
Should only be included by those that needs it ?
As this is config settings, would it not be possible to set a conf file in the zboss headers, similar to how mbedTLs does things.

The imacro will cause this to be used for all C-files, which should really be for really global things like the autoconf.h

zboss/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+#
+
+include(${NRFXLIB_DIR}/common.cmake)
+
+# Common includes, not related to ZBOSS source code.
+zephyr_include_directories(src/zb_error)
+zephyr_include_directories(include/addons)
+
+# Source files
+zephyr_sources(src/zb_error/zb_error_to_string.c)
```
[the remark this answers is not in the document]
tomchy: The reason is quite simple:
The zephyr_library_* extension is mentioned and documented only in the CMake source code (at least - I cannot find it elsewhere). This makes it hard to find, understand and use. The only examples you may find is to go through other modules and try to find something similar to your needs. It would be great if you can point me to the tutorial on how to use zephyr_library_* extensions.
On the other hand - if you try to use "raw" CMake commands (fall-back to what you can find on the Internet), it will fail to link to the default zephyr target.

This should not be a surprise that, becasue zephyr_sources/include_directories/link_libraries functions are easiest to use, they are the first choice for a developer with small experience in Zephyr's extensions to CMake files.

From what I see - our project needs to define at least three libraries. I hope that this approach will meet your requirements.

zboss/CMakeLists.txt
tomchy: Thank you for the set of inspiring links!

zboss/CMakeLists.txt
tomchy: Well - it replicates exactly what is happening when you have source code with the corresponding Kconfig file, which brings us to a "unified" approach for all build types.

In precompiled builds, those defines are needed by:
 - Zigbe subsys
 - Zigbee applications (those with CONFIG_ZIGBEE == y)
 - Zigbee API test cases (those with CONFIG_ZIGBEE == y)

Additionally, this header must not be included if the ZBOSS source code is available.

If you look at those cases, you should see that the "imacro" approach is the simplest and shortest one you can imagine.

If we need to fix this, than yes, this should be pulled in via the default `#include <zboss_api.h>` directive. For example:
```
// file: zb_vendor.h
#include "libzboss_config_generic.h"
```
and - pulled in in a way, that it gets removed in source code builds:
```
// file: libzboss_config_generic.h.template
#cmakedefine LIBZBOSS_CONFIG_FILE
#ifdef LIBZBOSS_CONFIG_FILE
#include "${LIBZBOSS_CONFIG_FILE}"
#endif
```

What do you think about this approach?

zboss/CMakeLists.txt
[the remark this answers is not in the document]
tomchy: Current proposal can be viewed here:
https://github.com/maciekfabia/nrfxlib/pull/2/files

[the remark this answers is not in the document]
tejlmand: > 
> 
> The reason is quite simple:
> The zephyr_library_* extension is mentioned and documented only in the CMake source code (at least - I cannot find it elsewhere). This makes it hard to find, understand and use. The only examples you may find is to go through other modules and try to find something similar to your needs. It would be great if you can point me to the tutorial on how to use zephyr_library_* extensions.

Agree, and I actually raised this 2 years ago: https://github.com/zephyrproject-rtos/zephyr/issues/9947
seems it was closed (resolved) without being properly resolved.

Reopened now.

> On the other hand - if you try to use "raw" CMake commands (fall-back to what you can find on the Internet), it will fail to link to the default zephyr target.
> 

I agree, which is why I raised this almost 2 years ago:
https://github.com/zephyrproject-rtos/zephyr/issues/8439

but until that is resolved, we have to stick to zephyr cmake wrappers.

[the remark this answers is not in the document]
tejlmand: > 
> 
> Current proposal can be viewed here:
> https://github.com/maciekfabia/nrfxlib/pull/2/files

Good direction 👍 
Placed some comments in there.

zboss/CMakeLists.txt
tejlmand: > If you look at those cases, you should see that the "imacro" approach is the simplest and shortest one you can imagine.

The shortest way is not always the safest ;)
This works when using `gcc`, but there are other toolchains out there which doesn't support imacros.
As well as the previous comment, (and strongest reason), there is no reason to expand this header when compiling Zephyr kernel. 

> What do you think about this approach?

Good direction, but I think can be reduced and still achieve same goal.
Placed some suggestions:
https://github.com/maciekfabia/nrfxlib/pull/2#pullrequestreview-412390777

rlubos:
@tomchy I trust @tejlmand in terms of CMake/Kconfig integration since he has way more knowledge about the build system than I do. IMO once he approves, we could merge this PR (since fw-nrfconnect-nrf` counterpart taht updates the manifest file is also green).

wbober:
@carlescufi Can we merge this please? It would unblock a lot of tasks for us.

rlubos:
@wbober Let's wait until https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/2208 review is finished, so we won't delay the manifest update. There shouldn't be any further delays here, I'll merge as soon as we're done in 2208.

PR #177, opened 2020-05-22
nrf_802154_sl: Add nrf_802154_sl library 0.1

ankuns:
This patch introduces nRF 802.15.4 Service Layer library required for advanced features
of nRF 802.15.4 Radio Driver in versions 2.x

Signed-off-by: Andrzej Kuros <andrzej.kuros@nordicsemi.no>


nrf_802154_sl/include/nrf_802154_sl_utils.h
```diff
@@ -0,0 +1,182 @@
...
+ * @file Definitions of utils used in the nRF 802.15.4 Service Layer.
+ * 
+ */
+
+#ifndef NRF_802154_SL_UTILS_H__
+#define NRF_802154_SL_UTILS_H__
+
+#include <assert.h>
+#include <stdint.h>
+#include "nrf.h"
```
rlubos, line 41: `#include <nrf.h>`

nrf_802154_sl/include/nrf_802154_sl_periphs.h
```diff
@@ -0,0 +1,160 @@
...
+ * @file Module that defines the 802.15.4 driver peripheral usage.
+ *
+ */
+
+#ifndef NRF_802154_SL_PERIPHS_H__
+#define NRF_802154_SL_PERIPHS_H__
+
+#include <nrf.h>
+#include <nrfx.h>
+#include "hal/nrf_ppi.h"
```
rlubos, line 42: `#include <hal/nrf_ppi.h>`

nrf_802154_sl/include/nrf_802154_sl_ant_div.h
```diff
@@ -0,0 +1,348 @@
...
+ * @brief 802.15.4 antenna diversity module.
+ */
+
+#ifndef NRF_802154_SL_ANT_DIV_H
+#define NRF_802154_SL_ANT_DIV_H
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#include "nrf.h"
```
rlubos, line 41: `#include <nrf.h>`

nrf_802154_sl/include/fem/nrf_fem_protocol_api.h
```diff
@@ -0,0 +1,235 @@
...
+ * By default, PA/LNA is automatically deactivated on the radio DISABLED event.
+ * This can be disabled, so that a manual deactivation can be performed instead.
+ */
+
+#ifndef NRF_FEM_PROTOCOL_API_H__
+#define NRF_FEM_PROTOCOL_API_H__
+
+#include "nrf_fem_control_config.h"
+
+#include "drivers/nrfx_errors.h"
```
rlubos, line 29: This and below should use angle brackets.

nrf_802154_sl/include/fem/nrf_fem_control_config.h
```diff
@@ -0,0 +1,63 @@
...
+
+#ifndef NRF_FEM_CONTROL_CONFIG_H_
+#define NRF_FEM_CONTROL_CONFIG_H_
+
+#include "nrf_fem_config.h"
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#include "drivers/nrfx_common.h"
```
rlubos, line 17: This and below should use angle brackets.

nrf_802154_sl/include/fem/nrf_fem_control_config.h
```diff
@@ -0,0 +1,63 @@
...
+ * @brief Configures the PA and LNA device interface.
+ *
+ * This function sets device interface parameters for the PA/LNA module.
+ * The module can then be used to control a power amplifier or a low noise amplifier (or both) through the given interface and resources.
+ *
+ * The function also sets the PPI and GPIOTE channels to be configured for the PA/LNA interface.
+ *
+ * @param[in] p_config Pointer to the interface parameters for the PA/LNA device.
+ *
+ * @retval   ::NRF_SUCCESS                 PA/LNA control successfully configured.
```
rlubos, line 38: `NRFX_SUCCESS`? Same for other error codes in this header.

nrf_802154_sl/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (c) 2019 Nordic Semiconductor
```
rlubos, line 2: 2020

nrf_802154_sl/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+
+# Calculates relative path to the library to be uses taking into account
+# build type of the library, SoC type, floating point ABI, supported FEM type.
+# Note: Subdirectory layout is specific to nrf_802154_sl lib
+function(calculate_lib_path lib_path)
+  # Select build type of the library to link with
+  set(build_type_dir "Release")
```
rlubos, line 12: What about "Debug" libs I've seen earlier? If they're not included, perhaps there's no need for this subdirectory in public repository?

nrf_802154_sl/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+set(NRF_802154_SL_LIB_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${lib_path})
+
+if(NOT EXISTS ${NRF_802154_SL_LIB_PATH})
+  message(WARNING "This combination of SoC and floating point ABI is not supported by the nrf_802154_sl lib."
+                  "(${NRF_802154_SL_LIB_PATH} doesn't exist.)")
+endif()
+
+zephyr_include_directories(include)
+
+zephyr_include_directories(include/fem/three_pin_gpio)
```
rlubos, line 58: Why do we add this and not other variants (simple_gpio)? Couldn't the code use the full path (`fem/xxx/nrf_fem_config.h`)?

rlubos:
I think we need CHANGELOG.rst at very least. Also, would be good to mention commit SHA the library was built from.

nrf_802154_sl/include/fem/nrf_fem_protocol_api.h
hubertmis: It would be against Nordic Firmware Coding Standard rule 4.5.1. Is any conflicting coding standard used in this repository?

rlubos: No, since it's an external project, you can follow Nordic standard, I'm just surprised it's defined like this in there, since CPP docs make it clear, that "" should be rather used for headers located in relative paths to the source file:
https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html
Zephyr follows this convention, but you don't need to comply here. Just ignore those comments.

nrf_802154_sl/CMakeLists.txt
ankuns: Consider if we would like to provide Debug versions to one of our customers. It would be so easy to tell them "copy files to nrfxlib/nrf_802154_sl/lib/Debug" and change build_type_dir to "Debug". That's it. If structure of directories is more flat it will be more difficult.

nrf_802154_sl/CMakeLists.txt
ankuns: Thanks for spotting this. However this is a temporary solution. Module responsible for handling FEMs and number of FEMs supported will change soon. This is simply to make the driver compile and work.

nrf_802154_sl/CMakeLists.txt
rlubos: I'd like to hear from other reviewers, quite unprecedented approach, but I see your motivation now.

nrf_802154_sl/include/fem/nrf_fem_control_config.h
ankuns: Changed to NRFX_SUCCESS

nrf_802154_sl/CMakeLists.txt
ankuns: Fixed

rugeGerritsen:
Should there be a simple README file explaining what this library is?

nrf_802154_sl/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+
+# Calculates relative path to the library to be uses taking into account
+# build type of the library, SoC type, floating point ABI, supported FEM type.
+# Note: Subdirectory layout is specific to nrf_802154_sl lib
+function(calculate_lib_path lib_path)
+  # Select build type of the library to link with
+  set(build_type_dir "Release")
+
+  # Select SoC
+  if(CONFIG_SOC_NRF52840)
```
rugeGerritsen, line 15: What is the reason why we need separate libraries for separate SoCs? Other libraries in this repo fit all SoCs

nrf_802154_sl/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+        set(float_dir softfp)
+    else()
+        assert(0 "Unreachable code")
+    endif()
+  else()
+	  set(float_dir soft)
+  endif()
+
+  # Select FEM type
+  set(fem_type_dir "three_pin_gpio")
```
rugeGerritsen, line 41: Is the intention that there will be a separate library for each FEM?

nrf_802154_sl/Kconfig
```diff
@@ -0,0 +1,26 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+
+if NRF_802154_RADIO_DRIVER
+
+choice NRF_802154_SL_CHOICE
```
rugeGerritsen, line 9: Is the intention that there will different flavors of the NRF_802154_SL?

nrf_802154_sl/CMakeLists.txt
hubertmis: We have some timing-related constants in these libraries (computation duration). The timings differ between SoCs. In the future we plan to replace timing-related constants with TIMER and PPIs. Then we'll deliver like other libraries.

nrf_802154_sl/CMakeLists.txt
rugeGerritsen: For the MPSL and BLE controller libraries we currently do not want to expose the debug libraries. If we want to do that in the future, I'm not sure if it is better to deliver both a debug and release lib, or just providing a debug lib in a separate branch/fork.

nrf_802154_sl/Kconfig
ankuns: Yes. There will be version in upsteram hal_nordic, working with pure zephyr called "open source" or "community". It will have reduced or simplified set of features. Full-featured version will be available as binary in NCS through nrfxlib. Upstream does not allow binaries, we need IP protection, but still want t have upstream version working to some extent.

nrf_802154_sl/CMakeLists.txt
ankuns: Current state of SL is "yes", it has support for given FEM type at compile time. This can be either enabled in runtime (end then given type is used) or disabled. To have other FEM type other build is needed, we have such builds but don't release them at this time. This is not final solution, but we need to release now.

ankuns:
> Should there be a simple README file explaining what this library is?

it is included now

nrf_802154_sl/CMakeLists.txt
hubertmis: We could deliver 802.15.4 "Debug" libraries publicly, or on-demand with NDA or other procedure (it is not decided yet).
The point is that the customer, that would use "Debug" libraries in a project, should have an easy way to switch between Release and Debug versions when building. In my opinion, keeping `Release` and `Debug` directories side by side is the most popular and straightforward way of handling this problem. Switching between repositories or branches would be more error-prone for a customer, what could unnecessarily increase support cost for such customers.
Because of that I would suggest to keep the `Release` directory here to keep space to enable `Debug` builds of the customers projects (regardless distribution method of `Debug` version we choose).

tejlmand: What does `Debug` build mean in this case.
In CMake, the term `CMAKE_BUILD_TYPE` exists, which supports `RELEASE` and `DEBUG`, but this is not used in Zephyr, as it is not compatible with Kconfig.
So in Zephyr we have no notion of Debug vs. Release.

But here we introduce a `Debug` concept foir switching libs, but there is no config option.

I would leave out this for this PR, and if this is a highly needed feature, then let's work on a common approach for a later release.

If you need to give a debig lib to a customer, he can just overwrite the existing lib.
Afterall, with the suggested solution in this PR, he would anyway need make changes to the CMakeLists.txt file.

nrf_802154_sl/CMakeLists.txt
tejlmand: > This is simply to make the driver compile and work. Additionally our tam decided to remove simple_gpio version from the release as there currently no simple way for the user to select implementations.

then I suggest to remove all references to simply_gpio.

nrf_802154_sl/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+  if(CONFIG_SOC_NRF52840)
+    set(soc_dir "nrf52840_xxAA")
+  elseif(CONFIG_SOC_NRF52833)
+    set(soc_dir "nrf52833_xxAA")
+  elseif(CONFIG_SOC_NRF52820)
+    set(soc_dir "nrf52820_xxAA")
+  elseif(CONFIG_SOC_NRF52811)
+    set(soc_dir "nrf52811_xxAA")
+  else()
+    message(WARNING "Selected SoC is not supported by the nrf_802154_sl lib.")
```
tejlmand, line 24: Nothing wrong as an additional safeguard, but I hope the depends in Kconfig are written correctly, so that this situation cannot arise when using `menuconfig`

nrf_802154_sl/CMakeLists.txt
ankuns: Discussed F2F

nrf_802154_sl/CMakeLists.txt
ankuns: Structure flattened Instead of `Debug/${soc_dir}/${float_dir}/${fem_type}`, there is only
`${soc_dir}/${float_dir}`

nrf_802154_sl/Kconfig
```diff
@@ -0,0 +1,26 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+
+if NRF_802154_RADIO_DRIVER
```
tejlmand, line 7: Think this should have an additional level, similar to BLE controller and Bluetooth Link Layer.
image

nrf_802154_sl/README.rst
```diff
@@ -0,0 +1,23 @@
...
+=============
+
+nRF 802.15.4 Service Layer (SL) is a library enabling advanced features of nRF 802.15.4 Radio Driver such as:
+- Multiprotocol support
+- Front-End module support
+- WiFi coexistence support
+- Antenna diversity support
+- Delayed transmission and reception
+- CSMA/CA support
+- Timestamping
```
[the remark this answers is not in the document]
ankuns: Fixed

PR #179, opened 2020-05-22
doc: fix api doc for bsdlib_init return values

b-gent:
Added doxygen group name to api.rst.

Screenshot of output -> 
return_values

The problem is that a new doxygen group was added for these values. With our current doxygen-RST integration, all doxygen groups must be explicitly included in RST, even if they are a subgroup.

Signed-off-by: Bartosz Gentkowski <bartosz.gentkowski@nordicsemi.no>

### DIFF
--- a/bsdlib/doc/api.rst
+++ b/bsdlib/doc/api.rst
@@ -11,6 +11,12 @@ BSD Library Management
    :project: nrfxlib
    :members:
 
+bsd_init() return values for modem firmware updates
+===================================================
+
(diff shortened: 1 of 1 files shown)


tejlmand:
@b-gent do you have a corresponding manifest update for this ?

b-gent:
> @b-gent do you have a corresponding manifest update for this ?

https://github.com/nrfconnect/sdk-nrf/pull/2336

PR #182, opened 2020-05-26
zboss: Link ZBOSS based on Kconfig option

tomchy:
In case APP_LINK_WITH_ZBOSS is not set, none of ZBOSS sources nor libraries should be included in the final application.

Signed-off-by: Tomasz Chyrowicz <tomasz.chyrowicz@nordicsemi.no>


tomchy:
@tejlmand Sync PR for (the one for a dessert):
https://github.com/nrfconnect/sdk-nrf/pull/2326

CMakeLists.txt
```diff
@@ -15,4 +15,4 @@ add_subdirectory_ifdef(CONFIG_MPSL                    mpsl)
 add_subdirectory_ifdef(CONFIG_BSD_LIBRARY             bsdlib)
 add_subdirectory_ifdef(CONFIG_NRFXLIB_CRYPTO          crypto)
 add_subdirectory_ifdef(CONFIG_NORDIC_SECURITY_BACKEND nrf_security)
-add_subdirectory_ifdef(CONFIG_ZIGBEE                  zboss)
+add_subdirectory_ifdef(CONFIG_APP_LINK_WITH_ZBOSS     zboss)
```
tejlmand, line 18: Are you sure you want this dependency ?
What if downstream user creates an intermediate Zephyr library, and then links his app to that lib.
Then he won't need to link the app itself with Zboss, and can just link to his intermediate lib.

PR #195, opened 2020-06-03
Fix doxygen warnings

utzig:
This fixes warnings generated by Doxygen, apart from those generated under the `zboss` and `nrf_802154_sl` directories.

### DIFF
--- a/bsdlib/include/bsd_platform.h
+++ b/bsdlib/include/bsd_platform.h
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
  */
 
-/**@file ipc/bsd_platform.h
- *
+/**
+ * @file bsd_platform.h
  * @defgroup bsd_platform_ipc BSD Platform
  * @ingroup bsd_platform
  * @{
--- a/bsdlib/include/nrf_socket.h
+++ b/bsdlib/include/nrf_socket.h
@@ -132,7 +132,7 @@ typedef uint32_t nrf_fd_set;
 
 /**@brief
  * Socket option to set role for the connection.
- * Accepts a @ref nrf_sec_role_t with values:
+ * Accepts an nrf_sec_role_t with values:
  *  - 0 - Client role.
  *  - 1 - Server role.
  */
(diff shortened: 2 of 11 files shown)


utzig:
> The change in 802.15.4 SL files looks OK to me. However, these headers are API copied from another repository from specified commit.
> Is it OK to modify files copied from other projects? It creates inconsistency between files in this repository and files pointed by the commit in the source repository.

Would be better to fix there first; I didn't know it was copied from somewhere else, could you point me to the originating repo?

hubertmis:
> could you point me to the originating repo?

It's actually a private NordicSemi repository. We have not publicly launch it yet.
I expect that most of the libraries present in sdk-nrfxlib repository are maintained in some private repositories and deployed here. I think that my comment regarding 802.15.4 SL applies to most of the files modified with this patch.

I'm just not sure what is the correct way to handle it, we've just added 802.15.4 SL to this repo last week and I don't know what is the expected workflow.

bsdlib/include/bsd_platform.h
```diff
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
  */
 
-/**@file ipc/bsd_platform.h
- *
+/**
+ * @file bsdlib/include/bsd_platform.h
```
joerchan, line 8: Maybe just bsd_platform.h ?

bsdlib/include/nrf_socket.h
```diff
@@ -132,7 +132,7 @@ typedef uint32_t nrf_fd_set;
 
 /**@brief
  * Socket option to set role for the connection.
- * Accepts a @ref nrf_sec_role_t with values:
+ * Accepts a nrf_sec_role_t with values:
```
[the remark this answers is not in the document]
utzig: Done

bsdlib/include/bsd_platform.h
utzig: Fixed

utzig:
Rebased on latest `master` and tested (with a few extra issues addressed).

utzig:
Complimentary testing PR: https://github.com/nrfconnect/sdk-nrf/pull/2448

hubertmis:
@joerchan , 802.15.4 related files are removed from this repository for now. I think my review is not needed.

PR #197, opened 2020-06-03
Fix antenna diversity in sl

JakubPegza:
Fixes antenna diversity hardware initialization and implements
notifications for energy detection procedure, allowing energy detection
to work properly with antenna diversity in automatic rx mode.

KRKNWK-6264
KRKNWK-6270


rlubos:
Please:
* prefix the commit title with `nrf_802154_sl:`, you can include the jira number in the commit description,
* Add `--signoff` entry,
* You include new binaries here, shouldn't the commit reference new library version, SHA it was built from and a new changelog entry? Up to you to manage this information.

hubertmis:
It seems still does not work properly.

PR #200, opened 2020-06-05
nrf_802154_sl: Update nRF 802.15.4 SL to v0.1.2

jciupis:
This patch aligns nRF 802.15.4 SL's public API to use the correct
license.

KRKNWK-6298

Signed-off-by: Ciupis, Jedrzej <jedrzej.ciupis@nordicsemi.no>

### DIFF
--- a/nrf_802154_sl/CHANGELOG.rst
+++ b/nrf_802154_sl/CHANGELOG.rst
@@ -5,6 +5,13 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+nRF 802.15.4 SL 0.1.2
+*********************
+
+Licenses of all public header files aligned to the correct version.
+No changes to the binary files.
+
+
 nRF 802.15.4 SL 0.1.1
 *********************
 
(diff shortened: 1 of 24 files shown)


nrf_802154_sl/CHANGELOG.rst
```diff
@@ -5,6 +5,13 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+nRF 802.15.4 SL 0.1.2
+*********************
+
+Licenses of all public header files aligned to the correct version.
+No changes to the binary files.
+
```
rlubos, line 13: nit: Single empty line is enough.

hubertmis:
@shantha-14 , the header files are copied here from a source repository. In the source repository there is no license reference in the root directory and because of that we keep full license text in the files. We would like to avoid any modifications in the files when integrating the library with NCS. Because of that I would recommend to keep full license text in the include files, like @jciupis proposed in the pull request.

PR #202, opened 2020-06-08
doc: add missing referenced link

utzig:
Build is failing because bsdlib/README.rst is referring to a link which is currently not available in the links.txt file, so add it.


utzig:
Btw, related PR with west.yml changes: https://github.com/nrfconnect/sdk-nrf/pull/2440

PR #212, opened 2020-06-17
zboss: Increase string members in zb_zcl_basic_attrs_ext_t

maciekfabia:
Increase string members in `zb_zcl_basic_attrs_ext_t` struct to have enough space
for string length subfield and maximum allowed string subfield

Signed-off-by: Maciej Fabia <maciej.fabia@nordicsemi.no>


tomchy:
Since this file is not copied from ZBOSS source code repository and is used only by the application code, it is allowed to merge it without updating ZBOSS libraries.

tomchy:
Please create a corresponding PR in the sdk-nrf repository, so we are confident that this change does not break CI.

PR #223, opened 2020-07-14
openthread: add support for OpenThread linking from library

MarekPorwisz:
OpenThread supports certification by inheritance. This allows
application to be Thread certified by linking certified library instead
of building OpenThread from sources.
This change integrates linking Thread from the library instead of
building it from zephyrs port sources. Included libraries are not yet
certified.

Signed-off-by: Marek Porwisz <marek.porwisz@nordicsemi.no>


openthread/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+include(${NRFXLIB_DIR}/common.cmake)
+nrfxlib_calculate_lib_path(lib_path)
+set(OT_LIB_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${lib_path})
+
+if (CONFIG_OPENTHREAD_LIBRARY_1_1)
```
joerchan, line 10: Given that it seems like the setup is preparing for having multiple version of the openthread library available, should we put this version inside a 1.1 folder?

MarekPorwisz: We are indeed considering adding multiple libraries, but all of them would be of the same OpenThread version (1.1) with different feature set. The intent of using the library is to provide pre-certified library that allows faster device certification. OpenThread community plans to introduce 1.2 version soon, so I placed the 1.1 in the name to indicate it is 1.1 compliant.
We'll need to discuss if we plan to provide both 1.1 and 1.2 libraries however I expect the 1.1 version will be dropped when we'll have 1.2 certified libraries.

MarekPorwisz: We've discussed this within the team. 1.2 version when introduced might be 1.1 compliant but this is not 100% confirmed, it is better to create subfolder as you suggested just in case.

rlubos:
Shouldn't header files for Thread 1.1 be in a separate (v1.1) directory as well? I imagine they might differ for Thread 1.2.

MarekPorwisz:
@b-gent Please have a look at it.

b-gent:
Please check my commit and squash before merging.

PR #232, opened 2020-08-03
nrf_security: Fixing combinatory issues for backend enabled features

frkv:
-Adds mask around backend-specific enabled features
-Fixes issue with dependent defines (like enabling key-exchange)

Ref: NCSDK-5994

------
Manifest update: https://github.com/nrfconnect/sdk-nrf/pull/2664


nrf_security/cmake/kconfig_mbedtls_configure.cmake
```diff
@@ -82,9 +82,10 @@ macro(mbedtls_config_define_depends mbedtls_config)
   set(${mbedtls_config} TRUE)
   foreach(arg ${ARGN})
     if (NOT (CONFIG_${arg} OR
-             CONFIG_CC310_${arg} OR
+             CONFIG_NRF_CRYPTO_BACKEND_COMBINATION_0 AND
+             (CONFIG_CC310_${arg} OR
              CONFIG_OBERON_${arg} OR
-             CONFIG_VANILLA_${arg}))
+             CONFIG_VANILLA_${arg})))
```
tejlmand, line 88: If any of: `CONFIG_CC310_${arg}`, `CONFIG_OBERON_${arg}`, or `CONFIG_VANILLA_${arg}` can be set when `CONFIG_NRF_CRYPTO_BACKEND_COMBINATION_0` is false, then we have an error somewhere in our Kconfig, and that error should be fixed.

frkv: No. That was the case from before, but we added the concept of "shared" code to ensure all features are available even though a backend that doesn't give access to the feature. An example of this is SHA-512 which is only available from the vanilla backend.

Instead of creating a duplicate cmake variable to address the need to enable compilation of the sha512.c file, we went for an approach that uses just one single config-variable. 

This is the main reason for this issue and th, which the additional call to test CONFIG_NRF_CRYPTO_BACKEND_COMBINATION_0 fixes

nrf_security/src/mbedtls/shared/CMakeLists.txt
```diff
@@ -36,21 +20,41 @@ add_library(mbedcrypto_shared STATIC
...
+    ${ARM_MBEDTLS_PATH}/library/chachapoly.c
+  )
+endif()
+
+if (CONFIG_VANILLA_MBEDTLS_RSA_C OR
+   (CONFIG_MBEDTLS_RSA_C AND NOT CONFIG_CC310_MBEDTLS_RSA_C))
+
+target_sources(mbedcrypto_shared PRIVATE
+  ${ARM_MBEDTLS_PATH}/library/rsa.c
+  ${ARM_MBEDTLS_PATH}/library/rsa_internal.c
```
tejlmand, line 53: Wrong indentation on those lines.

nrf_security/src/mbedtls/shared/CMakeLists.txt
```diff
@@ -36,21 +20,41 @@ add_library(mbedcrypto_shared STATIC
...
-)
-target_sources_ifdef(CONFIG_VANILLA_MBEDTLS_CHACHAPOLY_C
-  mbedcrypto_shared PRIVATE ${ARM_MBEDTLS_PATH}/library/chachapoly.c
-)
-target_sources_ifdef(CONFIG_VANILLA_MBEDTLS_RSA_C
-  mbedcrypto_shared PRIVATE ${ARM_MBEDTLS_PATH}/library/rsa.c
+if (CONFIG_VANILLA_MBEDTLS_CHACHA20_C OR
+  (CONFIG_MBEDTLS_CHACHA20_C AND NOT CONFIG_CC310_MBEDTLS_CHACHA20_C))
+
+  target_sources( mbedcrypto_shared PRIVATE
```
tejlmand, line 26:
```suggestion
  target_sources(mbedcrypto_shared PRIVATE
```

nrf_security/src/mbedtls/shared/CMakeLists.txt
frkv: Fixed

nrf_security/src/mbedtls/shared/CMakeLists.txt
frkv: Fixed

nrf_security/cmake/kconfig_mbedtls_configure.cmake
frkv: We found a better scheme than the proposed method which didn't misrepresent the kconfig variables

nrf_security/Kconfig
```diff
@@ -861,7 +861,7 @@ config MBEDTLS_ECJPAKE_C
 	select CC310_MBEDTLS_ECJPAKE_C if CC310_MBEDTLS_ECP_C
 	select OBERON_MBEDTLS_ECJPAKE_C if OBERON_MBEDTLS_ECP_C
 	select VANILLA_MBEDTLS_ECJPAKE_C if VANILLA_MBEDTLS_ECP_C
-	default y
+	default n
```
tejlmand, line 864:
```suggestion
```

No need to specify `default n` as that is the default when nothing is specified.

So line can be removed.

nrf_security/src/mbedtls/shared/CMakeLists.txt
```diff
@@ -36,25 +20,54 @@ add_library(mbedcrypto_shared STATIC
...
+  )
+
+endif()
+
 target_sources_ifdef(CONFIG_MBEDTLS_SHA512_C
   mbedcrypto_shared PRIVATE ${ARM_MBEDTLS_PATH}/library/sha512.c
 )
 
+# Special case: Oberon only backend and enabled DHM
+if (CONFIG_MBEDTLS_DHM_C AND NOT NRF_CRYPTO_BACKEND_COMBINATION_0 AND NOT CONFIG_CC310_MBEDTLS_DHM_C)
```
tejlmand, line 62: Based on your comment, wouldn't this be a little more correct ?
```suggestion
if (CONFIG_MBEDTLS_DHM_C AND CONFIG_OBERON_BACKEND)
```

after all, `CONFIG_CC310_MBEDTLS_DHM_C` can only be true if `NRF_CRYPTO_BACKEND_COMBINATION_0` is also true.

https://github.com/nrfconnect/sdk-nrfxlib/blob/master/nrf_security/Kconfig#L682-L711

Secondly, `CONFIG_MBEDTLS_DHM_C` cannot be true, if `NRF_CRYPTO_BACKEND_COMBINATION_0` is also true:
https://github.com/nrfconnect/sdk-nrfxlib/blob/master/nrf_security/Kconfig#L671-L680
specifically because of:
https://github.com/nrfconnect/sdk-nrfxlib/blob/master/nrf_security/Kconfig#L675

nrf_security/src/mbedtls/shared/CMakeLists.txt
```diff
@@ -36,25 +20,54 @@ add_library(mbedcrypto_shared STATIC
...
+if (CONFIG_VANILLA_MBEDTLS_CHACHAPOLY_C OR
+  (CONFIG_MBEDTLS_CHACHAPOLY_C AND NOT CONFIG_CC310_MBEDTLS_CHACHAPOLY_C))
+
+  target_sources(mbedcrypto_shared PRIVATE
+    ${ARM_MBEDTLS_PATH}/library/chachapoly.c
+  )
+endif()
+
+if (CONFIG_VANILLA_MBEDTLS_RSA_C OR
+   (CONFIG_MBEDTLS_RSA_C AND NOT CONFIG_CC310_MBEDTLS_RSA_C))
```
tejlmand, line 48: If you have single backend CC310 and enable `MBEDTLS_RSA_C` then those files gets added, because `CONFIG_CC310_MBEDTLS_RSA_C` is only true if there are multiple backends.
https://github.com/nrfconnect/sdk-nrfxlib/blob/master/nrf_security/Kconfig#L973-L1003

Is that intended ?

Or would this be more correct:
```suggestion
if (CONFIG_VANILLA_MBEDTLS_RSA_C OR
   (CONFIG_MBEDTLS_RSA_C AND NOT CONFIG_CC310_BACKEND))
```

nrf_security/Kconfig
frkv: Fixed

nrf_security/src/mbedtls/shared/CMakeLists.txt
frkv: Changed to CONFIG_MBEDTLS_DHM_C AND CONFIG_OBERON_BACKEND instead

nrf_security/src/mbedtls/shared/CMakeLists.txt
frkv: Unfortunately it is a little bit deeper than that as there is cmake-logic as well. I will keep it as-is for now but this (and other related topics) will be handled in the issue NCSDK-6047

...in which more logic is moved from cmake to Kconfig. One of the effects of this work is that  CONFIG_CC310_MBEDTLS_RSA_C will be set on CC310 as single-backend directly in Kconfig

PR #239, opened 2020-08-21
bsdlib: version 0.7.8

evenl:
Library version: 7ed4e46cffa093616d8bbb1d0f3535d42b4aee69

* Fix issue when modem communication would not work after a shutdown-init sequence.
* Added TLS server support

Manifest update PR: https://github.com/nrfconnect/sdk-nrf/pull/2748


rlubos (approved):

bsdlib/doc/CHANGELOG.rst
```diff
@@ -5,6 +5,13 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+bsdlib 0.7.8
+************
+
+* Changed UDP sendto retval from ENETDOWN to EBADF in some cases when the socket is no longer valid.
+* Fix issue when modem communication would not work after a shutdown-init sequence.
```
umapraseeda, line 12:
```suggestion
* Fixed the issue where the modem communication would not work after a shutdown-init sequence.
```

umapraseeda (approved):
LGTM

eibe85 (approved):

lemrey (changes requested):
We should not return `EBADF` if the socket descriptor was valid. If `socket()` returned a valid descriptor and suddenly it's no longer valid, the application should know what happened to it so that it can the proper steps before attempting to re-open it. In general the descriptor was valid and the application made no mistake in passing it to the API (as opposed to what `EBADF` means, imo).

What situations are we discussing here specifically, i.e. when is a socket no longer valid and we return `EBADF`?

evenl:
This particular situation is on an open UDP socket when the network do an de-attach and then a re-attach. The IP is now changed and the modem closes the socket automatically. So from the modem side, this socket is invalid. However when I look at the code again all errno codes from the modem for UDP sockets will be converted to BADF. I think we will revert this change and rethink what we should do in this case. I will update the artifact to not contain this change.

evenl:
Removed EBADF change from the list of changes in this PR.

lemrey (approved):
Thanks

PR #244, opened 2020-08-26
openthread: add Nordic library features sets

edmont:
This commit adds a Kconfig choice for selecting some predefined Thread stack features set which will be used for Thread Certification of libraries later on.

Pulled by https://github.com/nrfconnect/sdk-nrf/pull/2707
KRKNWK-7073

Signed-off-by: Eduardo Montoya <eduardo.montoya@nordicsemi.no>


rlubos (approved):
LGTM, just please mention the SHA the libraries were built from in the commit message as well, it will be easier to find this information just looking at the git log.

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
 
 endchoice
 
+# OpenThread configuration features sets from Nordic Semiconductor
+
+choice
```
MarekPorwisz, line 31: I don't quite get how does it work. We set some predefined values, which is ok and seemed to be missing as those are no longer configurable when building with library, but I can't find any connection with the libraries being selected by linker. For example I was expecting that OPENTHREAD_NORDIC_LIBRARY_MTD will be selected automatically when CONFIG_OPENTHREAD_MTD is selected (or the other way around i.e. select CONFIG_OPENTHREAD_MTD when OPENTHREAD_NORDIC_LIBRARY_MTD). Currently it is possible to select OPENTHREAD_FTD and OPENTHREAD_NORDIC_LIBRARY_MTD at the same time.

MarekPorwisz (commented):
Looks ok, but I have a feeling there is something missing - see my comment to Kconfig

openthread/Kconfig
edmont: Yes, that's possible, but those are different things:
- `OPENTHREAD_FTD/MTD` is an OpenThread stack configuration that defines the device's possible roles.
- `OPENTHREAD_NORDIC_LIBRARY_MASTER/FTD/MTD` are sets of predefined extra configurations for OpenThread intended to be certified as a pack in order to ease certification by inheritance for Nordic customers. The sets were already present in nRF5 SDK: https://infocenter.nordicsemi.com/index.jsp?topic=%2Fsdk_tz_v4.1.0%2Fthread_openthread_libraries.html

Actually if you look at the table it's not so strange to use `OPENTHREAD_MTD` with `OPENTHREAD_NORDIC_LIBRARY_FTD`, there could be some use case of an MTD needing to announce a Thread Service, for instance.

KRKNWK-6714 intends to update this in order to have a matching library for each of the `OPENTHREAD_NORDIC_LIBRARY_MASTER/FTD/MTD` options.

MarekPorwisz (approved):

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+	prompt "Nordic library features set"
+	default OPENTHREAD_NORDIC_LIBRARY_NONE
+	help
+	  This option selects the Thread stack features set used for building the precompiled libraries.
+
+config OPENTHREAD_NORDIC_LIBRARY_NONE
+	bool "None"
+	help
+	  Do not enable any specific set of OpenThread features.
+config OPENTHREAD_NORDIC_LIBRARY_MASTER
```
tejlmand, line 41: Please add a blank line above.

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+	select OPENTHREAD_JOINER
+	select OPENTHREAD_MAC_FILTER
+	select OPENTHREAD_MTD_NETDIAG
+	select OPENTHREAD_RAW
+	select OPENTHREAD_SLAAC
+	select OPENTHREAD_SNTP_CLIENT
+	select OPENTHREAD_UDP_FORWARD
+	help
+	  Enable the complete set of OpenThread features.
+config OPENTHREAD_NORDIC_LIBRARY_FTD
```
tejlmand, line 66: Please add a blank line above.

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+	select OPENTHREAD_IP6_FRAGM
+	select OPENTHREAD_JAM_DETECTION
+	select OPENTHREAD_JOINER
+	select OPENTHREAD_MAC_FILTER
+	select OPENTHREAD_SLAAC
+	select OPENTHREAD_SNTP_CLIENT
+	select OPENTHREAD_UDP_FORWARD
+	help
+	  Enable optimized OpenThread features for FTD
+config OPENTHREAD_NORDIC_LIBRARY_MTD
```
tejlmand, line 84: Please add a blank line above.

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+choice
+	prompt "Nordic library features set"
+	default OPENTHREAD_NORDIC_LIBRARY_NONE
+	help
+	  This option selects the Thread stack features set used for building the precompiled libraries.
+
+config OPENTHREAD_NORDIC_LIBRARY_NONE
+	bool "None"
+	help
+	  Do not enable any specific set of OpenThread features.
```
tejlmand, line 40: Would this be more of `Custom` configuration ?
```suggestion
config OPENTHREAD_NORDIC_LIBRARY_CUSTOM
	bool "Custom OpenThread configuration"
	help
	  Custom configuration of OpenThread features.
```

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
 
 endchoice
 
+# OpenThread configuration features sets from Nordic Semiconductor
+
+choice
+	prompt "Nordic library features set"
+	default OPENTHREAD_NORDIC_LIBRARY_NONE
+	help
+	  This option selects the Thread stack features set used for building the precompiled libraries.
```
tejlmand, line 35: How about adding Nordic Semiconductor in here.
It's a help text, so better to be a little more informative.
```suggestion
  Nordic Semiconductor features set selection for OpenThread stack.
```

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+	select OPENTHREAD_ENABLE_SERVICE
+	select OPENTHREAD_IP6_FRAGM
+	select OPENTHREAD_JAM_DETECTION
+	select OPENTHREAD_JOINER
+	select OPENTHREAD_MAC_FILTER
+	select OPENTHREAD_SLAAC
+	select OPENTHREAD_SNTP_CLIENT
+	select OPENTHREAD_UDP_FORWARD
+	help
+	  Enable optimized OpenThread features for FTD
```
tejlmand, line 83:
```suggestion
	  Nordic Semiconductor configuration settings for OpenThread FTD.
```

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+	select OPENTHREAD_DHCP6_CLIENT
+	select OPENTHREAD_ECDSA
+	select OPENTHREAD_IP6_FRAGM
+	select OPENTHREAD_JAM_DETECTION
+	select OPENTHREAD_JOINER
+	select OPENTHREAD_MAC_FILTER
+	select OPENTHREAD_SLAAC
+	select OPENTHREAD_SNTP_CLIENT
+	help
+	  Enable optimized OpenThread features for MTD.
```
tejlmand, line 99:
```suggestion
	  Nordic Semiconductor configuration settings for OpenThread MTD.
```

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+	select OPENTHREAD_MAC_FILTER
+	select OPENTHREAD_MTD_NETDIAG
+	select OPENTHREAD_RAW
+	select OPENTHREAD_SLAAC
+	select OPENTHREAD_SNTP_CLIENT
+	select OPENTHREAD_UDP_FORWARD
+	help
+	  Enable the complete set of OpenThread features.
+config OPENTHREAD_NORDIC_LIBRARY_FTD
+	bool "Optimized FTD set"
```
tejlmand, line 67:
```suggestion
	bool "Nordic Optimized FTD configuration"
```

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+	select OPENTHREAD_JAM_DETECTION
+	select OPENTHREAD_JOINER
+	select OPENTHREAD_MAC_FILTER
+	select OPENTHREAD_MTD_NETDIAG
+	select OPENTHREAD_RAW
+	select OPENTHREAD_SLAAC
+	select OPENTHREAD_SNTP_CLIENT
+	select OPENTHREAD_UDP_FORWARD
+	help
+	  Enable the complete set of OpenThread features.
```
tejlmand, line 65:
```suggestion
	Nordic Semiconductor configuration settings for OpenThread master usage.
	This will enable all features in the OpenThread stack.
```

openthread/Kconfig
```diff
@@ -24,4 +26,79 @@ config OPENTHREAD_LIBRARY_1_1
...
+	default OPENTHREAD_NORDIC_LIBRARY_NONE
+	help
+	  This option selects the Thread stack features set used for building the precompiled libraries.
+
+config OPENTHREAD_NORDIC_LIBRARY_NONE
+	bool "None"
+	help
+	  Do not enable any specific set of OpenThread features.
+config OPENTHREAD_NORDIC_LIBRARY_MASTER
+	bool "Master set"
```
tejlmand, line 42:
```suggestion
	bool "Nordic OpenThread Master configuration"
```

tejlmand (changes requested):
Much better location.
A couple of small suggestions, nothing major.

openthread/Kconfig
edmont: OpenThread has more features that are not enabled in these sets. Better to avoid saying _all_.

tejlmand: Good point, thanks for the updated version.

openthread/Kconfig
```diff
@@ -24,4 +26,82 @@ config OPENTHREAD_LIBRARY_1_1
 
 endchoice
 
+# OpenThread configuration features sets from Nordic Semiconductor
+
+choice
+	prompt "Nordic library features set"
```
tejlmand, line 32: Current 
--------

This looks as:
image
and
image

Suggestion
-------

How about making it a config, and when enabled, then choose the configuration, like this instead:
image

and when selected:
image

and selection of configuration:
image

That way there is no need for a custom, and it is very easy to see that when a Nordic configuration is used.

Diff: 
```
index 1e5c309..4f09ee2 100644
--- a/openthread/Kconfig
+++ b/openthread/Kconfig
@@ -28,17 +28,17 @@ endchoice
 
 # OpenThread configuration features sets from Nordic Semiconductor
 
+config OPENTHREAD_NORDIC_LIBRARY
+       bool
+       prompt "Nordic library feature sets"
+
 choice
-       prompt "Nordic library features set"
-       default OPENTHREAD_NORDIC_LIBRARY_CUSTOM
+       prompt "Feature set"
+       default OPENTHREAD_NORDIC_LIBRARY_MASTER
+       depends on OPENTHREAD_NORDIC_LIBRARY
        help
          Nordic Semiconductor features set selection for OpenThread stack.
 
-config OPENTHREAD_NORDIC_LIBRARY_CUSTOM
-       bool "Custom OpenThread configuration"
-       help
-         Custom configuration of OpenThread features.
-
 config OPENTHREAD_NORDIC_LIBRARY_MASTER
        bool "Nordic OpenThread master configuration"
        select OPENTHREAD_BORDER_AGENT
```
```

**Review by tejlmand — CHANGES_REQUESTED** [review]
Small suggestion on the menu layout.

File: openthread/Kconfig (hunk 12 above)
Comment (edmont) [reply openthread/Kconfig:32]: Yes, good idea.

File: openthread/Kconfig [hunk 13]
```diff
@@ -4,14 +4,16 @@
 # SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
 #
 
+if NET_L2_OPENTHREAD
```
tejlmand, line 7: Nothing wrong, nut I would like to understand the difference between this Kconfig file, and this:
https://github.com/nrfconnect/sdk-nrf/blob/master/subsys/net/openthread/Kconfig

What goes where, and why ?

tejlmand (approved):
Thanks for this.

openthread/Kconfig
edmont: I reverted these changes since they forced to set `OPENTHREAD_NORDIC_LIBRARY=y` in the projects.

Setting `OPENTHREAD_NORDIC_LIBRARY_MASTER=y` alone would result in a warning and none of that configuration being built.

edmont:
Updated the libraries with mention to the used compiler.

openthread/Kconfig
tejlmand: Not sure I understand that argument.
Now users can do:
```
OPENTHREAD_NORDIC_LIBRARY=y
```
and then `OPENTHREAD_NORDIC_LIBRARY_MASTER=y` will be set per default, because of:
```
choice
       prompt "Feature set"
       default OPENTHREAD_NORDIC_LIBRARY_MASTER
```

tejlmand (changes requested):
need some clarification on latest changes.

openthread/Kconfig
edmont: Yes, but that doesn't seem clean to me, normally it would require two settings:

```
OPENTHREAD_NORDIC_LIBRARY=y
OPENTHREAD_NORDIC_LIBRARY_FTD=y
```

Is this acceptable?

tejlmand: > Is this acceptable?

I would say so.
That is aligned with, for example Bluetooth:
```
CONFIG_BT=y
CONFIG_BT_CENTRAL=y
```
although Bluetooth does default to Nordic softdevice, when using NCS, so that `BT_LL_SOFTDEVICE` is not required above.
```
choice BT_LL_CHOICE
...
	default BT_LL_SOFTDEVICE
```
https://github.com/nrfconnect/sdk-nrfxlib/blob/master/softdevice_controller/Kconfig#L29

It could also make sense to have
```
config OPENTHREAD_NORDIC_LIBRARY
       bool
       prompt "Nordic library feature sets"
       default y
```
but I didn't propose it initially, as I don't know if you default prefer users to use custom or a nordic reference conf when using NCS.

That way, `OPENTHREAD_NORDIC_LIBRARY_FTD=y` would be sufficient in a conf file.
But I see no problem in specifying both conf settings.

edmont: OK, reverting back the changes and updating the master PR.

edmont:
Just updated the libraries with the missing `CONFIG_OPENTHREAD_NORDIC_LIBRARY=y`.

tejlmand (approved):

LuDuda (approved):

edmont:
Updated libraries with mbedTLS support for nRF833.

PR #245, opened 2020-08-26
crypto: nrf_oberon v3.0.6

frkv:
Adds nrf_oberon v3.0.6 with bugfixes:

-Adds alignment for mbed-crypto SHA-1/SHA-256 context
-Add missing type in ecjpake context for TLS
-Adding missing symbol for hkdf/hmac SHA-1

Manifest update: https://github.com/nrfconnect/sdk-nrf/pull/2780


crypto/nrf_oberon/include/mbedtls/ecjpake_alt.h
```diff
@@ -33,6 +33,7 @@ extern "C" {
 /* @brief Oberon replacement ECJPAKE context */
 typedef struct mbedtls_ecjpake_context {
     mbedtls_ecjpake_role role;                                  //!< Role, either client or server.
+    int point_format;                                           //!< Point format
```
torsteingrindvik, line 36: This comment does not add value.

crypto/nrf_oberon/include/ocrypto_aes_key.h
```diff
@@ -23,9 +23,9 @@
 extern "C" {
 #endif
 
-#define ocrypto_aes128_KEY_BYTES ((size_t) 16)  //!< AES-128 key size in bytes.
-#define ocrypto_aes192_KEY_BYTES ((size_t) 24)  //!< AES-192 key size in bytes.
-#define ocrypto_aes256_KEY_BYTES ((size_t) 32)  //!< AES-256 key size in bytes.
+#define ocrypto_aes128_KEY_BYTES ((size_t) 16)  //!< AES-128 key size in bytes
```
torsteingrindvik, line 26: FYI the punctuation-at-end might lack consistency (I don't mind either way).

torsteingrindvik (approved):

tejlmand (approved):

nrf_security/Kconfig
```diff
@@ -1108,7 +1108,6 @@ comment "Advanced configuration section"
 menuconfig NRF_SECURITY_ADVANCED
 	bool
 	prompt "Advanced mbed TLS Configuration Settings"
-	depends on MBEDTLS_VANILLA_BACKEND
```
tejlmand, line 1110: Should those be exposed for all backends now ?
Are all advanced option possibilities valid for cc310 and Oberon, or should there maybe still be some filtering with `depends on` ?

nrf_security/src/mbedtls/CMakeLists.txt
```diff
@@ -135,6 +135,7 @@ if (CONFIG_ENTROPY_CC310)
   zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/cc310/replacements/entropy.c)
 elseif(CONFIG_ENTROPY_NRF5_RNG)
   zephyr_library_sources(${NRF_SECURITY_ROOT}/src/backend/nrf5x/entropy_nrf5x.c)
+  zephyr_library_sources(${ARM_MBEDTLS_PATH}/library/entropy.c)
```
tejlmand, line 138: > -Adding building of entropy.c as OpenThread require to use the APIs

can you elaborate ?
Why is it only required in OpenThread case?

tejlmand (commented):
some comments on latest changes.

nrf_security/Kconfig
frkv: There are filtering on all items that are limited to the vanilla backend already....
But we need to have a way to set the supportes ciphersuites

nrf_security/src/mbedtls/CMakeLists.txt
frkv: They utilize a mechanism for getting entropy that is different from the rest of Zephyr (as they interface more directly to mbed TLS). They need access to the APIs directly, and these APIs haven't been made available otherwise

tejlmand (approved):

crypto/nrf_oberon/include/ocrypto_aes_key.h
frkv: The //!< is for headers, they correspond to function/macro briefs which will be lead to their labels being a top-level grouping, hence they should not have punctuation.

crypto/nrf_oberon/include/mbedtls/ecjpake_alt.h
frkv: It will at least be added into the documentation. The rest of the context is by design opaque, but the actual addressable types. Will add changes to the later 3.0.7 version (because building and testing this library again, takes an hour of runtime)

torsteingrindvik (approved):

PR #246, opened 2020-08-26
cmake: nrf_security: use same python as rest of build system for west

tejlmand:
This commit aligns the usage of west to upstream Zephyr, where Python
is used directly to ensure the same python is used for west.

This updates nrfxlib to use same principle as here: https://github.com/zephyrproject-rtos/zephyr/pull/26060

-----
Manifest PR: https://github.com/nrfconnect/sdk-nrf/pull/2781


frkv (approved):
LGTM. Python code works

PR #251, opened 2020-08-31
openthread: Add opethread library and source code cmake version checking.

XPAWAN:


openthread/CMakeLists.txt
```diff
@@ -6,6 +6,19 @@
 include(${NRFXLIB_DIR}/common.cmake)
 nrfxlib_calculate_lib_path(lib_path)
 
+file (READ "../../modules/lib/openthread/include/openthread/instance.h" INSTANCE_FILE_CONTENT)
```
rlubos, line 9: This whole check needs to be placed underneath `if (CONFIG_OPENTHREAD_LIBRARY_1_1)`, otherwise the warning will be printed even when the library is not used. Plus I'd consider making a cmake function out of this code (in case it would be needed to verify API version for 1.2 library).

rlubos, line 9: nit: Extra space after `file`

openthread/CMakeLists.txt
```diff
@@ -6,6 +6,19 @@
...
+string(REGEX MATCH "#define[ ]*OPENTHREAD_API_VERSION[ ]*\\([0-9]+\\)" MACRO_DEF "${INSTANCE_FILE_CONTENT}")
+string(REGEX MATCH "[0-9]+" OPENTHREAD_SOURCE_API_VERSION "${MACRO_DEF}")
+
+file (READ "include/openthread/instance.h" INSTANCE_FILE_CONTENT)
```
rlubos, line 13: nit: Extra space after `file`

openthread/CMakeLists.txt
```diff
@@ -6,6 +6,19 @@
...
+string(REGEX MATCH "#define[ ]*OPENTHREAD_API_VERSION[ ]*\\([0-9]+\\)" MACRO_DEF "${INSTANCE_FILE_CONTENT}")
+string(REGEX MATCH "[0-9]+" OPENTHREAD_LIB_API_VERSION "${MACRO_DEF}")
+
+#check if the Openthread API changed
+if ( NOT (${OPENTHREAD_SOURCE_API_VERSION} EQUAL ${OPENTHREAD_LIB_API_VERSION}))
```
rlubos, line 18: nit: Extra space after `if` and `(`

openthread/CMakeLists.txt
```diff
@@ -6,6 +6,19 @@
...
+  message( WARNING "The Openthread Source Code version (${OPENTHREAD_SOURCE_API_VERSION}) is not equal to compiled Openthread library version (${OPENTHREAD_LIB_API_VERSION}).")
```
rlubos, line 19: nit: Extra space after `(`. No need to capitalize `Source Code`.

rlubos (commented):
You need to follow certain rules for the commit message:
https://docs.zephyrproject.org/latest/contribute/index.html#commit-guidelines

In general:
* The title needs `openthread: ` prefix
* The commit needs some description in the body (rationale for this change)
* You need `--signoff` entry.

Please sign the CLA.

edmont:
Would be checking `OPENTHREAD_API_VERSION` enough to assure libraries compatibility? I would say no, since it can happen that the API version remains the same but for instance some configuration switch was missing when building them. This already happened in the past and https://github.com/nrfconnect/sdk-nrfxlib/pull/244/commits/2e1f9a63809a602868f66def5923ed337cdb6e7f was needed to address that issue.

I would say the only safe way would be to allow the usage of libraries only for the specific release they were created for, based on the manifest commit, for example.

openthread/CMakeLists.txt
tejlmand, line 9: Please refer to the OpenThread module dir.
```suggestion
file(READ "${ZEPHYR_OPENTHREAD_MODULE_DIR}/../include/openthread/instance.h" INSTANCE_FILE_CONTENT)
```

openthread/CMakeLists.txt
```diff
@@ -6,6 +6,19 @@
 include(${NRFXLIB_DIR}/common.cmake)
 nrfxlib_calculate_lib_path(lib_path)
 
+file (READ "../../modules/lib/openthread/include/openthread/instance.h" INSTANCE_FILE_CONTENT)
+string(REGEX MATCH "#define[ ]*OPENTHREAD_API_VERSION[ ]*\\([0-9]+\\)" MACRO_DEF "${INSTANCE_FILE_CONTENT}")
+string(REGEX MATCH "[0-9]+" OPENTHREAD_SOURCE_API_VERSION "${MACRO_DEF}")
+
+file (READ "include/openthread/instance.h" INSTANCE_FILE_CONTENT)
+string(REGEX MATCH "#define[ ]*OPENTHREAD_API_VERSION[ ]*\\([0-9]+\\)" MACRO_DEF "${INSTANCE_FILE_CONTENT}")
+string(REGEX MATCH "[0-9]+" OPENTHREAD_LIB_API_VERSION "${MACRO_DEF}")
```
tejlmand, line 15: No need for the second regex, just make the first regex a matched expression, and extract it using `${CMAKE_MATCH_1}`, like this:
```suggestion
string(REGEX MATCH "#define[ ]*OPENTHREAD_API_VERSION[ ]*\\(([0-9]+)\\)" MACRO_DEF "${INSTANCE_FILE_CONTENT}")
set(OPENTHREAD_LIB_API_VERSION ${CMAKE_MATCH_1})
```

openthread/CMakeLists.txt
```diff
@@ -6,6 +6,19 @@
 include(${NRFXLIB_DIR}/common.cmake)
 nrfxlib_calculate_lib_path(lib_path)
 
+file (READ "../../modules/lib/openthread/include/openthread/instance.h" INSTANCE_FILE_CONTENT)
+string(REGEX MATCH "#define[ ]*OPENTHREAD_API_VERSION[ ]*\\([0-9]+\\)" MACRO_DEF "${INSTANCE_FILE_CONTENT}")
+string(REGEX MATCH "[0-9]+" OPENTHREAD_SOURCE_API_VERSION "${MACRO_DEF}")
```
tejlmand, line 11: No need for the second regex, just make the first regex a matched expression, and extract it using `${CMAKE_MATCH_1}`, like this:
```suggestion
string(REGEX MATCH "#define[ ]*OPENTHREAD_API_VERSION[ ]*\\(([0-9]+)\\)" MACRO_DEF "${INSTANCE_FILE_CONTENT}")
set(OPENTHREAD_SOURCE_API_VERSION ${CMAKE_MATCH_1})
```

tejlmand (changes requested):
@XPAWAN Thanks for this.
A couple of small improvement suggestions.

tejlmand:
> Would be checking `OPENTHREAD_API_VERSION` enough to assure libraries compatibility?

@edmont Although staying in the same API version is no guarantee, then having different version is more likely to break compatibility, I assume.
So in that context, this check at least warns the user that the API version are different, and thus a high risk.

It's not stating that identical API versions are safe.

tejlmand (commented):
Changes looks good, thanks :smile: 

Only some small things left.
- Squash the two commits into one:
```
git rebase -i HEAD~2
```
and then change last commit from `pick` to `squash`, like this:
```
pick eae52e1 Add opethread library and source code cmake version checking.                          
squash c63f2b1 Remarks after review 
```

After doing so, then you will also have the opportunity to change the commit message, as suggested here:
https://github.com/nrfconnect/sdk-nrfxlib/pull/251#pullrequestreview-479516133

tejlmand (approved):
Looks good.
Please fix the line length issues in the commit message.

rlubos:
@XPAWAN I've finally got the CI to run properly on this PR.

The commit message needs adjustments, as listed by gitlint (btw there's no need to indent the commit body, it can start at the same level as the title). 

Plus we need a PR in `sdk-nrf` updating [this line](https://github.com/nrfconnect/sdk-nrf/blob/master/west.yml#L100) to `pull/251/head`, the get a full CI run before we can merge.

LuDuda (approved):

MarekPorwisz (approved):

rlubos (approved):

PR #259, opened 2020-09-11
openthread: Remove incorrect setting

MarekPorwisz:
OPENTHREAD_RAW shall not be enabled by default.
This breakes CLI sample and the htread is not able to communicate.

Signed-off-by: Marek Porwisz <marek.porwisz@nordicsemi.no>


rlubos (approved):
Looks good, but please open a manifest update.

MarekPorwisz:
https://github.com/nrfconnect/sdk-nrf/pull/2902

edmont (changes requested):
@MarekPorwisz if we remove the KConfig setting then we should update the README.rst and also the libraries binaries.

But what exactly is the issue? Shouldn't we fix the bug which is making the thread unresponsive? What if some user wants to enable the link raw support?

Notice that the original idea was to make the libraries settings match the existing ones for nRF5 SDK: https://infocenter.nordicsemi.com/topic/sdk_tz_v4.1.0/thread_openthread_libraries.html

MarekPorwisz:
> @MarekPorwisz if we remove the KConfig setting then we should update the README.rst and also the libraries binaries.
> 
> But what exactly is the issue? Shouldn't we fix the bug which is making the thread unresponsive? What if some user wants to enable the link raw support?
> 
> Notice that the original idea was to make the libraries settings match the existing ones for nRF5 SDK: https://infocenter.nordicsemi.com/topic/sdk_tz_v4.1.0/thread_openthread_libraries.html

The libraries itself work with this setting just fine so we may leave it unchanged. Separate task was created to investigate the issue and fix it. The purpose of this PR was to unblock CLI sample.

edmont:
To be fixed by https://github.com/zephyrproject-rtos/zephyr/pull/28402

edmont:
Fixed by https://github.com/nrfconnect/sdk-nrf/pull/2924

PR #260, opened 2020-09-14
openthread: Add install_openthread_libraries target

pszkotak:
With this extension OpenThread libraries will be able to use OpenThread
stack build from source to overwrite the nrfxlib binaries.

Signed-off-by: Piotr Szkotak <piotr.szkotak@nordicsemi.no>

Functionality extracted from: https://github.com/nrfconnect/sdk-nrf/pull/2655

### DIFF
--- a/openthread/CMakeLists.txt
+++ b/openthread/CMakeLists.txt
@@ -6,6 +6,8 @@
 include(${NRFXLIB_DIR}/common.cmake)
 nrfxlib_calculate_lib_path(lib_path)
 
+include(${CMAKE_CURRENT_LIST_DIR}/cmake/extensions.cmake)
+
 function(checkOpenthreadVersion)
   file(READ "${ZEPHYR_OPENTHREAD_MODULE_DIR}/../include/openthread/instance.h" INSTANCE_FILE_CONTENT)
   string(REGEX MATCH "#define[ ]*OPENTHREAD_API_VERSION[ ]*\\(([0-9]+)\\)" MACRO_DEF "${INSTANCE_FILE_CONTENT}")
@@ -20,6 +22,36 @@ function(checkOpenthreadVersion)
     message(WARNING "The Openthread source code version (${OPENTHREAD_SOURCE_API_VERSION}) is not equal to compiled Openthread library version (${OPENTHREAD_LIB_API_VERSION}).")
   endif()
 endfunction()
+
+if(CONFIG_OPENTHREAD_SOURCES)
+  message(DEBUG "Building OT from sources, config file will be generated.")
+
+  set(OPENTHREAD_CONFIG_FILE "${CMAKE_BINARY_DIR}/openthread_lib_configuration.txt")
+
+  openthread_libs_configuration_write(${OPENTHREAD_CONFIG_FILE})
+
+  nrfxlib_calculate_lib_path(lib_path)
+  set(OPENTHREAD_DST_DIR "${CMAKE_CURRENT_LIST_DIR}/${lib_path}/v${CONFIG_OPENTHREAD_THREAD_VERSION}")
+  set(OPENTHREAD_HEADERS_DIR "${ZEPHYR_OPENTHREAD_MODULE_DIR}/../include")
+
+  add_custom_target(install_openthread_libraries
+    COMMAND ${CMAKE_COMMAND} -E make_directory "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:openthread-ftd> "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:openthread-mtd> "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:openthread-radio> "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:openthread-cli-ftd> "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:openthread-cli-mtd> "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:openthread-ncp-ftd> "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:openthread-ncp-mtd> "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:openthread-rcp> "${OPENTHREAD_DST_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy_directory
+      ${OPENTHREAD_HEADERS_DIR}
+      ${NRFXLIB_DIR}/openthread/include
+    COMMAND ${CMAKE_COMMAND} -E copy
+     ${OPENTHREAD_CONFIG_FILE}
+     ${NRFXLIB_DIR}/openthread/
+    )
+endif()
 
 if (CONFIG_OPENTHREAD_LIBRARY_1_1)
   set(version_dir "v1.1")
--- a/openthread/cmake/extensions.cmake
+++ b/openthread/cmake/extensions.cmake
@@ -1,0 +1,32 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+
+# Store the configuration of the compiled OpenThread libraries
+# and set source and destination paths.
+function(openthread_libs_configuration_write CONFIG_FILE)
+  find_package(Git QUIET)
+  if(GIT_FOUND)
+    execute_process(
+      COMMAND           ${GIT_EXECUTABLE} rev-parse HEAD
+      WORKING_DIRECTORY ${ZEPHYR_OPENTHREAD_MODULE_DIR}
+      RESULT_VARIABLE   git_return
+      OUTPUT_VARIABLE   openthread_git_hash
+      )
+endif()
+
+  # Store all OT related variables
+  get_cmake_property(_variableNames VARIABLES)
+  foreach (_variableName ${_variableNames})
+    if("${_variableName}" MATCHES "^CONFIG_OPENTHREAD_.*|^OT_.*")
+      list(APPEND OPENTHREAD_SETTINGS "${_variableName}=${${_variableName}}\n")
+    endif()
+  endforeach()
+
+  list(SORT OPENTHREAD_SETTINGS)
+  list(INSERT OPENTHREAD_SETTINGS 0 "OpenThread_commit=${openthread_git_hash}\n")
+  FILE(WRITE ${CONFIG_FILE} ${OPENTHREAD_SETTINGS})
+
+endfunction(openthread_libs_configuration_write)


openthread/cmake/extensions.cmake
```diff
@@ -0,0 +1,73 @@
+#
```
tejlmand, line 1: Seems you missed this comment:
https://github.com/nrfconnect/sdk-nrf/pull/2655#discussion_r487766990

An extension.cmake file should only contain functions/macros, not general code.

The general code should go to this file:
https://github.com/nrfconnect/sdk-nrfxlib/blob/master/openthread/CMakeLists.txt

openthread/cmake/extensions.cmake
```diff
@@ -0,0 +1,73 @@
...
+# Copyright (c) 2020 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+set(NRFXLIB_DIR ${ZEPHYR_NRFXLIB_MODULE_DIR})
+assert_exists(NRFXLIB_DIR)
```
tejlmand, line 7: You are already in this repo.
Also be aware that this variable is not the one being used inside `openthread_libs_configuration_write()`, cause during execution of `openthread_libs_configuration_write()` then is solely depends on the `NRFXLIB_DIR` value in the scope of the caller.

openthread/cmake/extensions.cmake
```diff
@@ -0,0 +1,73 @@
...
+include(${NRFXLIB_DIR}/common.cmake)
```
tejlmand, line 8: Not needed if this file is included from here: https://github.com/nrfconnect/sdk-nrfxlib/blob/master/openthread/CMakeLists.txt

openthread/cmake/extensions.cmake
```diff
@@ -0,0 +1,73 @@
...
+# and set source and destination paths.
+function(openthread_libs_configuration_write)
+  set(OPENTHREAD_SRC_DIR "${CMAKE_BINARY_DIR}/modules/openthread/build/src"
+    CACHE STRING "Directory containg the OpenThread source code.")
+
+  find_package(Git QUIET)
+  if(GIT_FOUND)
+    execute_process(
+      COMMAND           ${GIT_EXECUTABLE} rev-parse HEAD
+      WORKING_DIRECTORY ${OPENTHREAD_SRC_DIR}
```
tejlmand, line 20: you haven't adressed this comment: https://github.com/nrfconnect/sdk-nrf/pull/2655#discussion_r487778529

tejlmand (changes requested):

openthread/cmake/extensions.cmake
pszkotak: Thanks, fixed.

openthread/cmake/extensions.cmake
pszkotak: Thanks, fixed.

openthread/cmake/extensions.cmake
pszkotak: Thanks, fixed.

openthread/cmake/extensions.cmake
pszkotak: Used `ZEPHYR_OPENTHREAD_MODULE_DIR` instead.

lmaciejonczyk (approved):

lmaciejonczyk:
@tejlmand It looks all your requests were addressed. Can we proceed?

pszkotak:
Rebased.

openthread/cmake/extensions.cmake
```diff
@@ -0,0 +1,41 @@
...
+      RESULT_VARIABLE   git_return
+      OUTPUT_VARIABLE   openthread_git_hash
+      )
+
+    list(APPEND OPENTHREAD_SETTINGS "OpenThread_commit=${openthread_git_hash}\n\n")
+  endif()
+
+  # Store all OT related variables
+  get_cmake_property(_variableNames VARIABLES)
+  list (SORT _variableNames)
```
tejlmand, line 24: Why spend time on sorting, if you loop all the vars anyway ?

openthread/cmake/extensions.cmake
```diff
@@ -0,0 +1,41 @@
...
+  foreach (_variableName ${_variableNames})
+    if("${_variableName}" MATCHES "^CONFIG_OPENTHREAD_.*|^OT_.*")
+      list(APPEND OPENTHREAD_SETTINGS "${_variableName}=${${_variableName}}\n")
+    endif()
+  endforeach()
+
+  set(OPENTHREAD_CONFIG_FILE
+    "${CMAKE_BINARY_DIR}/openthread_lib_configuration.txt"
+    CACHE STRING "File containg OpenThread build configuration parameters.")
```
tejlmand, line 33: The cache issue is still not addressed:
https://github.com/nrfconnect/sdk-nrf/pull/2655#discussion_r486278481

There is no reason to use a cached variable here, remember cached variable lives BETWEEN CMake invocations, and are not updated if already present in the cache.

Instead you can provide the conf file by the caller, and return dst dir.

```
function(openthread_libs_configuration_write CONFIG_FILE DST_DIR)
```
and then use `CONFIG_FILE` inside the function and return dst dir as:
```
set(${DST_DIR} "${ZEPHYR_NRFXLIB_MODULE_DIR}/openthread/${lib_path}v${CONFIG_OPENTHREAD_THREAD_VERSION}"
  PARENT_SCOPE
)

```

and then update ` openthread/CMakeLists.txt`:
```
set(OPENTHREAD_CONFIG_FILE
    "${CMAKE_BINARY_DIR}/openthread_lib_configuration.txt"
)
openthread_libs_configuration_write(${OPENTHREAD_CONFIG_FILE} OPENTHREAD_DST_DIR)
```
inside the function.

You can also decide to return both variables from the function, or return just the OpenThread version from the function (instead of the), and then let the caller construct the dst dir using the returened version value.
Returning just the OpenThread version remove the need for the function to know about directory structures, which generally is nice as it makes the function more generic and reusable for other purposes.

tejlmand (changes requested):
> @tejlmand It looks all your requests were addressed. Can we proceed?

Except the old cache comment in original review.

openthread/cmake/extensions.cmake
tejlmand: actually, the version seems to be a Kconfig setting, so no reason it should be inside the function:
`v${CONFIG_OPENTHREAD_THREAD_VERSION}` <-- Assume this is Kconfig value, correct ?

openthread/cmake/extensions.cmake
pszkotak: Sorted only the filtered part going to the file and then prepended with the commit value which I want to be at the beginning.

openthread/cmake/extensions.cmake
pszkotak: Correct. No cached variables, passing only the `CONFIG_FILE` now.

openthread/cmake/extensions.cmake
```diff
@@ -0,0 +1,34 @@
...
+  # Store all OT related variables
+  get_cmake_property(_variableNames VARIABLES)
+  foreach (_variableName ${_variableNames})
+    if("${_variableName}" MATCHES "^CONFIG_OPENTHREAD_.*|^OT_.*")
+      list(APPEND OPENTHREAD_SETTINGS "${_variableName}=${${_variableName}}\n")
+    endif()
+  endforeach()
+
+  list(SORT OPENTHREAD_SETTINGS)
+  list(PREPEND OPENTHREAD_SETTINGS "OpenThread_commit=${openthread_git_hash}\n")
```
tejlmand, line 29: `list(PREPEND ...)` is not supported by CMake 3.13.
Please use `list(INSERT <list> 0 <element>)` instead:
https://cmake.org/cmake/help/v3.13/command/list.html

```suggestion
  list(INSERT OPENTHREAD_SETTINGS 0 "OpenThread_commit=${openthread_git_hash}\n") 
```

openthread/cmake/extensions.cmake
```diff
@@ -0,0 +1,34 @@
...
+  FILE(WRITE ${CONFIG_FILE} ${OPENTHREAD_SETTINGS})
+
+  nrfxlib_calculate_lib_path(lib_path)
```
tejlmand, line 32: Guess this should be moved, or things will not work.

tejlmand (changes requested):
getting closer to the target.

openthread/cmake/extensions.cmake
pszkotak: Fixed.

openthread/cmake/extensions.cmake
pszkotak: Fixed.

openthread/CMakeLists.txt
```diff
@@ -21,6 +23,36 @@ function(checkOpenthreadVersion)
...
 
+if(CONFIG_OPENTHREAD_SOURCES)
+  message(DEBUG "Building OT from sources, config file will be generated.")
+
+  set(OPENTHREAD_CONFIG_FILE "${CMAKE_BINARY_DIR}/openthread_lib_configuration.txt")
+
+  openthread_libs_configuration_write(${OPENTHREAD_CONFIG_FILE})
+
+  nrfxlib_calculate_lib_path(lib_path)
+  set(OPENTHREAD_DST_DIR "${ZEPHYR_NRFXLIB_MODULE_DIR}/openthread/${lib_path}/v${CONFIG_OPENTHREAD_THREAD_VERSION}")
```
tejlmand, line 34: while we are waiting patiently for CI to work, then you might update this to `${CMAKE_CURRENT_LIST_DIR}`.
We are already inside this repo now, so no need to use the module reference.
```suggestion
  set(OPENTHREAD_DST_DIR "${CMAKE_CURRENT_LIST_DIR}/${lib_path}/v${CONFIG_OPENTHREAD_THREAD_VERSION}")
```

tejlmand (commented):
Looks good.
When looking over everything one last time, then I noticed one var that can be changed.

openthread/CMakeLists.txt
pszkotak: Done.

tejlmand: Next time, you can just hit the `Commit suggestion`, you have an `openthread` too much.

openthread/CMakeLists.txt
```diff
@@ -21,6 +23,36 @@ function(checkOpenthreadVersion)
...
 
+if(CONFIG_OPENTHREAD_SOURCES)
+  message(DEBUG "Building OT from sources, config file will be generated.")
+
+  set(OPENTHREAD_CONFIG_FILE "${CMAKE_BINARY_DIR}/openthread_lib_configuration.txt")
+
+  openthread_libs_configuration_write(${OPENTHREAD_CONFIG_FILE})
+
+  nrfxlib_calculate_lib_path(lib_path)
+  set(OPENTHREAD_DST_DIR "${CMAKE_CURRENT_LIST_DIR}/openthread/${lib_path}/v${CONFIG_OPENTHREAD_THREAD_VERSION}")
```
tejlmand, line 34: This file is:
`<nrfxlib-repo>/openthread/CMakeLists.txt`
therefore `CMAKE_CURRENT_LIST_DIR == <nrfxlib-repo>/openthread` and thus the `openthread` can be removed:

```suggestion
  set(OPENTHREAD_DST_DIR "${CMAKE_CURRENT_LIST_DIR}/${lib_path}/v${CONFIG_OPENTHREAD_THREAD_VERSION}")
```

tejlmand (changes requested):

openthread/CMakeLists.txt
pszkotak: Fixed

tejlmand (approved):
thanks for the patience

PR #242, opened 2020-08-24
cc312 support

frkv:
Initial PR for CC312 support

- CC312 content is experimental
- The binary files may change, as may a lot of content in this PR

ref: NCSDK-4638

manifest PR: https://github.com/nrfconnect/sdk-nrf/pull/2758
Zephyr downstream PR: https://github.com/nrfconnect/sdk-zephyr/pull/346

Frank Audun Kvamtrø <frank.kvamtro@nordicsemi.no>

PR #268, opened 2020-09-22
openthread: Create separate libraries for differend crypto backends

lmaciejonczyk:
OpenThread depends on mbedtls configuration file which is generated by nrf_security based on active crypto backends. This force us to maintain separate prebuild OT libraries. What more, we want to have prebuild separate OT libraries which provide different OT feature option set like we did for nRF5 SDK (master, FDK, MDK).

Adjust install_openthread_libraries target to copy compiled libraries to the right location taking in consideration used nrf_security backend and OpenThread feature option set.
Add new custom target to strip prebuild libraries.
Check OpenThead dependencies when the app links OpenThread prebuild libraries.

- [x] Waits for Upmerge https://github.com/nrfconnect/sdk-nrf/pull/2918
- [x] Needs updates after https://github.com/nrfconnect/sdk-nrfxlib/pull/242 is in


openthread/cmake/extensions.cmake
```diff
@@ -3,20 +3,44 @@
 #
 # SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
 #
+include(${NRFXLIB_DIR}/common.cmake)
+include(${NRFXLIB_DIR}/nrf_security/cmake/extensions.cmake)
+
+# The Build the path of the library taking in nrf_security backend type,
+# OpenThread version and feature set  
+function(openthread_calculate_lib_path ot_version lib_path)
+  if(CONFIG_OPENTHREAD_NORDIC_LIBRARY_MASTER)
```
pszkotak, line 12: These three defines(`CONFIG_OPENTHREAD_NORDIC_LIBRARY_MASTER`, `CONFIG_OPENTHREAD_NORDIC_LIBRARY_FTD`, `CONFIG_OPENTHREAD_NORDIC_LIBRARY_MTD`) are basically used to pick the source/destination library folder depending if we are building from source or libraries.

I think that we should keep in mind to expand the `ug_thread_configuring.rst` as it currently states:
```
* :option:`CONFIG_OPENTHREAD_NORDIC_LIBRARY_MASTER` - Enable the complete set of OpenThread features.
* :option:`CONFIG_OPENTHREAD_NORDIC_LIBRARY_FTD` - Enable optimized OpenThread features for FTD.
* :option:`CONFIG_OPENTHREAD_NORDIC_LIBRARY_MTD` - Enable optimized OpenThread features for MTD.
```
Which is true only for building from library. When building libraries user will just overwrite selected libraries with whatever configuration he already has. I think that while reading this documentation one may think that it would also preconfigure the stack with features appropriate for selected configuration.

openthread/cmake/extensions.cmake
```diff
@@ -26,7 +50,61 @@ endif()
...
+  string(REGEX MATCHALL ${PATTERN} match_list ${FILE_CONTENT})
+    foreach(element ${match_list})
+      string(REGEX MATCH ${PATTERN} MACRO_DEF ${element})
+      list(APPEND match1_list "${CMAKE_MATCH_1}\n")
+      set(${returnMatch1List} ${match1_list} PARENT_SCOPE)
+    endforeach()
+endfunction()
+
+function(check_openthread_dependencies ot_lib_nrf_security_mbedtls_config_file)
+  set(nrf_security_mbedtls_config_file "${CMAKE_CURRENT_BINARY_DIR}/../nrf_security/include/${CONFIG_MBEDTLS_CFG_FILE}")
```
edmont, line 82: This is checking the `CMAKE_CURRENT_BINARY_DIR`, but what if the build directory is empty? This is failing for me:
```

cd nrf/samples/openthread/cli
rm -rf build
west build -b nrf52840dk_nrf52840 -t install_openthread_libraries
rm -rf build
west build -b nrf52840dk_nrf52840 -- -DOVERLAY_CONFIG="harness/overlay-cert.conf"
```

edmont (changes requested):

openthread/cmake/extensions.cmake
```diff
@@ -3,20 +3,44 @@
...
+    set(ot_feature_set "ftd")
+  elseif(CONFIG_OPENTHREAD_NORDIC_LIBRARY_MTD)
+    set(ot_feature_set "mtd")
+  endif()
+
+  if(CONFIG_CC310_BACKEND)
+    set(nrf_security_backend "CC310")
+  elseif(CONFIG_OBERON_BACKEND)
+    set(nrf_security_backend "nrf_oberon")
+  endif()
```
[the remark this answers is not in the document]
lmaciejonczyk: We decided not to support it right now.

openthread/cmake/extensions.cmake
lmaciejonczyk: Any related documentation will be updated in reference sdk-nrf pull request.
Currently install_openthread_library target is active only if CONFIG_OPENTHREAD_NORDIC_LIBRARY which imply one of possible option: master, ftd, mtd. If we want a user to create library with custom configuration we could add another option there. What to you think?

pszkotak: It makes sense. With the current approach he would be forced to overwrite one of existing configurations.

openthread/cmake/extensions.cmake
lmaciejonczyk: It can't be empty. The referential file is created just before checking if only nrf_security is enabled.

edmont: What you mean with "_if only nrf_security is enabled_"? `CONFIG_OPENTHREAD_NRF_SECURITY` is enabled by default for OT samples, but with the previous steps I got this error:
```
/home/edmo/vscode/ncs/nrfxlib/openthread/lib/cortex-m4/soft-float/v1.1/master//nrf_security_mbedtls_configuration.h
/home/edmo/vscode/ncs/nrf/samples/openthread/cli/build/modules/nrfxlib/openthread/../nrf_security/include/
CMake Error at /home/edmo/vscode/ncs/nrfxlib/openthread/cmake/extensions.cmake:71 (file):
  file failed to open for reading (No such file or directory):

    /home/edmo/vscode/ncs/nrf/samples/openthread/cli/build/modules/nrfxlib/openthread/../nrf_security/include/
Call Stack (most recent call first):
  /home/edmo/vscode/ncs/nrfxlib/openthread/cmake/extensions.cmake:85 (get_active_mbedtls_configs_from_file)
  /home/edmo/vscode/ncs/nrfxlib/openthread/CMakeLists.txt:54 (check_openthread_dependencies)


-- Configuring incomplete, errors occurred!
```

tejlmand (changes requested):
This PR contains two large files, and consider how frequent those libs are updated, this starts to impact the time it takes to initialize NCS.

```
 BIN -51.4 MB openthread/lib/cortex-m4/soft-float/v1.1/libopenthread-ftd.a 
 BIN -36 MB openthread/lib/cortex-m4/soft-float/v1.1/libopenthread-mtd.a 
```
are there no ways to reduce the sizes of the lib.
(I assume the contain debugging symbols, but is that needed, if users could created identical lib locally based on conf file).

@carlescufi @mbolivar-nordic Any thoughts on this ?

openthread/cmake/extensions.cmake
```diff
@@ -3,20 +3,44 @@
 #
 # SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
 #
+include(${NRFXLIB_DIR}/common.cmake)
+include(${NRFXLIB_DIR}/nrf_security/cmake/extensions.cmake)
```
tejlmand, line 7: we should not include extensions from extensions.

please move all `include(<path>/extensions.cmake)` to toplevel CMakeLists.txt:
https://github.com/nrfconnect/sdk-nrfxlib/blob/master/CMakeLists.txt

so that we have a single location.
This will also align with how extensions are loaded in `sdk-nrf`

openthread/cmake/extensions.cmake
```diff
@@ -26,7 +50,61 @@ endif()
   endforeach()
 
   list(SORT OPENTHREAD_SETTINGS)
-  list(INSERT OPENTHREAD_SETTINGS 0 "OpenThread_commit=${openthread_git_hash}\n")
+
+  find_package(Git QUIET)
+  if(GIT_FOUND)
+    get_mbedtls_dir(ARM_MBEDTLS_PATH)
+    get_git_hash(${ARM_MBEDTLS_PATH})
+    list(INSERT OPENTHREAD_SETTINGS 0 "MBEDTLS_commit=${git_hash}\n")
```
tejlmand, line 60: I know a SHA is unique, but in `west.yml` we prefer to point to a tag.
The same when doing an NCS release, where all Nordic repos are tagged.

So in case the SHA is also a tag, I think it would be nice to include the tag as well.
Makes it easier to find exact version used (and the fact it was a release).
Keep the SHA also.

tejlmand (changes requested):
a quick look, with first observations.

nrf_security/cmake/extensions.cmake
```diff
@@ -0,0 +1,34 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+
+# Header guard
+if(__NRF_SECURITY_EXTENSIONS_CMAKE__)
+  return()
+endif()
```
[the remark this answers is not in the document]
lmaciejonczyk: Fixed also for common.cmake

openthread/cmake/extensions.cmake
lmaciejonczyk: Done for common.cmake. 
openthread/cmake/extension.cmake is only for internal use. I don't think it should be loaded if openthread is not enabled.

openthread/cmake/extensions.cmake
lmaciejonczyk: @edmont Could you check if it works for you now? Build from here: https://github.com/nrfconnect/sdk-nrf/pull/2974

edmont (approved):
It works great now, thanks!

edmont:
@lmaciejonczyk do you plan to add the different libraries in subsequent commits?

lmaciejonczyk:
> @lmaciejonczyk do you plan to add the different libraries in subsequent commits?

Yes I do, but I need strip them first. Look at https://github.com/nrfconnect/sdk-nrfxlib/pull/268#pullrequestreview-494508948

mbolivar-nordic:
> This PR contains two large files, and consider how frequent those libs are updated, this starts to impact the time it takes to initialize NCS.
> 
> ```
>  BIN -51.4 MB openthread/lib/cortex-m4/soft-float/v1.1/libopenthread-ftd.a 
>  BIN -36 MB openthread/lib/cortex-m4/soft-float/v1.1/libopenthread-mtd.a 
> ```
> 
> are there no ways to reduce the sizes of the lib.
> (I assume the contain debugging symbols, but is that needed, if users could created identical lib locally based on conf file).
> 
> @carlescufi @mbolivar-nordic Any thoughts on this ?

Yikes, those are pretty big if they are getting updated frequently. Is git LFS an option? Or isn't there a way to pull down artifacts from a URL using CMake?

tomchy (approved):

tejlmand:
> > This PR contains two large files, and consider how frequent those libs are updated, this starts to impact the time it takes to initialize NCS.
> > ```
> >  BIN -51.4 MB openthread/lib/cortex-m4/soft-float/v1.1/libopenthread-ftd.a 
> >  BIN -36 MB openthread/lib/cortex-m4/soft-float/v1.1/libopenthread-mtd.a 
> > ```
> > 
> > 
> > are there no ways to reduce the sizes of the lib.
> > (I assume the contain debugging symbols, but is that needed, if users could created identical lib locally based on conf file).
> > @carlescufi @mbolivar-nordic Any thoughts on this ?
> 
> Yikes, those are pretty big if they are getting updated frequently.

```
- commit b68cbe8ea8b6ccc7bf3fda72c6324f65a7dd587a
  Date:   Thu Sep 3 11:09:20 2020 +0200
- commit 108e18666370a96f72461177256eee80c54b6a32
  Date:   Mon Jul 6 13:07:09 2020 +0200
```
and then this PR.
I consider that frequent.

> Is git LFS an option?

then we should support that through `west`

> Or isn't there a way to pull down artifacts from a URL using CMake?

that would start to break the run `west update` and you are good to go.
No need to explain the `west update` and rush to airport, just to discover you haven't updates latest openthread libs.

common.cmake
```diff
@@ -46,3 +40,26 @@ function(nrfxlib_calculate_lib_path lib_path)
...
+  # Do nothing, just use the provided path
+  elseif(WEST)
+  ## Use `west` to find the mbedtls tree we use (this is not the
+  ## mbedtls tree distributed as a zephyr module).
+    execute_process(
+      COMMAND ${PYTHON_EXECUTABLE} -c
+      "from west.manifest import Manifest; manifest = Manifest.from_file(); print(manifest.get_projects(['mbedtls-nrf'])[0].posixpath)"
+      OUTPUT_VARIABLE west_project_output
+      RESULT_VARIABLE result
+    )
```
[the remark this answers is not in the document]
mbolivar-nordic: > Consider using the `west list` command. It might be more readable for you:

The existing code is done on purpose to avoid calling west with a different python interpreter than the one used by the rest of the build system. Later zephyr changes make it unnecessary, but they are not yet included in NCS, so I think this should remain as-is.

Your suggested change could be made once Zephyr commit 79b1f89f7d4229d494c8b72c0acd715d61108456 is included in NCS.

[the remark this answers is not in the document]
tejlmand: No, don't do that:
https://github.com/zephyrproject-rtos/zephyr/pull/26060

mbolivar-nordic:
> > Is git LFS an option?
> 
> then we should support that through `west`

I'm asking the sdk-nrfxlib maintainers if they consider it an option. We can discuss any missing west support if they say "yes", but it's not worth thinking about if they say "no".

lmaciejonczyk:
> > > This PR contains two large files, and consider how frequent those libs are updated, this starts to impact the time it takes to initialize NCS.
> > > ```
> > >  BIN -51.4 MB openthread/lib/cortex-m4/soft-float/v1.1/libopenthread-ftd.a 
> > >  BIN -36 MB openthread/lib/cortex-m4/soft-float/v1.1/libopenthread-mtd.a 
> > > ```
> > > 
> > > 
> > > are there no ways to reduce the sizes of the lib.
> > > (I assume the contain debugging symbols, but is that needed, if users could created identical lib locally based on conf file).
> > > @carlescufi @mbolivar-nordic Any thoughts on this ?
> > 
> > 
> > Yikes, those are pretty big if they are getting updated frequently.
> 
> ```
> - commit b68cbe8ea8b6ccc7bf3fda72c6324f65a7dd587a
>   Date:   Thu Sep 3 11:09:20 2020 +0200
> - commit 108e18666370a96f72461177256eee80c54b6a32
>   Date:   Mon Jul 6 13:07:09 2020 +0200
> ```
> 
> and then this PR.
> I consider that frequent.
> 
> > Is git LFS an option?
> 
> then we should support that through `west`
> 
> > Or isn't there a way to pull down artifacts from a URL using CMake?
> 
> that would start to break the run `west update` and you are good to go.
> No need to explain the `west update` and rush to airport, just to discover you haven't updates latest openthread libs.

I agree that debugging symbols are unnecessary considering users can create identical lib with them locally.
Libraries will be updated every NCS release or less often. It has not been decided.
Notice that this PR does not add any lib yet but remove old ones.

common.cmake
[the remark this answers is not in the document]
tomchy: Thank you for that update!

edmont:
When the debug symbols are stripped, those big libraries get reduced to ~1 MB each. I believe that would allow us to move forward this PR and add the different variants.

lmaciejonczyk:
@tejlmand Are the fixes ok for you?

joerchan (approved):
approving changes to MPSL and softdevice controller.

openthread/CMakeLists.txt
```diff
@@ -3,38 +3,39 @@
...
+    COMMAND ${CMAKE_STRIP} --strip-debug
+      $<TARGET_FILE:openthread-ftd>
+      $<TARGET_FILE:openthread-mtd>
+      $<TARGET_FILE:openthread-radio>
+      $<TARGET_FILE:openthread-cli-ftd>
+      $<TARGET_FILE:openthread-cli-mtd>
+      $<TARGET_FILE:openthread-ncp-ftd>
+      $<TARGET_FILE:openthread-ncp-mtd>
+      $<TARGET_FILE:openthread-rcp>
+  )
```
tejlmand, line 35: This is very GNU centric.
Zephyr is working towards supporting more toolchains, so I think we should use the bintools abstraction here:
https://github.com/zephyrproject-rtos/zephyr/blob/master/cmake/bintools/bintools_template.cmake#L150-L158
https://github.com/zephyrproject-rtos/zephyr/blob/master/cmake/bintools/gnu/target_bintools.cmake#L85

For inspiration, see:
https://github.com/zephyrproject-rtos/zephyr/blob/master/CMakeLists.txt#L1221-L1226

Then this will also work when Zephyr starts to have support for more toolchains.

Note: consider to have one command per target_file, as not all strip tools may interpret a missing `outfile` as overriding the input file.

openthread/CMakeLists.txt
```diff
@@ -3,38 +3,39 @@
...
 
   add_custom_target(install_openthread_libraries
```
tejlmand, line 37: should stripping be default when calling `install_openthread_libraries` ?

tejlmand (changes requested):

openthread/CMakeLists.txt
lmaciejonczyk: I think not, user shoukd have option to create libraries with debug symbols.

openthread/CMakeLists.txt
lmaciejonczyk: Unfortunately bintools abstraction is provided by zephyr v2.4.0 and currently is not available in downstream.

lmaciejonczyk: Done in accordance with your recommendation, but waits for update zephyr in current https://github.com/nrfconnect/sdk-nrf/pull/2918.

tejlmand: > Unfortunately bintools abstraction is provided by zephyr v2.4.0 and currently is not available in downstream.

right, had forgotten that :)
Thanks for updating to Zephyr 2.4 :+1:

openthread/CMakeLists.txt
tejlmand: then I think it would be nice to follow same pattern as Zephyr:
https://github.com/zephyrproject-rtos/zephyr/blob/ed7263bf6b91d4542573d2f44ed742a2acf542df/Kconfig.zephyr#L327-L331
```
config BUILD_OUTPUT_STRIPPED
	bool "Build a stripped binary"
	help
	  Build a stripped binary zephyr/zephyr.strip in the build directory.
	  The name of this file can be customized with CONFIG_KERNEL_BIN_NAME.
```
maybe like a:
```
config OPENTHREAD_BUILD_OUTPUT_STRIPPED
    .....
```

Remember, you can always enable this like:
```
cmake -DCONFIG_OPENTHREAD_BUILD_OUTPUT_STRIPPED=y <build_folder>; ninja -C<build_folder> install_openthread_libraries
```

lmaciejonczyk: Done. Now stripping is optional when calling `install_openthread_libraries`. Added new Kconfig option:
```
config OPENTHREAD_STRIP_LIBRARIES_BEFORE_INSTALL
	bool "Strip OpenThread libraries before installation"
	default y
	help
	  Valid for custom target: install_openthread_libraries.
```

openthread/CMakeLists.txt
```diff
@@ -3,66 +3,114 @@
...
+              $<TARGET_PROPERTY:bintools,strip_flag_debug>
+              $<TARGET_PROPERTY:bintools,strip_flag_infile>$<TARGET_FILE:openthread-ncp-mtd>
+              $<TARGET_PROPERTY:bintools,strip_flag_outfile>"${OPENTHREAD_DST_DIR}/$<TARGET_FILE_NAME:openthread-ncp-mtd>"
+              $<TARGET_PROPERTY:bintools,strip_flag_final>
+      COMMAND $<TARGET_PROPERTY:bintools,strip_command>
+              $<TARGET_PROPERTY:bintools,strip_flag>
+              $<TARGET_PROPERTY:bintools,strip_flag_debug>
+              $<TARGET_PROPERTY:bintools,strip_flag_infile>$<TARGET_FILE:openthread-rcp>
+              $<TARGET_PROPERTY:bintools,strip_flag_outfile>"${OPENTHREAD_DST_DIR}/$<TARGET_FILE_NAME:openthread-rcp>"
+              $<TARGET_PROPERTY:bintools,strip_flag_final>
```
tejlmand, line 75: Please do this in a loop.

```suggestion
foreach(target openthread-ftd openthread-mtd openthread-radio openthread-cli-ftd openthread-cli-mtd openthread-ncp-ftd openthread-ncp-mtd openthread-rcp)
    list(APPEND INSTALL_COMMANDS
      COMMAND $<TARGET_PROPERTY:bintools,strip_command>
              $<TARGET_PROPERTY:bintools,strip_flag>
              $<TARGET_PROPERTY:bintools,strip_flag_debug>
              $<TARGET_PROPERTY:bintools,strip_flag_infile>$<TARGET_FILE:${target}>
             $<TARGET_PROPERTY:bintools,strip_flag_outfile>"${OPENTHREAD_DST_DIR}/$<TARGET_FILE_NAME:${target}>"
              $<TARGET_PROPERTY:bintools,strip_flag_final>
    )
endforeach()
```

openthread/Kconfig
```diff
@@ -100,8 +100,20 @@ config OPENTHREAD_NORDIC_LIBRARY_MTD
...
 	  Nordic Semiconductor optimized OpenThread features for MTD.
+
+config OPENTHREAD_USER_CUSTOM_LIBRARY
+	bool "User custom configuration"
+	help
+	  No extra features selected.
+
 endchoice
 
+config OPENTHREAD_STRIP_LIBRARIES_BEFORE_INSTALL
```
tejlmand, line 111: nothing wrong, but you could have chosen to make it possible to strip the libraries independently of install.

Meaning, user only has to decide if stripped version should also be made available, like is done for Zephyr.
And then install will choose to install what is available.
But I leave the decision to you.

tejlmand (changes requested):

openthread/Kconfig
lmaciejonczyk: I agree that it would be the best solution, nevertheless it would require changes in zephyr for what is quite late concerning freeze code before release. I'll stick with current solution.

tejlmand: > nevertheless it would require changes in zephyr for what is quite late concerning freeze code before release. 

No, made a proposal here:
https://github.com/lmaciejonczyk/sdk-nrfxlib/pull/2

without stripping:
```
$ cmake -GNinja -DBOARD=nrf52840dk_nrf52840 -DEXTRA_CFLAGS="-Werror" -DCONFIG_OPENTHREAD_BUILD_OUTPUT_STRIPPED=n -Bbuild samples/openthread/cli/
...
$ ninja -Cbuild
...
$ find build . -name 'libopenthread*'
build/modules/openthread/build/src/cli/libopenthread-cli-ftd.a
build/modules/openthread/build/src/core/libopenthread-ftd.a
build/zephyr/subsys/net/lib/openthread/platform/libopenthread_platform.a
```
and with stripping:
```
$ cmake -GNinja -DBOARD=nrf52840dk_nrf52840 -DEXTRA_CFLAGS="-Werror" -Bbuild -DCONFIG_OPENTHREAD_BUILD_OUTPUT_STRIPPED=y samples/openthread/cli/
...
$ ninja -Cbuild
...
$ find build . -name 'libopenthread*'
build/modules/openthread/build/src/cli/libopenthread-cli-mtd.a
build/modules/openthread/build/src/cli/libopenthread-cli-mtd.a.strip
build/modules/openthread/build/src/cli/libopenthread-cli-ftd.a
build/modules/openthread/build/src/cli/libopenthread-cli-ftd.a.strip
...
```

tejlmand: Note: I disabled stripping per default in my proposal, as that will significantly reduce build time.

lmaciejonczyk: Very nice, thank you.

tejlmand (approved):
Thanks for the patience.
Looks good now.

The sizes of each lib is much more acceptable, but still a little concerned on the overall size usage.
A lot of libs adds up to total size, but I assume all the variants are needed:
```
$ du . -h 
3.0M    ./lib/cortex-m4/soft-float/v1.1/master/cc310
3.0M    ./lib/cortex-m4/soft-float/v1.1/master/oberon
5.9M    ./lib/cortex-m4/soft-float/v1.1/master
2.6M    ./lib/cortex-m4/soft-float/v1.1/mtd/cc310
2.6M    ./lib/cortex-m4/soft-float/v1.1/mtd/oberon
5.2M    ./lib/cortex-m4/soft-float/v1.1/mtd
2.7M    ./lib/cortex-m4/soft-float/v1.1/ftd/cc310
2.7M    ./lib/cortex-m4/soft-float/v1.1/ftd/oberon
5.3M    ./lib/cortex-m4/soft-float/v1.1/ftd
17M     ./lib/cortex-m4/soft-float/v1.1
17M     ./lib/cortex-m4/soft-float
17M     ./lib/cortex-m4
17M     ./lib
```

17M in total adds up pretty fast after a couple of PRs.
48 files in total:
```
$ find . -name 'lib*.a'|wc -l
48
```
@mbolivar-nordic Thoughts ?

mbolivar-nordic:
> @mbolivar-nordic Thoughts ?

I think we should pick a short term resolution then see if we need new west features (like shallow clones or sparse checkouts or something) for the future.

edmont:
Several tasks are pending on the presence of these libraries. It would be nice to have them already and maybe introduce optimizations in the future.

lmaciejonczyk:
It's ready for merge. @rlubos can we proceed?

rlubos:
@lmaciejonczyk I'd like to see a green CI on https://github.com/nrfconnect/sdk-nrf/pull/2974 first, there's not a single green one in recent builds.

lmaciejonczyk:
> @lmaciejonczyk I'd like to see a green CI on [nrfconnect/sdk-nrf#2974](https://github.com/nrfconnect/sdk-nrf/pull/2974) first, there's not a single green one in recent builds.

How this can be green when there is a conflict in manifest about this PR? How can I fixes it?

rlubos:
@lmaciejonczyk Rebase in sdk-nrf PR, to allow the CI to run. I'm aware it's a bit frustrating, but there's a lot of PR that are being merged right now so conflicts can happen.

MarekPorwisz (approved):

MarekPorwisz (approved):
Tested and seems to work

lmaciejonczyk:
@tejlmand @rlubos CI is green, merge please

frkv (approved):
LGTM

PR #288, opened 2020-10-07
nrf_security: Optimize memory footprint

lmaciejonczyk:
Add choice option to enable or disable majority of NRF Security
features by default. Selecting only required cryptographic features can
be beneficial for optimizing memory footprint for more advance users.

### DIFF
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -162,7 +162,7 @@ menuconfig NRF_SECURITY_RNG
 	bool
 	prompt "Random Number Generator support"
 	depends on ENTROPY_HAS_DRIVER
-	default y
+	default y if !NRF_SECURITY_DEFAULT_ENABLE_NONE
 	select MBEDTLS_SHA256_C
 	help
 	  The Random Number Generator support in nRF Security provides a
@@ -195,7 +195,7 @@ endif # NRF_SECURITY_RNG
 menuconfig MBEDTLS_AES_C
 	bool
 	prompt "AES   - Advanced Encryption Standard"
-	default y
+	default y if !NRF_SECURITY_DEFAULT_ENABLE_NONE
 	select CC310_MBEDTLS_AES_C if CC310_SINGLE_BACKEND
 	select OBERON_MBEDTLS_AES_C if OBERON_SINGLE_BACKEND
 	select VANILLA_MBEDTLS_AES_C if MBEDTLS_VANILLA_SINGLE_BACKEND
@@ -335,7 +335,7 @@ comment "Cipher Selection"
 
 config MBEDTLS_CIPHER_MODE_CBC
 	bool "AES-CBC - AES Cipher Block Chaining mode"
-	default y
+	default y if !NRF_SECURITY_DEFAULT_ENABLE_NONE
 	help
 	  Enable the AES Cipher Block Chaining (CBC) mode,
 	  MBEDTLS_CIPHER_MODE_CBC setting in mbed TLS config file.
@@ -403,7 +403,7 @@ endif # MBEDTLS_CIPHER_MODE_CBC
 
 config MBEDTLS_CIPHER_MODE_CTR
 	bool "AES-CTR - AES Counter Block Cipher mode"
-	default y
+	default y if !NRF_SECURITY_DEFAULT_ENABLE_NONE
 	help
 	  Enable the AES Counter Block Cipher mode (CTR) mode,
 	  MBEDTLS_CIPHER_MODE_CTR setting in mbed TLS config file.
@@ -438,7 +438,7 @@ endif # MBEDTLS_CIPHER_MODE_CTR
 config MBEDTLS_CIPHER_MODE_CFB
 	bool "AES-CFB - AES Cipher Feedback mode"
 	depends on OBERON_MBEDTLS_AES_C || VANILLA_MBEDTLS_AES_C
-	default y
+	default y if !NRF_SECURITY_DEFAULT_ENABLE_NONE
 	help
 	  Enable the AES Cipher Feedback mode (CFB) mode,
 	  MBEDTLS_CIPHER_MODE_CFB setting in mbed TLS config file.
@@ -468,7 +468,7 @@ endif # MBEDTLS_CIPHER_MODE_CFB
 config MBEDTLS_CIPHER_MODE_OFB
 	bool "AES-OFB - AES Output Feedback mode"
 	depends on OBERON_MBEDTLS_AES_C || VANILLA_MBEDTLS_AES_C
-	default y
+	default y if !NRF_SECURITY_DEFAULT_ENABLE_NONE
 	help
 	  Enable the AES Output Feedback mode (OFB) mode,
 	  MBEDTLS_CIPHER_MODE_OFB setting in mbed TLS config file.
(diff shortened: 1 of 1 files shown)


edmont (approved):
LGTM. Maybe it should be documented somewhere that samples must explicitly enable the desired options?

Please fix the commit title typo.

nrf_security/Kconfig
```diff
@@ -173,18 +173,14 @@ menuconfig NRF_SECURITY_RNG
 if NRF_SECURITY_RNG
 
 config MBEDTLS_CTR_DRBG_C
-	bool
-	prompt "PRNG using CTR_DRBG"
-	default y
+	bool "PRNG using CTR_DRBG"
```
tejlmand, line 176: Nothing wrong on this line, but comments cannot be place on the commit message:
https://github.com/nrfconnect/sdk-nrfxlib/pull/288/commits/ec31b1998947cabd9691b27e87139711c035ebc6

Please spellcheck the commit message title:
```
nrf_security: Optimize momory footprint
```

tejlmand (changes requested):
This change has quite a big impact this close to code freeze.
All samples (and customer applications) that expects the current set of defaults will now fail.

This PR does not seem to address this problem in any of our samples.
If this should go in, it should as minimum ensure that apps/samples in NCS the uses nRF Security are updated.

Also such a change should have a clear entry in the release notes.

lmaciejonczyk:
> LGTM. Maybe it should be documented somewhere that samples must explicitly enable the desired options?
> 
> Please fix the commit title typo.

Done. Release notes should be enough.

nrf_security/Kconfig
lmaciejonczyk: It looks like only OpenThread samples uses nrf_security and as you's seen the changes have been addressed. There is also SPM_SECURE_SERVICES but it uses NRF_SECURITY_RNG which hasn't been changed.
For customers the release notes should be enough and they will get lower memory footprint in return.

lmaciejonczyk: I can obviously disable unnecessary nrf_security options only for OpenThread samples, but this would be a walk around fix which I believe we don't aim into.

frkv:
This PR is not accepted, as it breaks our CI.

lmaciejonczyk:
@frkv @tejlmand Is this PR acceptable after my last changes?

nrf_security/Kconfig
```diff
@@ -1459,6 +1459,15 @@ config MBEDTLS_SSL_CIPHERSUITES
 	  Warning: This field has offers no validation checks.
 	  MBEDTLS_SSL_CIPHERSUITES setting in mbed TLS config file.
 
+config NRF_SECURITY_DEFAULT_ENABLE_NONE
+	bool
+	help
+	  Disable all NRF Security features by default.
+	  It can save memory footprint by disabled not necessary features
```
frkv, line 1466: features, <-- comma

nrf_security/Kconfig
```diff
@@ -1459,6 +1459,15 @@ config MBEDTLS_SSL_CIPHERSUITES
 	  Warning: This field has offers no validation checks.
 	  MBEDTLS_SSL_CIPHERSUITES setting in mbed TLS config file.
 
+config NRF_SECURITY_DEFAULT_ENABLE_NONE
+	bool
+	help
+	  Disable all NRF Security features by default.
```
frkv, line 1465: NRF => nRF

nrf_security/Kconfig
```diff
@@ -1459,6 +1459,15 @@ config MBEDTLS_SSL_CIPHERSUITES
...
+	  It can save memory footprint by disabled not necessary features
+	  but requires explicitly selecting needed.
+	  This is promptless by purpose and state that another kconfig is the only way
+	  to set this configuraiton. The effect of the choice only works once.
```
frkv, line 1469: configuraiton => configuration

frkv (approved):
LGTM

nrf_security/Kconfig
```diff
@@ -1459,6 +1459,14 @@ config MBEDTLS_SSL_CIPHERSUITES
...
 	  MBEDTLS_SSL_CIPHERSUITES setting in mbed TLS config file.
 
+config NRF_SECURITY_DEFAULT_ENABLE_NONE
+	bool
+	help
+	  Disable all nRF Security features by default.
+	  It can save memory by disabling features that are not needed.
+	  This is promptless by purpose and state that another kconfig is the only way
+	  to set this configuration. The effect of the choice only works once.
+
```
greg-fer, line 1469:
```suggestion
	  Disable all nRF Security features.
	  It can save memory by disabling features that are not needed.
	  When enabled, if the project requires specific nRF Security features,
	  these features must be selected explicitly one by one.
	  This option is promptless and can be only set manually from within
	  a configuration file. The effect of disabling nRF Security features
	  only works once.
```
Added some lines discussed with Eduardo. Still, I don't understand the last sentence, and it needs to be clarified.

greg-fer (changes requested):

greg-fer:
Apologies for adding the 1.4.0 milestone -- I understood all concerns have been addressed and this milestone was ommitted by mistake. Now I see from the comments that there are some concerns about documentation of samples other than Thread ones?

I'm adding doc-review label, as there is my comment pending for the option description, but please let me know first about impact on non-Thread sample docs.

frkv:
> Apologies for adding the 1.4.0 milestone -- I understood all concerns have been addressed and this milestone was ommitted by mistake. Removed it for now.
> 
> I see from the comments that there are some concerns from @tejlmand about documentation of samples other than Thread ones. I'm adding doc-review label, as there is my comment pending for the option description, but @lmaciejonczyk please let me know first about impact on non-Thread sample docs and include this change in the release notes.

We have been in discussions with Torsten on the feature and are in agreement. This feature has no impact for regular use as the configuration is only settible by other Kconfig parameters. This is not something the end-user will see unless browsing for hidden symbols. This is mainly going to be used in CI and for specific projects, meaning there are ample opportunity to document this outside of this feature

nrf_security/Kconfig
lmaciejonczyk: Description has been updated, I hope it's ok now.

lmaciejonczyk:
CI is green, waiting only for reviewers.

frkv (approved):
Still LGTM

greg-fer (approved):

rlubos:
@tejlmand Could you please re-check this PR? Your change request is blocking the merge.

nrf_security/Kconfig
```diff
@@ -1476,6 +1476,16 @@ config MBEDTLS_SSL_CIPHERSUITES
 	  Warning: This field has offers no validation checks.
 	  MBEDTLS_SSL_CIPHERSUITES setting in mbed TLS config file.
 
+config NRF_SECURITY_DEFAULT_ENABLE_NONE
+	bool
```
tejlmand, line 1480: I have given this some more thoughts.
This deviates to much from how defaults are used elsewhere.
For NCS users, this becomes hard to understand when a default will be `y` and when it will be `n`.

If a user examines this file and reads:
```
default y if !NRF_SECURITY_DEFAULT_ENABLE_NONE
```
he has no way of predicting whether this code will default to `y` or `n`.

Compared to:
https://github.com/nrfconnect/sdk-zephyr/blob/dc1d8b4da3d58bbe5f4b8c290d2e1da3ce921cbd/subsys/bluetooth/Kconfig#L100
```
	default y if !BT_OBSERVER
```
where user can easily spot that this will be enabled when not making a Bluetooth observer.
or
https://github.com/nrfconnect/sdk-zephyr/blob/dc1d8b4da3d58bbe5f4b8c290d2e1da3ce921cbd/subsys/bluetooth/controller/Kconfig.ll_sw_split#L330
```
	default y if BT_HCI_RAW
```


It is not clear who might set `NRF_SECURITY_DEFAULT_ENABLE_NONE=y` and it is not a setting user can select for himself in a conf file.

menuconfig is not of much help:
image

If something should be done, then we should follow the principle of upstream.
Default everything to `n` and allow users a single `Enable all` for each group.
https://github.com/nrfconnect/sdk-zephyr/blob/master/modules/Kconfig.tls-generic#L46
```
config MBEDTLS_KEY_EXCHANGE_ALL_ENABLED
```
https://github.com/nrfconnect/sdk-zephyr/blob/master/modules/Kconfig.tls-generic#L106
```
config MBEDTLS_ECP_ALL_ENABLED
```

tejlmand (changes requested):

PR #324, opened 2020-11-10
nrf_security: Add module for storing keys in flash memory

lmaciejonczyk:
The sst_settings module make use of Zephyr's "settings" module for providing access to persistent storage in a key/value manner.

### DIFF
--- a/nrf_security/sst_settings/Kconfig
+++ b/nrf_security/sst_settings/Kconfig
@@ -1,0 +1,17 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+config SST_SETTINGS
+	bool "Settings module for Secure Storage"
+	help
+	  Enable to get access for flashin operation based on Settings subsys.
+
+if SST_SETTINGS
+
+module = SST_SETTINGS
+module-str = Secure Storage Settings
+source "${ZEPHYR_BASE}/subsys/logging/Kconfig.template.log_config"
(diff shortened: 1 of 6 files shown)


nrf_security/sst_settings/Kconfig
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+config SST_SETTINGS
+	bool "Settings module for Secure Storage"
+	help
+	  Enable to get access for flashin operation based on Settings subsys.
```
MarekPorwisz, line 9: Not sure about this sentence. 
Maybe "Enable Secure Storage flash read/write operations using the Settings subsystem."

nrf_security/sst_settings/Kconfig
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+#
+config SST_SETTINGS
+	bool "Settings module for Secure Storage"
```
MarekPorwisz, line 7: I think we are missing Settings dependency here. 
Do not allow enabling this option if settings subsystem is not enabled.

MarekPorwisz (commented):
I've added two comments.

lmaciejonczyk:
It's replaced by https://github.com/nrfconnect/sdk-nrfxlib/pull/395

PR #327, opened 2020-11-13
bsdlib: doc: gps conceptual doc update

evenl:
Adding documentation about the low accuracy use case mode in the GPS
module.

Manifest update: https://github.com/nrfconnect/sdk-nrf/pull/3338

### DIFF
--- a/bsdlib/doc/gnss_extension.rst
+++ b/bsdlib/doc/gnss_extension.rst
@@ -3,10 +3,6 @@
 GNSS socket
 ###########
 
-.. contents::
-   :local:
-   :depth: 2
-
 `Global navigation satellite system (GNSS)`_ socket is one of the socket types supported by the BSD library.
 This socket type is used to configure and fetch GPS position fix data from the GPS module, and to write `A-GPS`_ data to the GPS module.
 
@@ -30,7 +26,7 @@ Starting the GPS
 ****************
 
 After the GNSS socket is created, the GPS module must be started for the module to start generating GPS fixes.
-This is done using the :c:func:`nrf_setsockopt` function with the :c:type:`NRF_SOL_GNSS` socket option level and the :c:type:`NRF_SO_GNSS_START` socket option.
+This is done using the :c:func:`nrf_setsockopt` function with the ``NRF_SOL_GNSS`` socket option level and the :c:type:`NRF_SO_GNSS_START` socket option.
 
 .. code-block:: c
 
@@ -43,7 +39,7 @@ This is done using the :c:func:`nrf_setsockopt` function with the :c:type:`NRF_S
 During a session (when the GPS is running), the socket stores the information to the non-volatile memory.
 This is done to generate a quick fix in a subsequent session and is termed as a hot or warm start (depending on the data that is used from the last session).
 Such session data can be deleted using the delete mask that is supplied to the :c:func:`nrf_setsockopt` function call while starting the GPS module.
-A value of 0 in the delete mask is an indication to keep all the previous data.
+A value of zero in the delete mask is an indication to keep all the previous data.
 The bit masks for the different types of data that can be deleted is defined in the GNSS socket API documentation of :c:type:`nrf_gnss_delete_mask_t`.
 
 Stopping the GPS
@@ -86,7 +82,11 @@ A configuration can also be read by the application by using the :c:func:`nrf_ge
 Configuration parameters
 ========================
 
+The various configuration parameters that can be used with the GNSS socket are described below.
+
 Fix interval
+------------
+
    | Socket option:  :c:type:`NRF_SO_GNSS_FIX_INTERVAL`
    | Datatype:       :c:type:`nrf_gnss_fix_interval_t`
    | Default:        1
@@ -96,21 +96,22 @@ The fix interval parameter controls the time between GNSS receiver-starts.
 It also controls the stopping of the GNSS receiver after a valid PVT
 estimate.
 The fix interval parameter determines the mode of navigation.
-There are three different navigation modes available: single-fix,
-continuous, and periodic.
+There are three different navigation modes available: single-fix, continuous, and periodic.
 
-Single-fix navigation mode is engaged by setting the fix interval to 0.
+Single-fix navigation mode is engaged by setting the fix interval to zero.
 In this mode, once a valid PVT estimate is produced, the GNSS receiver is turned off indefinitely.
 It does not resume navigation without explicit actions by the application processor.
 To do another single-fix, the application processor must first stop the GNSS module and then start the GNSS module again.
 
-Continuous navigation mode is engaged by setting the fix interval to 1.
+Continuous navigation mode is engaged by setting the fix interval to one.
 In this mode, the GNSS receiver continues to produce fixes at 1 Hz rate without any time limit.
 
-Periodic navigation mode is engaged by setting the fix interval to a value other than 0 or 1.
+Periodic navigation mode is engaged by setting the fix interval to a value other than zero or one.
 In this mode the GNSS receiver is turned off after each valid PVT estimate, and turned back on periodically after each fix interval has passed.
 
 Fix retry
+---------
+
    | Socket option:  :c:type:`NRF_SO_GNSS_FIX_RETRY`
    | Datatype:       :c:type:`nrf_gnss_fix_retry_t`
    | Default:        60
@@ -118,9 +119,11 @@ Fix retry
 
 The fix retry parameter controls the maximum time the GNSS receiver can run while trying to produce a valid PVT estimate.
 If the fix retry time is non-zero, the GNSS receiver is turned off after the fix retry time is up regardless of whether a valid PVT estimate was produced or not.
-If the fix retry parameter is set to 0, the GNSS receiver can run indefinitely until a valid PVT estimate is produced.
+If the fix retry parameter is set to zero, the GNSS receiver can run indefinitely until a valid PVT estimate is produced.
 
 NMEA mask
+---------
+
    | Socket option: :c:type:`NRF_SO_GNSS_NMEA_MASK`
    | Datatype:      :c:type:`nrf_gnss_nmea_mask_t`
    | Default:       0x0000
@@ -136,6 +139,8 @@ This mask is used to enable different `NMEA`_ string types.
 Multiple NMEA string types can be enabled at the same time.
 
 Elevation mask
+--------------
+
    | Socket option:  :c:type:`NRF_SO_GNSS_ELEVATION_MASK`
    | Datatype:       :c:type:`nrf_gnss_elevation_mask_t`
    | Default         5
@@ -145,15 +150,62 @@ The elevation mask parameter controls the elevation threshold angle (in degrees)
 Satellites with elevation angle less than the threshold are automatically excluded from the PVT estimation.
 
 Use case
+--------
+
    | Socket option:  :c:type:`NRF_SO_GNSS_USE_CASE`
    | Datatype:       :c:type:`uint8_t`
-   | Default         1
-   | Allowed values: 0, 1
+   | Default:        :c:type:`NRF_GNSS_USE_CASE_MULTIPLE_HOT_START` | :c:type:`NRF_GNSS_USE_CASE_NORMAL_ACCURACY`
+   | Allowed values:
 
-The use case parameter determines the target performance.
-A value of 0 indicates single cold start and one indicates multiple hot starts.
+   | :c:type:`NRF_GNSS_USE_CASE_SINGLE_COLD_START`
+   | :c:type:`NRF_GNSS_USE_CASE_MULTIPLE_HOT_START`
+   | :c:type:`NRF_GNSS_USE_CASE_NORMAL_ACCURACY`
+   | :c:type:`NRF_GNSS_USE_CASE_LOW_ACCURACY`
+
+This bitmask is used to specify the combination of the target GNSS performance (single cold or multiple hot start) and the accuracy (normal or low).
+
+Low accuracy mode
++++++++++++++++++
+
+If you enable the low accuracy mode, the GNSS receiver demonstrates a looser acceptance criterion for a fix.
+The error in position calculation, when compared to the actual position, can be larger than the error in normal accuracy mode.
+In addition, in the low accuracy mode, the GNSS receiver might use only three satellites to determine a fix.
+In normal accuracy mode, four or more satellites are used.
+
+For a possible position fix using three satellites, one of the following criteria must be satisfied initially and when the criterion is met, it will be valid for a time duration of 24 hours:
+
+* At least a single fix must have taken place using five or more satellites. If you initiate the GPS to start in low accuracy mode, the GNSS engine will not be able to utilize the fix using three satellites before the aforementioned condition has taken place. In the subsequent time window following the fix using five satellites, any fix that uses five or more satellites results in the resetting of the time window and extension of the 24-hour time window.
+
+* The altitude (in meters), altitude uncertainty (0 percent), and confidence (100 percent) must be given as input to the GNSS engine using the :c:type:`nrf_gnss_agps_data_location_t` A-GPS data location struct as shown in the code below. To be able to remain in low accuracy mode, this step must be repeated at least once for every 24-hour duration, results in resetting the timer and extension of the 24-hour time window.
+
+
+.. important::
+   The altitude must be accurate to a value within ±10 meters of the actual altitude of the device. An erroneous altitude will result in a severe error in the position fix calculation using three satellites.
+
+.. code-block:: c
+
+   nrf_gnss_agps_data_location_t location;
+
+   location.latitude          = 0;        /* Unknown, set to 0 */
+   location.longitude         = 0;        /* Unknown, set to 0 */
+   location.altitude          = altitude; /* Actual altitude of the device in meters */
+   location.unc_semimajor     = 127;      /* Set to 127 when location is unknown */
+   location.unc_semiminor     = 127;      /* Set to 127 when location is unknown */
+   location.orientation_major = 0;        /* Unknown, set to 0 */
+   location.unc_altitude      = 0;        /* Set to 0 */
+   location.confidence        = 100;      /* Set to 100 */
+
+If the actual altitude of the device changes with respect to the altitude stored in the GNSS engine (for example, when the device moves around), the accuracy of the position fix using three satellites will be degraded.
+
+The low accuracy mode is different from the 2D fix that is documented in the NMEA reports.
+All fixes, including the low accuracy fixes, will be reported as 3D fixes.
+See the `NMEA report sample`_ and number of IDs of SVs used in the position fix to get information of the number of satellites are used for the position fix.
+
+
 
 Start GPS module
+----------------
+
    | Socket option: :c:type:`NRF_SO_GNSS_START`
    | Datatype:      :c:type:`nrf_gnss_delete_mask_t`
    | Default:       NA
@@ -162,14 +214,18 @@ This parameter makes the GPS module start generating fixes.
 The delete mask is used to delete the data that the GPS module has stored for any previous session.
 
 Stop GPS module
+---------------
+
    | Socket option: :c:type:`NRF_SO_GNSS_STOP`
    | Datatype:      :c:type:`nrf_gnss_delete_mask_t`
-   | Default        NA
+   | Default:       NA
 
 This parameter makes the GPS module stop generating fixes.
 The delete mask is used to delete data that the GPS module has stored from any previous session.
 
 Power save modes
+----------------
+
    | Socket option: :c:type:`NRF_SO_GNSS_POWER_SAVE_MODE`
    | Datatype:      :c:type:`nrf_gnss_power_save_mode_t`
    | Default:       :c:type:`NRF_GNSS_PSM_DISABLED`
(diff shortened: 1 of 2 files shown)


bsdlib/doc/gnss_extension.rst
```diff
@@ -143,11 +143,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+    {
+        int32_t latitude          = 0;     // Not known, set to 0
+        int32_t longitude         = 0;     // Not known, set to 0
+        int16_t altitude          = meter; // Your actual height in meter
+        uint8_t unc_semimajor     = 127;   // Both set to 127 when latitude
+        uint8_t unc_semiminor     = 127;   // and longitude are unknown
+        uint8_t orientation_major = 0;     // Not known, set to 0
+        uint8_t unc_altitude      = 0;     // Set to 0
+        uint8_t confidence        = 100;   // Set to 100
+    } nrf_gnss_agps_data_location_t;
```
jtguggedal, line 188: Is it intentional that the typedef and struct definition along with values? Or was the intention to show how to declare a struct to set the altitude?

bsdlib/doc/gnss_extension.rst
```diff
@@ -143,11 +143,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+   | :c:type:`NRF_GNSS_USE_CASE_SINGLE_COLD_START`
+   | :c:type:`NRF_GNSS_USE_CASE_MULTIPLE_HOT_START`
+   | :c:type:`NRF_GNSS_USE_CASE_NORMAL_ACCURACY`
+   | :c:type:`NRF_GNSS_USE_CASE_LOW_ACCURACY`
+
+This bitmask is used to specify the combination of the target GNSS performance (single cold or multiple hot start)
+and accuracy (normal or low).
+
+..Note::
+  Normal and Low accuracy is available in MFW 1.2.2 and later
```
jtguggedal, line 158: Perhaps write "nRF9160 modem firmware" or something like that instead of "MFW".

lemrey (commented):
the `sdk-nrf` PR needs a rebase

rlubos (approved):

bsdlib/doc/gnss_extension.rst
```diff
@@ -143,11 +143,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+
+Enabling low accuracy mode will result in the GNSS receiver having a looser acceptance criterion
+for a fix (i.e. error in calculation vs actual position can be larger than Normal Accuracy mode).
+In addition, in the Low Accuracy mode the GNSS receiver may use only 3 satellites to determine a
+fix (in Normal Accuracy mode, it needs 4 or more satellitesi).
+
+For a possible 3 satellite fix to take place, one of the following criteria must be in place first
+and will be valid for a time window of 24 hour:
+
+ | 1) At least 1 fix must have been taken place using 5 or more satellites. You initiate the GPS to start Low accuracy mode, but the radio will not be able to utilize 3 satellite fixing before the criteria above has taken place. In the following time window, any fix that is done using 5 or more satellites will reset the time window and extend the 24-hour period.
```
jtguggedal, line 170: Is there a `been` too much in the first sentence?
The other sentences could need some rephrasing to clarify. I don't think it's the radio that uses the satellite information, but rather the GNSS engine. By this `In the following time window` I assume it's the 24-hour time window following a fix using 5 satellites?

bsdlib/doc/gnss_extension.rst
```diff
@@ -143,11 +143,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+ |
+ | 2) Injecting Altitude (in meter), Altitude uncertainty (0) , and Confidence (100 percent) to the GPS radio using agps data location struct as described below. To be able to stay in Low Accuracy mode this step has to be repeated at least every 24 hour.
```
jtguggedal, line 172: GPS radio -> GNSS engine (if that's the correct term).
` to stay in Low Accuracy mode this step has to be repeated at least every 24 hour`, unless there's a 5-or-more-satellite fix within that 24-hour time window?

bsdlib/doc/gnss_extension.rst
```diff
@@ -143,11 +143,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+ |
+ | **Caution:** The Altitude in meter has to have an accuracy in the range of +/- 10’s of meter of your actual height. An erroneous altitude will result in a quite sever error in 3 satellite position fix calculation.
```
jtguggedal, line 174: your actual height -> the device's actual altitude
sever -> severe

bsdlib/doc/gnss_extension.rst
```diff
@@ -143,11 +143,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+
+.. code-block:: c
 
-The use case parameter determines the target performance.
-A value of 0 indicates single cold start and one indicates multiple hot starts.
+    typedef struct
+    {
+        int32_t latitude          = 0;     // Not known, set to 0
+        int32_t longitude         = 0;     // Not known, set to 0
+        int16_t altitude          = meter; // Your actual height in meter
```
jtguggedal, line 182: your actual height -> the device's actual altitude

bsdlib/doc/gnss_extension.rst
```diff
@@ -143,11 +143,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+ |
+ | **Caution:** The Altitude in meter has to have an accuracy in the range of +/- 10’s of meter of your actual height. An erroneous altitude will result in a quite sever error in 3 satellite position fix calculation.
+
+.. code-block:: c
 
-The use case parameter determines the target performance.
-A value of 0 indicates single cold start and one indicates multiple hot starts.
+    typedef struct
+    {
+        int32_t latitude          = 0;     // Not known, set to 0
```
jtguggedal, line 180: Perhaps use C-style comments to align with NCS coding style

bsdlib/doc/gnss_extension.rst
```diff
@@ -143,11 +143,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+        int16_t altitude          = meter; // Your actual height in meter
+        uint8_t unc_semimajor     = 127;   // Both set to 127 when latitude
+        uint8_t unc_semiminor     = 127;   // and longitude are unknown
+        uint8_t orientation_major = 0;     // Not known, set to 0
+        uint8_t unc_altitude      = 0;     // Set to 0
+        uint8_t confidence        = 100;   // Set to 100
+    } nrf_gnss_agps_data_location_t;
+
+If the actual altitude of where the device is located changes vs the altitude stored in the
+GNSS engine (e.g. the device is moving around), the accuracy of the 3 sat fix will be degraded.
```
jtguggedal, line 191: `3 sat fix` -> `fix using 3 satellites` maybe?

bjorn-tore-taraldsen: agree

bsdlib/doc/gnss_extension.rst
bjorn-tore-taraldsen: Actually not sure if this line shall be included. here we need to follow how it's done for other bsdlib documentation where we have added feature that has been made available in a specific version of the MFW

bsdlib/doc/gnss_extension.rst
bjorn-tore-taraldsen: I wrote it like this to describe how to set the values in the struct. Thought this was easy and clear compare to write it as text. Open to suggestion here.

bsdlib/doc/gnss_extension.rst
tokangas: This is actually a problem in this document, but I'm a bit confused by NRF_GNSS_USE_CASE_SINGLE_COLD_START and NRF_GNSS_USE_CASE_NORMAL_ACCURACY. These are defined as "0 << 0" and "0 << 1" in nrf_socket.h, so both have a value of 0. Using these as normal bitmask values doesn't work, for example:

`if ((value & NRF_GNSS_USE_CASE_SINGLE_COLD_START) == NRF_GNSS_USE_CASE_SINGLE_COLD_START)`

This would always be true, regardless of the `value`. While setting the bits with these defines works, it's still confusing (at least to me).

bsdlib/doc/gnss_extension.rst
bjorn-tore-taraldsen: @jtguggedal If this struct aproach does make sence - Do you have a suggestion on how we could document which values to set (alternative to use the struct) ?

bsdlib/doc/gnss_extension.rst
```diff
@@ -147,11 +147,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+
+..Note::
+  Normal and Low accuracy is available in MFW 1.2.2 and later
+
+Low Accuracy:
+
+Enabling low accuracy mode will result in the GNSS receiver having a looser acceptance criterion
+for a fix (i.e. error in calculation vs actual position can be larger than Normal Accuracy mode).
+In addition, in the Low Accuracy mode the GNSS receiver may use only 3 satellites to determine a
+fix (in Normal Accuracy mode, it needs 4 or more satellitesi).
```
tokangas, line 169: `satellitesi `-> `satellites`

bsdlib/doc/gnss_extension.rst
```diff
@@ -147,11 +147,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+
+For a possible 3 satellite fix to take place, one of the following criteria must be in place first
+and will be valid for a time window of 24 hour:
```
tokangas, line 172: `hour `-> `hours`

bsdlib/doc/gnss_extension.rst
```diff
@@ -147,11 +147,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+
+ | 1) At least 1 fix must have been taken place using 5 or more satellites. You initiate the GPS to start Low accuracy mode, but the radio will not be able to utilize 3 satellite fixing before the criteria above has taken place. In the following time window, any fix that is done using 5 or more satellites will reset the time window and extend the 24-hour period.
+ |
+ | 2) Injecting Altitude (in meter), Altitude uncertainty (0) , and Confidence (100 percent) to the GPS radio using agps data location struct as described below. To be able to stay in Low Accuracy mode this step has to be repeated at least every 24 hour.
```
tokangas, line 176: `hour `-> `hours`

bsdlib/doc/gnss_extension.rst
```diff
@@ -147,11 +147,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+ |
+ | **Caution:** The Altitude in meter has to have an accuracy in the range of +/- 10’s of meter of your actual height. An erroneous altitude will result in a quite sever error in 3 satellite position fix calculation.
```
tokangas, line 178: `meter `-> `meters`

bsdlib/doc/gnss_extension.rst
tokangas: How about this?

```
nrf_gnss_agps_data_location_t location;

location.latitude          = 0;        /* Unknown, set to 0 */
location.longitude         = 0;        /* Unknown, set to 0 */
location.altitude          = altitude; /* Device's actual altitude in meters */
location.unc_semimajor     = 127;      /* Set to 127 when location is unknown */
location.unc_semiminor     = 127;      /* Set to 127 when location is unknown */
location.orientation_major = 0;        /* Unknown, set to 0 */
location.unc_altitude      = 0;        /* Set to 0 */
location.confidence        = 100;      /* Set to 100 */
```

bsdlib/doc/gnss_extension.rst
tokangas: `actual altitude of where the device is located` -> `actual altitude of the device` ?

bsdlib/doc/gnss_extension.rst
jtguggedal: Yeah, something like that or alternatively
```
nrf_gnss_agps_data_location_t location = {
        .altitude          = altitude, /* Device's actual altitude in meters */
        .unc_semimajor     = 127,      /* Set to 127 when location is unknown */
        .unc_semiminor     = 127,      /* Set to 127 when location is unknown */
        .confidence        = 100,      /* Set to 100 */
};
```

I don't have strong preference as long as it's valid C

bsdlib/doc/gnss_extension.rst
```diff
@@ -147,11 +147,57 @@ Satellites with elevation angle less than the threshold are automatically exclud
...
+        uint8_t orientation_major = 0;     // Not known, set to 0
+        uint8_t unc_altitude      = 0;     // Set to 0
+        uint8_t confidence        = 100;   // Set to 100
+    } nrf_gnss_agps_data_location_t;
+
+If the actual altitude of where the device is located changes vs the altitude stored in the
+GNSS engine (e.g. the device is moving around), the accuracy of the 3 sat fix will be degraded.
+
+This Low Accuracy mode is not the same as the 2D fix you see documented in the NMEA reports.
+All fixes, including the Low Accuracy fix, will be reported as 3D fix.
```
tokangas, line 198: `All fixes, including the Low Accuracy fix, will be reported as 3D fix.` -> `All fixes, including Low Accuracy fixes, will be reported as 3D fixes.`

bsdlib/doc/gnss_extension.rst
tokangas: @jtguggedal I like that approach. Just wondering if we should also include the values which need to be zero for clarity?

jtguggedal: We could include them for clarity. In real code, I usual omit those as they're zeroed implicitly anyway, but no harm in being explicit in the docs.

bsdlib/doc/gnss_extension.rst
bjorn-tore-taraldsen: Had a look in existing BSD documentation and we haven't made such note earlier. We seems to referre to MFW change log for addd features. Then we shouldn't start doing it for this either. I.e. remove it.  (Only place where we could mention this will be Change log)

bjorn-tore-taraldsen: @tokangas Isn't what you point's out here a problem in ".h" and/or ".c" and not as used in this documentation?
We have a pattern on how to define and use bit values in HW definition. Should we copy that for this?

bsdlib/doc/gnss_extension.rst
bjorn-tore-taraldsen: The reason I added the whole struct was to also include the filed that are not mention in text. I.e. the one set to zero. I think that is needed

bsdlib/doc/gnss_extension.rst
tokangas: Those two zero bitmap defines are in nrf_socket.h:

[https://github.com/nrfconnect/sdk-nrfxlib/blob/master/bsdlib/include/nrf_socket.h#L357](https://github.com/nrfconnect/sdk-nrfxlib/blob/master/bsdlib/include/nrf_socket.h#L357)

This documentation should of course reflect what's defined in nrf_socket.h. I think we should only have defines in there which have the corresponding bit set. The defines which do not have any bit set are meaningless and can cause confusion.

I will discuss this with Even. I guess we can keep this as it is now because it reflects what's in nrf_socket.h. If nrf_socket.h is changed, this document needs to be also updated.

evenl: Yes, I agree that thous defines is a bit strange, and I was not sure how to do them. The idea is that every setting should have a define, in this case a zero value in both bit 0 and bit 1 is a valid setting (not just turning something off) as documented in the rpc documentation.

Then you can write code lines like:

config.use_case = NRF_GNSS_USE_CASE_SINGLE_COLD_START | NRF_GNSS_USE_CASE_NORMAL_ACCURACY;

or

config.use_case = NRF_GNSS_USE_CASE_SINGLE_COLD_START | NRF_GNSS_USE_CASE_LOW_ACCURACY;

which would make the lines self documenting. I was not thinking about the use case where the application would read back the configuration and test it. I am also not sure how likely it is, but for it is possible and then these defines would not work.

bsdlib/doc/gnss_extension.rst
```diff
@@ -145,61 +150,61 @@ The elevation mask parameter controls the elevation threshold angle (in degrees)
...
+   location.orientation_major = 0;        /* Unknown, set to 0 */
+   location.unc_altitude      = 0;        /* Set to 0 */
+   location.confidence        = 100;      /* Set to 100 */
+
+If the actual altitude of the device changes with respect to the altitude stored in the GNSS engine (for example, when the device moves around), the accuracy of the position fix using three satellites will be degraded.
+
+The low accuracy mode is different from the 2D fix that is documented in the NMEA reports.
+All fixes, including the low accuracy fixes, will be reported as 3D fixes.
+See the $GPGSA report and number of IDs of SVs used in the position fix to get information of the number of satellites are used for the position fix.
+Only the SV IDs which are not empty are valid.
```
tokangas, line 203: I think the line `Only the SV IDs which are not empty are valid.` could be removed. The satellites used in a fix can be checked from $GPGSA with a 3D fix, i.e. a line which starts with `$GPGSA,A,3`. The previous line could be changed for example like this:

To see the number of satellites used in a fix, check the $GPGSA line starting with $GPGSA,A,3 for a list of SVs used for that fix.

bjorn-tore-taraldsen: My thinking why I wrote it like I did was: you will always get "A,3" regarddles of if this was a 3, 4, or 5 sat fix, but by refering like I did you know explicitly if this was a 3 sat fix or if it has been able to track more than 3 sat.

tokangas: @bjorn-tore-taraldsen The number of satellites listed in the $GPGSA message tell how many satellites were used in a fix and I think it should be self explanatory that empty fields do not count as used satellites, but this is just my opinion. For example:

3 sat fix:
`$GPGSA,A,3,01,03,22,,,,,,,,,,5.25,4.46,2.76,1*18`
4 sat fix:
`$GPGSA,A,3,01,03,22,31,,,,,,,,,5.25,4.46,2.76,1*18`
5 sat fix:
`$GPGSA,A,3,01,03,22,31,32,,,,,,,,5.25,4.46,2.76,1*18`

bjorn-tore-taraldsen: Had a chat with Tommi and agreed that we just remove the last sentence "Only the SV IDs which are not empty are valid" and let the rest stay as is.

bjorn-tore-taraldsen (approved):

tokangas (approved):

bsdlib/doc/gnss_extension.rst
```diff
@@ -96,31 +96,34 @@ The fix interval parameter controls the time between GNSS receiver-starts.
 It also controls the stopping of the GNSS receiver after a valid PVT
 estimate.
 The fix interval parameter determines the mode of navigation.
-There are three different navigation modes available: single-fix,
-continuous, and periodic.
+There are three different navigation modes available: single-fix, continuous, and periodic.
 
-Single-fix navigation mode is engaged by setting the fix interval to 0.
+Single-fix navigation mode is engaged by setting the fix interval to zero.
```
jtguggedal, line 101: I think it would be good to point out in the docs that the retry value is not used in single-fix mode. It will not produce a timeout similar to that of periodic mode. This has been a recurring source of confusion on DevZone.

bsdlib/doc/gnss_extension.rst
```diff
@@ -145,15 +150,63 @@ The elevation mask parameter controls the elevation threshold angle (in degrees)
...
+
+Low accuracy mode
++++++++++++++++++
+
+If you enable the low accuracy mode, the GNSS receiver demonstrates a looser acceptance criterion for a fix.
+The error in position calculation, when compared to the actual position, can be larger than the error in normal accuracy mode.
+In addition, in the low accuracy mode, the GNSS receiver might use only three satellites to determine a fix.
+In normal accuracy mode, four or more satellites are used.
+
+A position fix will be valid for a time duration of 24 hours. For a possible position fix using three satellites, one of the following criteria must be satisfied initially:
```
jtguggedal, line 175: The first sentence here is a bit confusing to me, I'm probably interpreting it wrongly. Is the meaning that a position fix will be valid for 24 hours in the context of the first bullet point below?

bjorn-tore-taraldsen: I Jan Tore is right, this could be missleading. I suggest changing the text to:
"For a possible position fix using three satellites, one of the following criteria must be satisfied initially. When the criteria is met it will be valid for a time duration of 24 hours:"

bjorn-tore-taraldsen (approved):

jtguggedal (approved):

bsdlib/doc/gnss_extension.rst
umapraseeda:
```suggestion
For a possible position fix using three satellites, one of the following criteria must be satisfied initially and when the criterion is met, it will be valid for a time duration of 24 hours:
```

bsdlib/doc/gnss_extension.rst
evenl: Just as side note, why do we discuss NMEA strings in this document which is not enabled by default? Would it make much more sense to discuss the default format which is also much more practical to process programmatically?

umapraseeda (approved):

PR #328, opened 2020-11-13
crypto: Add atomic variant for CC310 locking

lmaciejonczyk:
New lock variant fasten mutually exclusive access to CC3XX resources at the expense of a need to handle potential access failure
in the upper layer.

needed by: https://github.com/nrfconnect/sdk-nrf/pull/3339

### DIFF
--- a/crypto/Kconfig
+++ b/crypto/Kconfig
@@ -34,6 +34,31 @@ if NRF_CC3XX_PLATFORM
 config HW_CC3XX_INTERRUPT
 	bool "Use interrupt version of nrf cc3xx platform library" if !(CC3XX_BACKEND || NRF_CC310_BL)
 
+choice CC3XX_LOCK_VARIANT
+	prompt "CC3XX lock variant"
+	default CC3XX_MUTEX_LOCK
+
+config CC3XX_MUTEX_LOCK
+	bool "Use mutex variables for mutually exclusive access to CC3XX resources"
+	help
+	  Enables exclusive access to the ARM CryptoCell hardware using an RTOS
+	  mutex that has priority inheritance. The mutex lock is safe across threads
+	  and/or interrupts with different priority. Using an RTOS mutex comes
+	  at a cost of lower performance.
+
+config CC3XX_ATOMIC_LOCK
+	bool "Use atomic variables for mutually exclusive access to CC3XX resources"
+	help
+	  Using atomic operations is the fastest way to ensure mutually exclusive
+	  access to the ARM CryptoCell hardware.
+	  Warning: If this configuration is set, every execution requiring use of
+	  the ARM CryptoCell hardware must happen in the same priority. Calling into
+	  mbed TLS APIs from a higher priority while an ongoing operation will lead
+	  to undefined behavior. It is highy recommended to to do all cryptographic
+	  operations in one single thread if this configuration is set.
+
+endchoice
+
 endif
 
 endmenu
--- a/crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_mutex_zephyr.c
+++ b/crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_mutex_zephyr.c
@@ -20,6 +20,8 @@
  */
 extern nrf_cc3xx_platform_abort_apis_t platform_abort_apis;
 
+#if CONFIG_CC3XX_MUTEX_LOCK
+
 /** @brief Definition of mutex for symmetric cryptography
  */
 K_MUTEX_DEFINE(sym_mutex_int);
@@ -28,10 +30,6 @@ K_MUTEX_DEFINE(sym_mutex_int);
  */
 K_MUTEX_DEFINE(asym_mutex_int);
 
-/** @brief Definition of mutex for random number generation
-*/
-K_MUTEX_DEFINE(rng_mutex_int);
-
 /** @brief Definition of mutex for power mode changes
 */
 K_MUTEX_DEFINE(power_mutex_int);
@@ -39,6 +37,32 @@ K_MUTEX_DEFINE(power_mutex_int);
 /** @brief Definition of mutex for heap allocations
  */
 K_MUTEX_DEFINE(heap_mutex_int);
+
+#elif CONFIG_CC3XX_ATOMIC_LOCK
+
+/** @brief Definition of mutex for symmetric cryptography
+ */
+atomic_t sym_mutex_int;
+
+/** @brief Definition of mutex for asymmetric cryptography
+ */
+atomic_t asym_mutex_int;
+
+/** @brief Definition of mutex for power mode changes
+*/
+atomic_t power_mutex_int;
+
+/** @brief Definition of mutex for heap allocations
+ */
+atomic_t heap_mutex_int;
+
+#else
+#error "Improper configuration of the lock variant!"
+#endif
+
+/** @brief Definition of mutex for random number generation
+*/
+K_MUTEX_DEFINE(rng_mutex_int);
 
 /** @brief Arbritary number of mutexes the system suppors
  */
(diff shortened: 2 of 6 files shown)


crypto/Kconfig
```diff
@@ -34,6 +34,28 @@ if NRF_CC3XX_PLATFORM
...
 	bool "Use interrupt version of nrf cc3xx platform library" if !(CC3XX_BACKEND || NRF_CC310_BL)
 
+choice CC310_LOCK_VARIANT
+	prompt "CC310 lock variant"
+	default CC310_MUTEX_LOCK
+
+config CC310_MUTEX_LOCK
+	bool "Use mutex variables for mutually exclusive access to CC310 resources"
+	help
+	  Locking by mutex is the most flexible solution for providing mutually
```
frkv, line 44: Enables exclusive access to the Arm CryptoCell HW using an RTOS mutex that has priority inheritance. The mutex lock is safe across threads and/or interrupts with different priority. Using an RTOS mutex comes at a cost.

crypto/Kconfig
```diff
@@ -34,6 +34,30 @@ if NRF_CC3XX_PLATFORM
...
 	bool "Use interrupt version of nrf cc3xx platform library" if !(CC3XX_BACKEND || NRF_CC310_BL)
 
+choice CC3XX_LOCK_VARIANT
+	prompt "CC3XX lock variant"
+	default CC3XX_MUTEX_LOCK
+
+config CC3XX_MUTEX_LOCK
+	bool "Use mutex variables for mutually exclusive access to CC3XX resources"
+	help
+	  Enables exclusive access to the Arm CryptoCell HW using an RTOS mutex
```
edmont, line 44:
```suggestion
	  Enables exclusive access to the ARM CryptoCell hardware using an RTOS mutex
```

crypto/Kconfig
```diff
@@ -34,6 +34,30 @@ if NRF_CC3XX_PLATFORM
...
+	  that has priority inheritance. The mutex lock is safe across threads
+	  and/or interrupts with different priority. Using an RTOS mutex comes
+	  at a cost.
```
edmont, line 47: Be more specific here? Which cost?

crypto/Kconfig
```diff
@@ -34,6 +34,30 @@ if NRF_CC3XX_PLATFORM
...
+
+config CC3XX_ATOMIC_LOCK
+	bool "Use atomic variables for mutually exclusive access to CC3XX resources"
+	help
+	  Locking by atomic is the fastest solution for providing mutually
+	  exclusive access to the resources. It is the most beneficial for high
+	  frequency requests from one thread at the same time. Access
+	  the resource failure for requests from different threads trying to get
```
edmont, line 55: Failure accesses to the resource...

edmont (changes requested):
Some editorial suggestions. But please check indentation, it's not uniform across the files.

crypto/Kconfig
lmaciejonczyk: @b-gent can you help refactor this sentence to be more readable?

crypto/Kconfig
lmaciejonczyk:
```suggestion
	  at a cost of lower performance.
```
?

edmont: Okay.

crypto/Kconfig
b-gent: Note that if different threads try to get the same resource at the same time, the access will fail. This situation must be resolved by higher application layer.

@lmaciejonczyk  does that work?

crypto/Kconfig
```diff
@@ -34,6 +34,30 @@ if NRF_CC3XX_PLATFORM
...
+	  the resource at the same time need to be resolved by higher application
+	  layer.
```
lmaciejonczyk, line 57:
```suggestion
	  frequency requests from one thread at the same time.
	  Note that if different threads try to get the same resource at the same time, the access will fail.
	  This situation must be resolved by higher application layer.
```

b-gent (approved):

edmont (changes requested):
LGTM, but please replace tabs by four spaces.

lmaciejonczyk:
> LGTM, but please replace tabs by four spaces.

Why? Tabs are commonly used in Kconfig files over all projects.

lmaciejonczyk:
@frkv If you are ok with the changes approve this PR so that we can proceed with merging.

edmont:
> Why? Tabs are commonly used in Kconfig files over all projects.

I mean in the C files. Currently there are mixed styles within the same files.

lmaciejonczyk:
> > Why? Tabs are commonly used in Kconfig files over all projects.
> 
> I mean in the C files. Currently there are mixed styles within the same files.

It's done. Check and approve if it's ok now please.

edmont (approved):

crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_mutex_zephyr.c
```diff
@@ -227,6 +228,47 @@ static int32_t mutex_unlock_platform(nrf_cc3xx_platform_mutex_t *mutex) {
...
+}
+
+static int32_t mutex_lock_platform(nrf_cc3xx_platform_mutex_t *mutex)
+{
+    /* Ensure that the mutex param is valid (not NULL) */
+    if (mutex == NULL) {
+            return NRF_CC3XX_PLATFORM_ERROR_PARAM_NULL;
+    }
+
+    return atomic_cas((atomic_t *)mutex, 0, 1) ?
```
frkv, line 254: This will still need to add the test for uninitialized mutexes (i.e. mutexes in contexts that enable their own mutex because of MBEDTLS_THREADING_C)

_Note that the scheme for the flags should probably be "is_set(SOME_FLAG)" instead of == since we will need to add another flag type to indicate distinction between atomic mutex and regular kernel mutex (see different comment)_

crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_mutex_zephyr.c
```diff
@@ -227,6 +228,47 @@ static int32_t mutex_unlock_platform(nrf_cc3xx_platform_mutex_t *mutex) {
     return NRF_CC3XX_PLATFORM_SUCCESS;
 }
 
+#elif CONFIG_CC3XX_ATOMIC_LOCK
+
+atomic_t sym_mutex;
+atomic_t asym_mutex;
+atomic_t rng_mutex;
```
frkv, line 235: The rng_mutex is very difficult to control since it can be called by anything in Zephyr that needs entropy. This merits handing RNG in a special way. Please change rng_mutex back to use a kernel mutex and handle the additional type difference with a new flag in nrf_cc3xx_platform_mutex.h

_The new flag could be called NRF_CC3XX_PLATFORM_MUTEX_MASK_IS_ATOMIC, so there is a way to distinguish between regular kernel mutex and atomic mutex_

frkv (changes requested):
Requests needs to be addressed before PR can be approved

crypto/Kconfig
```diff
@@ -34,6 +34,30 @@ if NRF_CC3XX_PLATFORM
...
+	help
+	  Enables exclusive access to the ARM CryptoCell hardware using an RTOS
+	  mutex that has priority inheritance. The mutex lock is safe across threads
+	  and/or interrupts with different priority. Using an RTOS mutex comes
+	  at a cost of lower performance.
+
+config CC3XX_ATOMIC_LOCK
+	bool "Use atomic variables for mutually exclusive access to CC3XX resources"
+	help
+	  Locking by atomic is the fastest solution for providing mutually
```
frkv, line 52: Using atomic operations is the fastest way to ensure mutually exclusive access to the ARM CryptoCell hardware.

Warning: If this configuration is set, every execution requiring use of the ARM CryptoCell hardware must happen in the same priority. Calling into mbed TLS APIs from a higher priority while an existing operation leads to undefined behavior.  It is highy recommended to to do all crypto operations in one single thread if this configuration is set.

frkv (approved):
LGTM

tejlmand:
where is the manifest PR that references this PR ?

MarekPorwisz:
I've cherry-picked and added changes from https://github.com/nrfconnect/sdk-nrfxlib/pull/350 here

tejlmand:
@lmaciejonczyk could you please check the compliance failures, seems to be coding violations.

lmaciejonczyk:
> @lmaciejonczyk could you please check the compliance failures, seems to be coding violations.

How can I identify it?

tejlmand (approved):

PR #360, opened 2021-01-11
doc: nrf_modem: Add note regarding cipher suite support

rlubos:
Add information of the supported cipher suites as a reference to the
modem documentation. Add simple example on how cipher suites used for
TLS session can be configured with socket options.

NCSDK-7725

Signed-off-by: Robert Lubos <robert.lubos@nordicsemi.no>

### DIFF
--- a/doc/links.txt
+++ b/doc/links.txt
@@ -16,6 +16,8 @@
 
 .. _`nRF9160 modem firmware zip file`: https://www.nordicsemi.com/Products/Low-power-cellular-IoT/nRF9160/Download#98C9E2578566420786ABD40B695FDB9B
 
+.. _`nRF9160 modem TLS cipher suites`: https://www.nordicsemi.com/Products/Low-power-cellular-IoT/nRF9160/Download#infotabs
+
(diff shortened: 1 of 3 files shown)


lemrey (approved):
Thanks

bjorn-tore-taraldsen (commented):
I think we now need a page name different than "security_tags:" for this page itself.

rlubos:
> I think we now need a page name different than "security_tags:" for this page itself.

How about `TLS configuration` for the page name, and then `Security tags`  subtitle within the page for the existing content?

bjorn-tore-taraldsen:
> > I think we now need a page name different than "security_tags:" for this page itself.
> 
> How about `TLS configuration` for the page name, and then `Security tags` subtitle within the page for the existing content?

rlubos:
Updated the PR based on the offline feedback from @bjorn-tore-taraldsen.

bjorn-tore-taraldsen (approved):

umapraseeda (approved):

PR #366, opened 2021-01-18
nrf_802154_sl: Add initial nRF53 multiprotocol support

pkwiek:
This commit adds initial support for multiprotocol for nRF53.

Signed-off-by: Pawel Kwiek <pawel.kwiek@nordicsemi.no>


nrf_802154_sl/CHANGELOG.rst
```diff
@@ -12,11 +12,10 @@ All notable changes to this project are documented in this file.
 nRF 802.15.4 Service Layer 0.3.0
```
hubertmis, line 12: It seems we should mark it here as version 0.4.0 with its changelog entry.

nrf_802154_sl/CHANGELOG.rst
```diff
@@ -12,11 +12,10 @@ All notable changes to this project are documented in this file.
...
 ********************************
 
-* PA/LNA implementation has been moved to MPSL.
-  Obsolete implementation and API have been removed.
+PA/LNA implementation has been moved to MPSL. Obsolete implementation and API have been removed.
 
 Removed
-=======
+=====
```
hubertmis, line 18: It looks like a bug (it's too short)

hubertmis (changes requested):
Looks good! Thanks for updating licenses.

I think we could improve changelog for this patch.

nrf_802154_sl/CHANGELOG.rst
pkwiek: Fixed

nrf_802154_sl/CHANGELOG.rst
pkwiek: Please check if the new entry is appropriate.

jciupis (approved):

hubertmis (approved):

FrancescoSer (changes requested):
@pkwiek @jciupis  Could you add some additional detail on how to utilize this service layer? There are 35 files changed in this PR (and about the same number in the previous one that added this SL to nrfxlib) and the doc files are way too synthetic in comparison.

Some info that would be useful:
- Why would any user actually use this? Example use cases of the 4 features that were implemented here
- Examples of how to actually implement them.
- Integration notes?
- Anything else that the user should know to easily use this

hubertmis:
@FrancescoSer, actually the only user of this library is 802.15.4 radio driver. No one else should use it. Because of that I don't think we need documentation here. Maybe just a note like "This library is used only by 802.15.4 radio driver. Don't use it"

FrancescoSer (approved):
Added distinction between NCS users and other users of the RD

hubertmis:
> Added distinction between NCS users and other users of the RD

Thanks! The description is much clearer now

e-rk:
https://github.com/nrfconnect/sdk-nrf/pull/3719 is green. Please merge.

PR #367, opened 2021-01-18
ci: compliance: fetch all west modules

utzig:
- Fetch all west modules when running compliance check.
- Remove DeviceTree module which is not used.


carlescufi:
Closing since it's not longer applicable.

PR #369, opened 2021-01-20
zboss: Change reset type enum to a set of macros

FilipZajdel:
Align with ZBOSS by replacing reset type enum with a set of macros.

Signed-off-by: Filip Zajdel <filip.zajdel@nordicsemi.no>


tomchy (changes requested):
Thank you for the quick PR!
We needed this change ASAP, thus those changes are already merged here: https://github.com/nrfconnect/sdk-nrfxlib/commit/2198a0984daeece7adc4b2bea2f1863463e16285

PR #371, opened 2021-01-20
actions: Align compliance with other repos

carlescufi:
Align compliance with the one in upstream zephyr, sdk-zephyr and
sdk-nrf.

Signed-off-by: Carles Cufi <carles.cufi@nordicsemi.no>

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -7,57 +7,86 @@ jobs:
     runs-on: ubuntu-latest
     name: Run compliance checks on patch series (PR)
     steps:
-      - name: Checkout the code
-        uses: actions/checkout@v2
-        with:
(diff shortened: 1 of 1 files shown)


utzig (approved):
LGTM. Btw, `-m DeviceTree` does not do anything anymore. Also I still find using `-m Kconfig` somewhat debatable, because it only makes real sense when running on `sdk-nrf`, because that's where all configuration build is glued together.

carlescufi:
> LGTM. Btw, `-m DeviceTree` does not do anything anymore. Also I still find using `-m Kconfig` somewhat debatable, because it only makes real sense when running on `sdk-nrf`, because that's where all configuration build is glued together.

I see, thanks. I'll leave it as-is for consistency with both upstream and the rest of repos, but I'll keep it in mind.

PR #374, opened 2021-01-21
actions: compliance: Remove checkpatch.

carlescufi:
The --exclude functionality doesn't seem to work correctly for multiple
folders, so remove the check for coding style altogether given that we
cannot enforce it selectively.


carlescufi:
Not the right fix.

PR #375, opened 2021-01-21
nrf_security: Make use of mbedtls user config

pdunaj:
Use zephyr's option to pass path to user config.

Jira:NCSDK-7689

Signed-off-by: Pawel Dunaj <pawel.dunaj@nordicsemi.no>

### DIFF
--- a/nrf_security/configs/nrf-config.h.template
+++ b/nrf_security/configs/nrf-config.h.template
@@ -3721,6 +3721,7 @@ it is (2^48 - 1), our restriction is :  (int - 0xFFFF - 0xF).*/
  * Allow user to override any previous default.
  *
  */
+#cmakedefine MBEDTLS_USER_CONFIG_FILE	"@MBEDTLS_USER_CONFIG_FILE@"
 #if defined(MBEDTLS_USER_CONFIG_FILE)
 #include MBEDTLS_USER_CONFIG_FILE
 #endif
--- a/nrf_security/configs/nrf-config.h.template
+++ b/nrf_security/configs/nrf-config.h.template
@@ -1561,7 +1561,7 @@
  *          configuration of this extension).
  *
  */
-//#define MBEDTLS_SSL_RENEGOTIATION
+#cmakedefine MBEDTLS_SSL_RENEGOTIATION
 
 /**
  * \def MBEDTLS_SSL_SRV_SUPPORT_SSLV2_CLIENT_HELLO
@@ -1687,7 +1687,7 @@
  *
  * Comment this macro to disable support for ALPN.
  */
-//#define MBEDTLS_SSL_ALPN
+#cmakedefine MBEDTLS_SSL_ALPN
 
 /**
  * \def MBEDTLS_SSL_DTLS_ANTI_REPLAY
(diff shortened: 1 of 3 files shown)


nrf_security/configs/nrf-config.h.template
```diff
@@ -3721,8 +3721,8 @@ it is (2^48 - 1), our restriction is :  (int - 0xFFFF - 0xF).*/
  * Allow user to override any previous default.
  *
  */
-#if defined(MBEDTLS_USER_CONFIG_FILE)
-#include MBEDTLS_USER_CONFIG_FILE
+#if defined(CONFIG_MBEDTLS_USER_CONFIG_FILE)
```
frkv, line 3724: We don't add kconfig in the template
The mechanism required is the addition of something that converts CONFIG_MBEDTLS_XXXX into MBEDTLS_XXX

I will contact you and show what is needed

frkv: Adding comment here as well for posterity:
We convert the configuration from **CONFIG_MBEDTLS_XXXX** in **nrf_security/cmake/kconfig_mbedtls_configure.cmake** with the macro **kconfig_mbedtls_config** which will strip CONFIG_ from the configuration (and make it the same format as mbed TLS expects):

It is necessary to make **#cmakedefine MBEDTLS_XXXX** inside nrf-config.h.template that will be replaced either with a defined version or an unset version (when the config is not to be present)

pdunaj: I fixed it and it's good thing that I did. There is another problem about it - the header file is not seen for some sources
```
/home/pdunaj/work/ncs/mbedtls/library/ecjpake.c:25:
modules/nrfxlib/nrf_security/include/nrf-config.h:3724:34: fatal error: config-tls-user.h: No such file or directory
 3724 | #define MBEDTLS_USER_CONFIG_FILE "config-tls-user.h"
```

The problem is the include paths added by nrf_security are very limitted and library does not see directory with zephyr generated files (where we put our config file). I also don't think it will be possible to add header directory by the application as it is already pass the boilerplate :/

Seems I will have to go and add every missing option (I planned to do it anyway but it seems I will have no other choice).

pdunaj: I have added the missing config files but left the user config option too - just in case it can be useful later on.

pdunaj:
Hi @frkv , can you re-review?

frkv (approved):
LGTM, but please open a nrf manifest PR so we can 

Add the link to the manifest PR in the description here, as well as a link in the manifest PR to this pull request for easy jumping in between.

frkv (approved):

Vge0rge (approved):

pdunaj:
@frkv done

pdunaj:
@frkv there seems to be an error due to an unrelated kconfig warning (something with `CC3XX_LOCK_VARIANT`).

If it can be merged to nrxlib it would be great. We don't need it in nrf yet as we use separate repo at the moment.

pdunaj:
rebased

pdunaj:
@frkv  tests are green :)
https://github.com/nrfconnect/sdk-nrf/pull/3756

PR #376, opened 2021-01-21
nrf_security: Fix linking to c library

pdunaj:
Add explicit dependenciy between libraries to make sure
no symbol is missing.

Jira:NCSDK-7689

Signed-off-by: Pawel Dunaj <pawel.dunaj@nordicsemi.no>


pdunaj:
@tejlmand This seems to do the trick for missing printf.

Still after the latest upmerge I found another issue that may be more complicated. In vanilla_base library there is a missing symbol (mbedtls_ecp_write_key) that is provided by cc310 (from file mbedtls/library/ecp.c).

The symbol is not in the library however. The problem is that even if I enable the symbol this file is not rebuilt. Even if I remove the build directory. Could it be cached? From what I see this is part of the extracted archives (build-nrf52840dk_nrf52840/modules/nrfxlib/nrf_security/src/mbedtls/cc310/extracted_archives/ecp.c.obj) so maybe something else is sourcing this file without the symbol. I will try to catch with you tomorrow about it.

frkv (approved):

pdunaj:
Hi @tejlmand , can this one be merged?

pdunaj:
rebased

pdunaj:
rebased

tejlmand (approved):

pdunaj:
rebased

pdunaj:
test fail at manifest update PR
```
FAILED: modules/nrfxlib/nrf_security/src/mbedcrypto_glue/libmbedcrypto_oberon_renamed.a 
....
/opt/zephyr-sdk/arm-zephyr-eabi/bin/arm-zephyr-eabi-objcopy --redefine-syms /home/work/twister-out/nrf52840dk_nrf52840/crypto/crypto.cc3xx_oberon/modules/nrfxlib/nrf_security/src/mbedcrypto_glue/symbol_rename_cc3xx.txt /home/work/twister-out/nrf52840dk_nrf52840/crypto/crypto.cc3xx_oberon/modules/nrfxlib/nrf_security/src/mbedtls/cc310/libmbedcrypto_cc3xx.a;-lc  libmbedcrypto_cc3xx_renamed.a
```
I have no idea what is wrong as it builds for us without any problems. Maybe it is one of many variants of how the crypto libraries are built.

Temporarily I have added the needed dependency link in the application cmakelists.txt so we are unblocked. 

@frkv @tejlmand If you tell me what's wrong I can update this PR but maybe it would be enough to wait for the change that removes printf reference from the precompiled library.

tejlmand:
This is wrong:
```
libmbedcrypto_cc3xx.a;-lc 
```
there should never be a `;` in the command line.
Must be a list somewhere that is not handled properly.

PR #377, opened 2021-01-21
checkpatch: Fix checkpatch.conf format

carlescufi:
For whatever reason checkpatch doesn't like a comma-separated list for
the set of excludes.


pkwiek (approved):
Great! Thank you!

PR #380, opened 2021-01-25
softdevice_controller: Rename to SOFTDEVICE_CONTROLLER_PERIPHERAL

rugeGerritsen:
Using PERIPHERAL instead of PERIPH matches the library name and
the naming convention of the other configurations.


joerchan (approved):

rugeGerritsen:
This configuration is not used anywhere in the SDK, so we do not need to create manifest-update PR for this one

martintv (approved):

PR #381, opened 2021-01-26
nrf_802154_sl: Enable die temperature measurement

pkwiek:
This commit enables die temperature sensor when using SL library. The
sensor is used by the 802.15.4 Radio Driver.

Signed-off-by: Pawel Kwiek <pawel.kwiek@nordicsemi.no>

### DIFF
--- a/nrf_802154_sl/Kconfig
+++ b/nrf_802154_sl/Kconfig
@@ -17,6 +17,8 @@ config NRF_802154_SL
 	select MPSL
 	select ZERO_LATENCY_IRQS
 	select DYNAMIC_INTERRUPTS
+	select SENSOR
+	select TEMP_NRF5_MPSL
 	help
 		Use of Nordic Semiconductor proprietary implementation of nRF 802.15.4 Service Layer.
 		This implementation enables advanced features of nRF 802.15.4 Radio Driver.
--- a/nrf_802154_sl/include/platform/temperature/nrf_802154_temperature.h
+++ b/nrf_802154_sl/include/platform/temperature/nrf_802154_temperature.h
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2018-2021, Nordic Semiconductor ASA
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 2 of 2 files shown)


nrf_802154_sl/Kconfig
```diff
@@ -17,6 +17,8 @@ config NRF_802154_SL
 	select MPSL
 	select ZERO_LATENCY_IRQS
 	select DYNAMIC_INTERRUPTS
+	select SENSOR
```
hubertmis, line 20: Why is it selected only for `NRF_802154_SL`, not for `NRF_802154_RADIO_DRIVER`? I thought RSSI correction is implemented in `NRF_802154_RADIO_DRIVER` regardless SL mode is selected.

pkwiek: Do you mean only `select SENSOR` or the two added lines?

hubertmis: I think both are necessary to do RSSI correction, right? Because of that I'm thinking if both belong here or in `NRF_802154_RADIO_DRIVER`

pkwiek: Right. However, currently there are two sources of temperature: `TEMP_NRF5` and `TEMP_NRF5_MPSL`. Those two cannot be used simultaneously. `NRF_802154_SL` is used with `MPSL`, hence this option is selected here. For `NRF_802154_SL_OPENSOURCE`, `TEMP_NRF5_MPSL` cannot be used and `TEMP_NRF5` selection is to be introduced to Zephyr by a separate PR.

hubertmis: Do I understand correctly `...SL_OPENSOURCE` selects `SENSOR` and `TEMP_NRF5_MPSL`, while `...SL` selects `SENSOR` and `TEMP_NRF5`?

pkwiek: It is exactly like that.

hubertmis (approved):

rlubos:
Please open a corresponding manifest update in `sdk-nrf`.

jciupis (approved):

pkwiek:
Force-push to add temperature platform header.

tejlmand:
if this PR is still valid, could you then please rebase.

PR #384, opened 2021-01-27
Add braces around ARM_MBEDTLS_PATH in nrf_security/CMakeLists.txt

bamkrs:
Otherwise the string "ARM_MBEDTLS_PATH" gets passed to `get_mbedtls_dir` instead the value of the variable `ARM_MBEDTLS_PATH`.
Without the braces, its not possible to pass the actual path to `get_mbedtls_dir`.

### DIFF
--- a/nrf_security/CMakeLists.txt
+++ b/nrf_security/CMakeLists.txt
@@ -26,7 +26,7 @@ target_include_directories(mbedtls_common INTERFACE
   ${generated_includes}
 )
 
-get_mbedtls_dir(ARM_MBEDTLS_PATH)
+get_mbedtls_dir(${ARM_MBEDTLS_PATH})
 
 set(mbedcrypto_glue_include_path
   "${CMAKE_CURRENT_LIST_DIR}/include/mbedcrypto_glue/mbedtls"


tejlmand (changes requested):
Thanks a lot for pointing out this issue.

> Otherwise the string "ARM_MBEDTLS_PATH" gets passed to get_mbedtls_dir instead the value of the variable ARM_MBEDTLS_PATH.
> Without the braces, its not possible to pass the actual path to get_mbedtls_dir.

The problem is not with the braces, as those must not be there, as we are returning a value (hence the `PARENT_SCOPE`).
https://github.com/nrfconnect/sdk-nrfxlib/blob/9017da739a02e51a3e152df8d2efc62bb4111b2f/common.cmake#L64

Instead the problem was that the function argument name hides the value of the real variable.
This is now fixed here:
https://github.com/nrfconnect/sdk-nrfxlib/pull/396

bamkrs:
> Instead the problem was that the function argument name hides the value of the real variable.
> This is now fixed here:
> #396

Oh, so I just misunderstood the intention of the function! Now this is making a lot more sense. Thanks for fixing it!

tejlmand:
> Thanks for fixing it!

And thanks for finding it.

PR #392, opened 2021-02-01
compliancecheck: using KconfigBasic compliance check instead of Kconfig

tejlmand:
Using `compliancecheck` in sdk-nrf generates >500 undefined Kconfig
symbol references. Many of those are false warnings, but to allow time
for proper cleanup, then this commit resorts to only do KconfigBasic
check.

This means that undefined Kconfig symbols in the Kconfig tree are
reported but not outside the tree, such as source code, rst files, conf
files.

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -60,7 +60,7 @@ jobs:
         # debug
         ls -la
         git log --pretty=oneline | head -n 10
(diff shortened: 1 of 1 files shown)


tejlmand:
no manifest required.
nrfxlib compliance test only.

thst-nordic (approved):

frkv (approved):
LGTM

PR #394, opened 2021-02-02
zboss: Simplify ZBOSS CMake targets

tomchy:
After adding a dependency on the nrfxlib module inside ZBOSS sources,
there is no need to define zboss-extensions CMake target.

Due to the Kconfig option override rules, the library type symbol has to
be dependent on the existance of ZBOSS sources in the west project.
After adding the dependency - the overriding Kconfig will be processed
after the one that intorduces option with defaults.

Signed-off-by: Tomasz Chyrowicz <tomasz.chyrowicz@nordicsemi.no>

### DIFF
--- a/zboss/Kconfig
+++ b/zboss/Kconfig
@@ -6,10 +6,13 @@
 
 menu "ZBOSS library configuration"
 
+config ZBOSS_SOURCES_AVAILABLE
+	bool
+
 # Select between the production and development version of libraries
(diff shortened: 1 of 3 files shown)


tejlmand (approved):
looks like nice cleanup.
Thanks

rlubos (approved):

sebastiandraus (approved):

PR #397, opened 2021-02-03
crypto: Link platform and mbecrypto libs when nrf security is disabled

Vge0rge:
-crypto: nrf_security Kconfig cleanup 
-crypto: Link platform and mbecrypto libs when nrf security is disabled
-crypto: Link mbedcrypto_cc3xx_imported when NRF_CC3XX_PLATFORM is set

Ref: NCSDK-8125

Relevant nrf PR: [sdk-nrf/pull/3793](https://github.com/nrfconnect/sdk-nrf/pull/3793)

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -113,7 +113,7 @@ if (CONFIG_NRF_CC3XX_PLATFORM)
   zephyr_sources(${NRF_CC3XX_PLATFORM_BASE}/src/nrf_cc3xx_platform_mutex_zephyr.c)
 endif()
 
-if (CONFIG_CC3XX_BACKEND)
+if (CONFIG_NRF_CC3XX_MBEDCRYPTO)
   if(CONFIG_SOC_NRF5340_CPUAPP)
     set (CC3XX_ARCH cc312)
   else()
@@ -146,10 +146,16 @@ if (CONFIG_CC3XX_BACKEND)
                              "${NRF_CC3XX_BASE}/include"
                              "${NRF_CC3XX_BASE}/include/mbedtls")
   target_link_libraries(mbedcrypto_cc3xx_imported INTERFACE platform_cc3xx)
+
+  if (NOT CONFIG_NORDIC_SECURITY_BACKEND)
+    target_link_libraries(platform_cc3xx INTERFACE mbedcrypto_cc3xx_imported)
+  endif()
+
   if (NOT CONFIG_NRF_SECURITY_MULTI_BACKEND)
     # Only CC3XX backend is selected for mbedtls API.
     # Thus export the alt headers through Zephyr interface library.
     zephyr_include_directories(${NRF_CC3XX_BASE}/include/mbedtls
                                ${NRF_CC3XX_BASE}/include)
   endif()
+
 endif()
--- a/crypto/Kconfig
+++ b/crypto/Kconfig
@@ -23,12 +23,21 @@ if NRF_CC310_BL
 	default n # Only the no-interrupts version currently verified to work with Zephyr.
 endif
 
+config NRF_CC3XX_MBEDCRYPTO
+	bool
+	depends on HW_CC3XX
+	help
+	  Use nrf_cc3xx_mbedcrypto library. This is automatically ensured
+	  when using nrf security with the cc3xx backend.
+	  Only select this setting without nrf security for special use
+	  cases, such as csrng support without nrf security.
+
 menuconfig NRF_CC3XX_PLATFORM
 	bool "nrf_cc3xx_platform HW crypto library for nRF devices with CryptoCell CC3xx."
 	select NRFXLIB_CRYPTO
 	depends on HW_CC3XX
 	help
-		To use, link with nrfxlib_crypto in CMake.
+	  To use, link with nrfxlib_crypto in CMake.
 
 if NRF_CC3XX_PLATFORM
 
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -145,6 +145,7 @@ menuconfig CC3XX_BACKEND
 	depends on CRYPTOCELL_USABLE
 	default y
 	select NRF_CC3XX_PLATFORM
+	select NRF_CC3XX_MBEDCRYPTO
 	help
 	  Enable cc3xx backend
 
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -152,12 +152,12 @@ if CC3XX_BACKEND
 
 config CC310_BACKEND
 	bool
-	depends on CC3XX_BACKEND && CRYPTOCELL_CC310_USABLE
+	depends on CRYPTOCELL_CC310_USABLE
 	default y
 
 config CC312_BACKEND
 	bool
-	depends on CC3XX_BACKEND && CRYPTOCELL_CC312_USABLE
+	depends on CRYPTOCELL_CC312_USABLE
 	default y
 
 


crypto/CMakeLists.txt
```diff
@@ -111,14 +111,7 @@ if (CONFIG_NRF_CC3XX_PLATFORM)
   #
   zephyr_sources(${NRF_CC3XX_PLATFORM_BASE}/src/nrf_cc3xx_platform_abort_zephyr.c)
   zephyr_sources(${NRF_CC3XX_PLATFORM_BASE}/src/nrf_cc3xx_platform_mutex_zephyr.c)
-endif()
 
-if (CONFIG_CC3XX_BACKEND)
-  if(CONFIG_SOC_NRF5340_CPUAPP)
```
frkv, line 117: Setting of CC3XX_ARCH should not be removed, this is used in line 122

frkv (approved):
LGTM

frkv (changes requested):
Typo I didn't catch in the PR. Must unapprove

frkv (changes requested):

frkv (approved):

crypto/CMakeLists.txt
```diff
@@ -111,14 +111,7 @@ if (CONFIG_NRF_CC3XX_PLATFORM)
...
-    set (CC3XX_ARCH cc312)
-  else()
-    set (CC3XX_ARCH cc310)
-  endif()
```
tejlmand, line 121: I don't see how this change relates to the commit message ?
> crypto: Link mbedcrypto library when cc3xx platform is enabled
>
> -Changed CMakeLists.txt in crypto to always link the mbedcrypto library

if this is a pure cleanup, then please make such cleanups in independent commits to make them easier to review.

crypto/CMakeLists.txt
```diff
@@ -111,14 +111,7 @@ if (CONFIG_NRF_CC3XX_PLATFORM)
...
   set(NRF_CC3XX_BASE ${CMAKE_CURRENT_SOURCE_DIR}/nrf_${CC3XX_ARCH}_mbedcrypto)
```
tejlmand, line 115: when removing `CC3XX_ARCH` variable, above this line will become:
```
nrf__mbedcrypto
```
which i'm confident will break, as there are two `_` after each other.

crypto/CMakeLists.txt
```diff
@@ -146,10 +140,17 @@ if (CONFIG_CC3XX_BACKEND)
...
-    # Only CC3XX backend is selected for mbedtls API.
-    # Thus export the alt headers through Zephyr interface library.
-    zephyr_include_directories(${NRF_CC3XX_BASE}/include/mbedtls
-                               ${NRF_CC3XX_BASE}/include)
-  endif()
+
+  if (NOT CONFIG_NORDIC_SECURITY_BACKEND)
+    target_link_libraries(platform_cc3xx INTERFACE mbedcrypto_cc3xx_imported)
+
+    if (NOT CONFIG_NRF_SECURITY_MULTI_BACKEND)
```
tejlmand, line 147: This line: https://github.com/nrfconnect/sdk-nrfxlib/blob/4c3b97997ebf041c3936300a280a4061b9f268c4/nrf_security/Kconfig#L345

is inside:
```
if NORDIC_SECURITY_BACKEND
...
 if NRF_SECURITY_MULTI_BACKEND 
...
endif # NORDIC_SECURITY_BACKEND

```

so `CONFIG_NRF_SECURITY_MULTI_BACKEND` is always `n` when `CONFIG_NORDIC_SECURITY_BACKEND` is `n`, so why is this test needed ?

tejlmand (changes requested):
please explain how this code is supposed to work.

crypto/CMakeLists.txt
Vge0rge: This is a pure cleanup indeed. I will split it to a different commit then.

crypto/CMakeLists.txt
Vge0rge: I didn't remove the variable CC3XX_ARCH. I merged two if statements to one and both of them set these variables at the top.  https://github.com/nrfconnect/sdk-nrfxlib/blob/71476ebb6101351a27559ea07cddd8787a9ff3fc/crypto/CMakeLists.txt#L80-L84

crypto/CMakeLists.txt
Vge0rge: You are correct, I will move the if not  if NRF_SECURITY_MULTI_BACKEND outside of the previous if.

crypto/CMakeLists.txt
```diff
@@ -146,10 +139,15 @@ if (CONFIG_CC3XX_BACKEND)
...
+  endif()
+
   if (NOT CONFIG_NRF_SECURITY_MULTI_BACKEND)
     # Only CC3XX backend is selected for mbedtls API.
     # Thus export the alt headers through Zephyr interface library.
     zephyr_include_directories(${NRF_CC3XX_BASE}/include/mbedtls
                                ${NRF_CC3XX_BASE}/include)
   endif()
-endif()
+endif()
```
oyvindronningstad, line 153: nit: missing newline

oyvindronningstad (approved):

crypto/CMakeLists.txt
tejlmand: right, I see that, because you also remove the guard: `if (CONFIG_CC3XX_BACKEND)`

But that means that the code part that used to be guarded by `if (CONFIG_CC3XX_BACKEND)` is now guarded by `if (CONFIG_NRF_CC3XX_PLATFORM)` instead, but I can have:
```
CONFIG_NRF_CC3XX_PLATFORM=y
CONFIG_CC3XX_BACKEND=n
```
so what is the reason that the code which used to be guarded by `if (CONFIG_CC3XX_BACKEND)` is now being included when `CONFIG_CC3XX_BACKEND=n` ?

tejlmand (changes requested):
Outstanding changes requested.

crypto/CMakeLists.txt
Vge0rge: You are correct, the previous version could have linked more libraries than needed when CONFIG_CC3XX_BACKEND=n. With the current solution we only linking these libraries only when needed.

frkv (approved):
LGTM

The addition of CONFIG_NRF_CC3XX_MBEDCRYPTO with a help-text stating that this should supposed to be used without nrf_security resolves the nesting issue

crypto/Kconfig
```diff
@@ -23,6 +23,13 @@ if NRF_CC310_BL
 	default n # Only the no-interrupts version currently verified to work with Zephyr.
 endif
 
+menuconfig NRF_CC3XX_MBEDCRYPTO
+	bool "nrf_cc3xx_mbedcrypto SW crypto library for nRF devices with CryptoCell CC3xx."
```
tejlmand, line 27: based on:
> -This will be usefull for the use cases where the nordic security
 backend is not enabled but the mbedcrypto library needs to be linked

this sounds like that other part of the build system needs `NRF_CC3XX_MBEDCRYPTO` but without using nrf security should have the responsibility to select this setting.
Not the end-user, so I suggest to remove the prompt.
```suggestion
	bool
```

crypto/Kconfig
```diff
@@ -23,6 +23,13 @@ if NRF_CC310_BL
 	default n # Only the no-interrupts version currently verified to work with Zephyr.
 endif
 
+menuconfig NRF_CC3XX_MBEDCRYPTO
```
tejlmand, line 26:
```suggestion
config NRF_CC3XX_MBEDCRYPTO
```

crypto/Kconfig
```diff
@@ -23,6 +23,13 @@ if NRF_CC310_BL
...
+	bool "nrf_cc3xx_mbedcrypto SW crypto library for nRF devices with CryptoCell CC3xx."
+	depends on ENTROPY_CC3XX
+	help
+		It is not recommended to use the nrf_cc3xx_mbedcrypto without using
+		nrf_security as it is lacking some mbed TLS APIs
```
tejlmand, line 31: If this is promptless, then probably adjust the help text a bit to specify when / why you would enable this.

Something like:
```suggestion
	help
	  Use nrf_cc3xx_mbedcrypto library. This is automatically ensured when using nrf security
	  with the cc3xx backend.
	  Only select this setting without nrf security for special use cases, such as csrng support
	  without nrf security. 
```

tejlmand (changes requested):
I don't think `NRF_CC3XX_MBEDCRYPTO` should be user configurable.

crypto/Kconfig
frkv: Agree on this. This prevents direct usage (makes it only possible to set it by Kconfig)

tejlmand (approved):

PR #407, opened 2021-02-12
openthread: Update pre-build OpenThread libraries

lmaciejonczyk:
New libs based on tag: v1.5.0-rc1


edmont (changes requested):
Something seems wrong to me, OpenThread commit shouldn't have changed.

edmont (approved):
Well, actually it seems the tag was taken from here: https://github.com/nrfconnect/sdk-nrfxlib/blob/v1.5.0-rc1/openthread/cmake/extensions.cmake#L43, i.e., output from `git rev-parse HEAD` instead of the `git describe --long`, which was used before.

Not big deal then since both descriptions seem to point to the same commit in the module's repo, but I wonder why we end up having such a difference when building libraries from different machines.

lmaciejonczyk:
> Well, actually it seems the tag was taken from here: https://github.com/nrfconnect/sdk-nrfxlib/blob/v1.5.0-rc1/openthread/cmake/extensions.cmake#L43, i.e., output from `git rev-parse HEAD` instead of the `git describe --long`, which was used before.
> 
> Not big deal then since both descriptions seem to point to the same commit in the module's repo, but I wonder why we end up having such a difference when building libraries from different machines.

I have never set vanilla openthread as remote repository for zephyr/OpenThread. This is why I haven't got this tag on my machine. I suspect you did it contrary during last upmerge vanila OT -> zephyr/OT.

MarekPorwisz (approved):

tejlmand:
why do we constantly need to update those libs ?
It's 28Mb of binary files into the repo each time:
image

MarekPorwisz:
The libraries need to be updated when there is a change in openthread repo or for some nrf security changes. Ecosystems base their software on those libraries and would like us to update them even more often.
Is there any other place where those libs could be placed and downloaded automatically with west update? Maybe it would be possible for west to do only a shallow clone of this repo. Wasn't the intent of nrfxlib to keep the binary files?

PR #411, opened 2021-02-15
nrf_modem: doc: PDN doc update

evenl:
* Adding some missing information and re-phrasing some sentences
  to be more clear.
* Update last sample code to be less pseudo code.

### DIFF
--- a/nrf_modem/doc/pdn_extension.rst
+++ b/nrf_modem/doc/pdn_extension.rst
@@ -69,7 +69,7 @@ See `Creating a PDN socket`_.
 
 .. note::
    If the PDN that the socket is being bound to has been deactivated by the network, setting the NRF_SO_BINDTODEVICE option returns an error stating that the APN parameter is not valid.
-   To fix this error, you must create another PDN socket that is associated with the APN, and bind the data socket to it.
+   To fix this error, you must re-create the PDN socket that is associated with the APN, and bind the data socket to it.
 
 Configuring the type of transport
 *********************************
@@ -148,7 +148,7 @@ The example code shows how to specify an APN to use for a DNS query.
 In this example, the DNS hints are extended with an extra linked-list hint that specifies the APN to use for the query.
 
 Like for data sockets, the APN name must be set up and activated before the DNS query.
-Otherwise, the query will fail.
+Otherwise, the query will fall back on the default PDN (if it exists).
 
 Waiting for an IPv6 connection
 ******************************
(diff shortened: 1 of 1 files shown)


nrf_modem/doc/pdn_extension.rst
```diff
@@ -148,7 +148,7 @@ The example code shows how to specify an APN to use for a DNS query.
 In this example, the DNS hints are extended with an extra linked-list hint that specifies the APN to use for the query.
 
 Like for data sockets, the APN name must be set up and activated before the DNS query.
-Otherwise, the query will fail.
+Otherwise, the query will fallback on the default PDN if it exist.
```
evenl, line 151: A bit unsure about this one, this is what's happens today. Is this the correct behavior or is the old documentation describing the correct behavior and we should update the implementation?

stig-bjorlykke: I would like `getaddrinfo()` to fail with `ENOTCONN` if a PDN socket does not exist for this APN, and `ENETDOWN` if the PDN socket exists but is not active.

ru-fu: No idea about this question, but a language fix in case the changed version is what should be used. ;)
```suggestion
Otherwise, the query will fall back on the default PDN (if it exists).
```

nrf_modem/doc/pdn_extension.rst
```diff
@@ -157,21 +157,43 @@ Waiting for an IPv6 connection
 
    at_fd = nrf_socket(NRF_AF_LTE, 0, NRF_NPROTO_AT);
    nrf_send(at_fd, "AT+CGEREP=1", strlen("AT+CGEREP=1"));
-   # -> AT reponse "OK"
+   nrf_recv(at_fd, buf, sizeof(buf), 0);
+
+   if (strcmp("OK", buf) == 0) {
+        /* If the command did not return OK we failed to subscribe on the
```
ru-fu, line 163:
```suggestion
        /* If the command did not return OK, we failed to subscribe to the
```

nrf_modem/doc/pdn_extension.rst
```diff
@@ -157,21 +157,43 @@ Waiting for an IPv6 connection
...
+
    nrf_close(at_fd);
 
-When requesting an IPv6 connection (which requires PDN type IPV6 or IPV4V6), the IPv6 address might not be correct even if the request is successful.
+When requesting an IPv6 connection (which requires PDN type IPV6 or IPV4V6), the IPv6 address might not be global even if the request is successful.
 The reason for this is that many networks use SLAAC, where the prefix of the IPv6 address is set after the PDN connection is created.
-Therefore, you might need to wait with using an IPv6 socket until the prefix is properly set.
-If the address cannot be obtained, a ``+CGEV: IPV6 FAIL <cid>`` error is returned.
+Therefore, you might need to wait before using an IPv6 socket until the prefix is properly set.
+If the address is obtained, a ``+CGEV: IPV6 <cid>`` notification is received from the AT socket, if not a ``+CGEV: IPV6 FAIL <cid>`` notification is received.
```
ru-fu, line 192:
```suggestion
If the address can be obtained, the AT socket sends a ``+CGEV: IPV6 <cid>`` notification.
Otherwise, it sends a ``+CGEV: IPV6 FAIL <cid>`` notification.
```

nrf_modem/doc/pdn_extension.rst
```diff
@@ -157,21 +157,43 @@ Waiting for an IPv6 connection
...
 
 The example code shows how to set up an AT socket that logs "CGEV" entries.
 It signals when IPv6 is available for use by an IPv6 socket, either while connecting to a PDN or afterwards.
 
-Note that the example code does not cover the default PDN (CID 0), because this PDN is always available.
-However, the algorithm for checking that IPv6 is available on the default PDN is the same as if it was manually set up with an APN.
+.. note::
+        This example code does not cover the default PDN (CID 0), because this PDN is usually available in most networks.
+        However, the algorithm for checking that IPv6 is available on the default PDN is the same as if it was manually set up with an APN.
```
ru-fu, line 199:
```suggestion
.. note::
   This example code does not cover the default PDN (CID 0), because this PDN is usually available in most networks.
   However, the algorithm for checking that IPv6 is available on the default PDN is the same as if it was manually set up with an APN.
```

ru-fu (changes requested):

evenl:
Sorry for the slow response @ru-fu, I have merged your proposals now.

ru-fu (approved):

PR #428, opened 2021-03-11
nrf_802154: integrate nrf_802154 component into nrfxlib

e-rk:
Add config option to exclude radio driver in nrfxlib from the
build process.

Signed-off-by: Rafał Kuźnia <rafal.kuznia@nordicsemi.no>


CODEOWNERS
```diff
@@ -28,7 +28,7 @@ doc/*                                     @b-gent
 /crypto/                                  @frkv @tejlmand
 /mpsl/                                    @joerchan @hurricaneFromMoscow
 /nfc/                                     @anangl @grochu
-/nrf_802154_sl/                           @hubertmis
+/nrf_802154/                              @hubertmis
```
hubertmis, line 31: Not sure if I should be still owner of this. We should discuss it

nrf_802154/LICENSE
```diff
@@ -0,0 +1,28 @@
+Copyright (c) 2017 - 2021, Nordic Semiconductor ASA
```
hubertmis, line 1: I thought we were to use Nordic 5-clause license in NCS and re-licensing to BSD 3-clause in Zephyr. Is using BSD here intentional?

nrf_802154/serialization/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+#
+# Copyright (c) 2021 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+# Platform needs paths to serialization headers
+target_include_directories(nrf-802154-platform
+  PRIVATE
+    spinel_serialization/zephyr/include
```
hubertmis, line 10: What's the purpose of `spinel_serialization/zephyr` part in each path?
It suggests this module depends on Zephyr, but I hope it does not.

nrf_802154/serialization/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+#
+# Copyright (c) 2021 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+# Platform needs paths to serialization headers
+target_include_directories(nrf-802154-platform
```
hubertmis, line 8: All public and pvirate header files seem to be mixed in two directories now. I can't tell which public headers should be included in host build, which in remote build, and which describe platform API.

I remember in one of private repositories we had it clearly separated. Could we create such separation here?

nrf_802154/sl/sl_opensource/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+target_include_directories(nrf-802154-sl
+  PUBLIC
+    ../include
+    .
+)
+
+target_sources(nrf-802154-sl
+  PRIVATE
+    nrf_802154_sl_ant_div.c
```
hubertmis, line 19: Could we add `src` and `include` subdirectories here for consistency with other libraries?

nrf_802154/zephyr/Kconfig.nrfxlib
```diff
@@ -1,9 +1,18 @@
 #
```
hubertmis, line 1: I thought that this file depends on platform (Zephyr or NCS) and shall be moved to sdk-nrf repository. Do you keep it here intentionally?

nrf_802154_sl/CHANGELOG.rst
```diff
@@ -1,55 +0,0 @@
-.. _nrf_802154_sl_changelog:
```
hubertmis, line 1: Is it intentional to remove documentation files? I think they are useful here

nrf_802154_sl/README.rst
```diff
@@ -1,51 +0,0 @@
-.. _nrf_802154_sl:
```
hubertmis, line 1: Another documentation file removed

hubertmis (changes requested):
This PR looks very close to what we discussed. Good work!

However, I've spotted some deviations from the decisions. I would like to make sure these are intentional before we merge

nrf_802154_sl/CHANGELOG.rst
e-rk: I wanted to bring those files up in subsequent PRs. If you think it is better to do it right away, then it can be done.

nrf_802154_sl/README.rst
e-rk: I wanted to bring those files up in subsequent PRs. If you think it is better to do it right away, then it can be done.

nrf_802154_sl/CHANGELOG.rst
hubertmis: Let's discuss with tech writers if it is OK to remove these files temporarily. I have no strong opinion

hubertmis: @FrancescoSer, what do you think?

nrf_802154/zephyr/Kconfig.nrfxlib
e-rk: For the time being I have put this here, because `Kconfig.nrfxlib` at nrfxlib root needs it. I could move the `nrfxlib/Kconfig.nrfxlib` to `sdk-nrf/nrfxlib/Kconfig` right away, but I wasn't sure if VV wanted this change to happen right away.
@tejlmand @carlescufi Do you think I should move the `nrfxlib/Kconfig.nrfxlib` to sdk-nrf in the scope of this change?

nrf_802154/sl/sl_opensource/CMakeLists.txt
e-rk: Okay, it can be done.

nrf_802154/serialization/CMakeLists.txt
e-rk: I think it can be removed.

nrf_802154/LICENSE
e-rk: I have put BSD 3-clause at the `nrf_802154` root and Nordic 5-clause at `nrf_802154/sl/sl/lib`. A similar approach was taken here: https://github.com/NordicSemiconductor/nRF-IEEE-802.15.4-radio-driver

hubertmis: I thought we planned to change it. But it's a minor. We can discuss it later

nrf_802154_sl/CHANGELOG.rst
FrancescoSer: Hi! I would leave them in place, for now, especially because these are already in RST (unless there is any specific technical reason for not doing so). The remaining doc pages in .dox will be ported also to RST in the near future.

e-rk: I can put those files back. Do you have any suggestions where the documentation should be located in new layout? Should there be a dedicated `doc` directory where RST files for driver/sl/serialization are located?

e-rk: I have added the `CHANGELOG` and `README`. Those still describe the SL though, but both can be refactored later to reflect the new reality.

FrancescoSer: If the SL folder will be contained inside the RD folder, I would structure the doc folders like the ones for the crypto libraries, with a main readme in the root folder, the changelog in the lib folder, and most of the descriptive pages within a doc folder.

By the way, it can stay like it is, for now, as it wouldn't make sense to make this change till the rest of the RD documentation gets converted to RST.

FrancescoSer (approved):

e-rk:
@FrancescoSer For the time being the directory root has an initial readme and changelog for nrf_802154 component, and SL readme/changelog were moved to `nrf_802154/sl/sl` directory.

hubertmis:
@e-rk, you will need to decide with your team if you prefer single CHANGELOG for all libraries, or three separated CHANGELOGs. I don't think it's good to maintain four CHANGELOGs (one common to all libraries, and one for each library).

My preference would be to maintain single CHANGELOG for all.

hubertmis (approved):

nrf_802154_sl/CHANGELOG.rst
FrancescoSer: I deleted my previous message. After the short catch up with Rafal yesterday I believe that the best course of action would be to remove all .rst files from here (including the readme, changelog and SL related ones).

hubertmis: @FrancescoSer , what's the intended documentation structure for `nrf_802154` directory?
I assume it's:

```
nrf_802154/README.rst
nrf_802154/CHANGELOG.rst
nrf_802154/docs/ (documentation pages)
```

Do we want more detailed files like
```
nrf_802154/driver/README.rst
nrf_802154/sl/README.rst
nrf_802154/serialization/README.rst
nrf_802154/driver/CHANGELOG.rst
```
?

FrancescoSer: The current one is:

nrf_802154/README.rst
nrf_802154/ (documentation pages)
nrf_802154/docs/CHANGELOG.rst

Mostly following the nrfmodem structure, so the SL docs are inside the docs folder.

Having a page about the serialization library is actually something I was thinking about, as there isn't any at the moment.
I think it is necessary to have one.

hubertmis: Looks good. Thanks for clarificaiton

nrf_802154/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+#
+# Configurable build parameters
+#
+set(NRF52_SERIES  OFF CACHE BOOL   "Set to TRUE if target is an nRF52 series SoC.")
+set(NRF53_SERIES  OFF CACHE BOOL   "Set to TRUE if target is an nRF53 series SoC.")
+set(RADIO_DRIVER  OFF CACHE BOOL   "Set to TRUE if target requires the radio driver.")
+set(SERIALIZATION OFF CACHE BOOL   "Set to TRUE if target requires serialization module.")
+set(SER_HOST      OFF CACHE BOOL   "Set to TRUE if target is serialization host. Otherwise it is assumed that the target is serialization remote.")
+set(SL_OPENSOURCE OFF CACHE BOOL   "Set to TRUE to link with open source SL library")
```
tejlmand, line 15: Are there a specific reason those must be present in the CMake cache ?

Cache variables are variables that should be persistent across multiple build invocation, and usually user controllable.
(In Zephyr, Kconfig is used for the user controllable variables).

To me, it seems `NRF52_SERIES` and `NRF53_SERIES` are not to be controlled directly by a user, but a setting derived from other knowledge.

Not sure on the other settings.

Using a cache variable like this also means any existing non-cache variable of same name are hidden.

`RADIO_DRIVER`, `SERIALIZATION`, and `SER_HOST` looks very generic and thus in risk of colliding with identical named symbols if used outside Zephyr context.

nrf_802154/driver/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
+#
+# Copyright (c) 2021 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+# Platform needs paths to driver headers.
+target_include_directories(nrf-802154-platform
```
tejlmand, line 8: as mentioned here: https://github.com/nrfconnect/sdk-nrf/pull/4146#discussion_r600302788 `nrf-802154-platform` is a lib defined in Zephyr glue code.
How is a user that tries to integrate directly to `nrfxlib/nrf_802154` aware that this library must be defined ?

Should there be one of following options:
- Check for existence of `nrf-802154-platform` before setting includes
- Create a `nrf-802154-platform` in nrfxlib if not defined externally
- Throw a CMake error if lib is not available

nrf_802154/driver/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+      NRF_802154_IFS_ENABLED=1
+  )
+endif()
+
+target_link_libraries(nrf-802154-driver
+  PUBLIC
+    -Wl,--start-group
+    nrf-802154-platform
+    nrf-802154-sl
+    -Wl,--end-group
```
tejlmand, line 100: this will not work as you expect, due to symbol de-duplication.

In general, doing:
```
# In this file
target_link_libraries(nrf-802154-driver PUBLIC nrf-802154-platform nrf-802154-sl)

# In CMake file creating nrf-802154-platform
target_link_libraries(nrf-802154-platform PUBLIC nrf-802154-sl)

# In CMake file creating nrf-802154-sl 
target_link_libraries(nrf-802154-sl PUBLIC nrf-802154-platform)
```

will tell CMake about the circular dependency, and CMake should be able to handle this correctly.

tejlmand (changes requested):
Initial review, belongs together with comments here: https://github.com/nrfconnect/sdk-nrf/pull/4146#pullrequestreview-619459815

nrf_802154/driver/CMakeLists.txt
tejlmand: also note, `nrf-802154-platform` is a Zephyr defined library, so may not be available.

nrf_802154/driver/CMakeLists.txt
tejlmand: actually, what's in principle wrong with this code is the fact that you add `include` and `src` to the platfrom lib.

When using CMake, the principle is to link to the lib from where you need to inherit includes, that is:
```
add_library(nrf-802154-driver)
target_include_directories(nrf-802154-driver
  PUBLIC include # Any depending (linking) on driver need to inherit this, and driver itself needs it.
...
)
```
and then elsewhere do:
```
# Platform needs to link to driver, and inherit public includes.
target_include_directories(nrf-802154-platform PRIVATE nrf-802154-driver)
```

e-rk: The expectation is that the user will always provide the `nrf-802154-platform` target before doing `add_subdirectory(nrfxlib/nrf_802154)`. This fact will be documented in an integration guide in the future.

You are right that it is confusing that some includes are added in a different place than where the target was first created. My initial intention was to prevent the user from supplying the path manually or linking the entire driver target to the platform. I didn't know it was possible to use `target_include_directories` with targets so that's a good tip.

tejlmand: >  I didn't know it was possible to use `target_include_directories` with targets so that's a good tip.

sorry, that was a typo, should have been `target_link_libraries()`

tejlmand: additional thought while reviewing https://github.com/zephyrproject-rtos/hal_nordic/pull/93.

May `nrf-802154` CMake code should always create the `nrf-802154-platform`.
Then the in integration guide, you may simply write:
> nrfxlib defines a `nrf-802154-platform` library for integrating the `nrf-802154-driver` to the platform.
> It is the responsibility of the project to add required platform sources to this library by calling:
> ```
> target_sources(nrf-802154-platform PRIVATE <platform-sources>)
> ```
>
> It is also possible to add additional compile options or include paths to `nrf-802154-platform` by calling relevant CMake functions.

That way, nrf-802154-driver is responsible for creating the lib, but project is responsible for adding the correct sources implementing the API.

nrf_802154/CMakeLists.txt
e-rk: Taking https://github.com/zephyrproject-rtos/hal_nordic/pull/93#discussion_r601213807 into context:

My expectation is that this file will never be the root of the build. My primary intention of writing those variables here was to document their presence. I think those lines can be completely removed and required variables may be described in a documentation.

nrf_802154/driver/CMakeLists.txt
e-rk: > May nrf-802154 CMake code should always create the nrf-802154-platform.
> Then the in integration guide, you may simply write:

I am not totally sure how that would work. If `nrf-802154-platform` were to be defined in `nrf-802154` cmake code, then adding platform sources in https://github.com/zephyrproject-rtos/zephyr/blob/51efef4ff24b04a523bb7da9a11b6484e1248133/modules/hal_nordic/CMakeLists.txt would be impossible, because https://github.com/zephyrproject-rtos/zephyr/blob/51efef4ff24b04a523bb7da9a11b6484e1248133/modules/hal_nordic/CMakeLists.txt#L95 may or may not be executed depending on `CONFIG_NRF_802154_SOURCE_HAL_NORDIC` choice. The `target_sources` doesn't accept not-yet-existent target.

nrf_802154/driver/CMakeLists.txt
e-rk: I have fixed the circular dependency as you suggested.

nrf_802154/CMakeLists.txt
e-rk: I have removed the cache variables.

nrf_802154/driver/CMakeLists.txt
e-rk: I have removed the statements that added include directories to `nrf-802154-platform`.

tejlmand: > because https://github.com/zephyrproject-rtos/zephyr/blob/51efef4ff24b04a523bb7da9a11b6484e1248133/modules/hal_nordic/CMakeLists.txt#L95 may or may not be executed depending on `CONFIG_NRF_802154_SOURCE_HAL_NORDIC` choice. The `target_sources` doesn't accept not-yet-existent target.

Correct. So you still need the platform lib, even if not using the 802154 hal nordic sources, I assume.

e-rk:
Rebased to unblock CI work.

tejlmand (approved):

e-rk:
@carlescufi 
The compliance check is failing due to coding style of the radio driver. Is it possible to give an exception to the coding style this time and decide the proper course of action later?

czeslawmakarski (approved):

tejlmand (changes requested):
Blocked due to this: https://github.com/nrfconnect/sdk-nrf/pull/4146#pullrequestreview-624653922

tejlmand (approved):
looks great

czeslawmakarski:
> @carlescufi
> The compliance check is failing due to coding style of the radio driver. Is it possible to give an exception to the coding style this time and decide the proper course of action later?
> 
> Other than that, the PR is ready for merge.

If I'm not mistaken the coding style of the radio driver is the Nordic one, not the Zephyr/NCS one. sdk-nrfxlib repo from the conceptual point of view does not depend on Zephyr/NCS. I think this is a false positive.

e-rk:
@czeslawmakarski Yes, I added the `nrf_802154` to checkpatch excludes. The compliance is not red any longer.

FrancescoSer (approved):

carlescufi (approved):

PR #431, opened 2021-03-15
nrf_rpc: Change name of the nrf_ernno.h to nrf_rpc_errno.h

KAGA164:
This is needed because the MPSL header has exactly the same name.
When application uses the MPLS and the nrf_rpc then on of this
header is included which may cause the build errors.


joerchan (approved):

joerchan:
@KAGA164 We need to get a common one.

I'm guessing this works because all defines have the same value?

KAGA164:
> @KAGA164 We need to get a common one.
> 
> I'm guessing this works because all defines have the same value?

Yes, This is a reason why this works

doki-nordic (approved):

grochu (approved):
Looks good

PR #461, opened 2021-04-27
nrf_security: kconfig: support promptless configurations

tejlmand:
When using 'select NORDIC_SECURITY_BACKEND' is modules, such as
OpenThread or Pelion, then config settings get stuck and user cannot
select another security configuration group.

The NORDIC_SECURITY_PROMPTLESS allows modules like Pelion or OpenThread
to disable the prompt when using pre-defined configuration groups.

### DIFF
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -1,8 +1,15 @@
 menu "Nordic Security"
 
+config NORDIC_SECURITY_PROMPTLESS
+	bool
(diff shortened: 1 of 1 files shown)


pdunaj (approved):

frkv (approved):
Looks VERY good to me!
Neat trick!

edmont (approved):

PR #462, opened 2021-04-30
doc: thread: update certification information

ru-fu:
Update the information about OpenThread prebuild libraries and
replace the list of certified libraries with links to the
compatibility matrices.

Signed-off-by: Ruth Fuchss <ruth.fuchss@nordicsemi.no>

### DIFF
--- a/doc/links.txt
+++ b/doc/links.txt
@@ -12,6 +12,10 @@
 
 .. _`Infocenter`: https://infocenter.nordicsemi.com/index.jsp
 
+.. _`Thread CIDs for nRF5340`: https://infocenter.nordicsemi.com/topic/comp_matrix_nrf5340/COMP/nrf5340/nrf5340_thread_cids.html
+.. _`Thread CIDs for nRF52840`: https://infocenter.nordicsemi.com/topic/comp_matrix_nrf52840/COMP/nrf52840/nrf52840_thread_cids.html
+.. _`Thread CIDs for nRF52833`: https://infocenter.nordicsemi.com/topic/comp_matrix_nrf52833/COMP/nrf52833/nrf52833_thread_cids.html
+
 .. ### Links to Nordic pages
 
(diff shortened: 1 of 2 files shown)


lmaciejonczyk (approved):

edmont (commented):
Looks good, however now we are eliding information like the OpenThread commit or the compiler version. Would those be relevant for the customers?

ru-fu:
> 
> 
> Looks good, however now we are eliding information like the OpenThread commit or the compiler version. Would those be relevant for the customers?

Would it make sense (and be technically feasible 😉) to add this information to the openthread_lib_configuration.txt files?

MarekPorwisz:
> Looks good, however now we are eliding information like the OpenThread commit or the compiler version. Would those be relevant for the customers?

Good question. But I think this is not something that a customer can not dig out himself. Maybe if this is something needed it needs to be added to the infocenter, not here.

MarekPorwisz (approved):

ru-fu:
> 
> 
> > Looks good, however now we are eliding information like the OpenThread commit or the compiler version. Would those be relevant for the customers?
> 
> Would it make sense (and be technically feasible 😉) to add this information to the openthread_lib_configuration.txt files?

Actually, the OpenThread commit is already in there.

edmont:
> Actually, the OpenThread commit is already in there.

But I believe that's the hash of the commit from the Zephyr's OpenThread fork.

ru-fu:
Would users need this information for all pre-built libraries, or only for those that are certified?
If all, I think the information should be added to the configuration text files. If we put it in the docs, it'll constantly be outdated - but we could maybe look into including the information from the text files in the docs automatically in the future.
If the information is needed for the certified libraries only, it should go into the comp matrix.

edmont:
Well, probably that's not vital information, but nice to have. Regarding the OpenThread commit, I don't think it's automatic to retrieve the one from the official repository, but maybe it's enough with having the one from the Zephyr's fork. Interested users could perform the research their shelves as Marek mentions. About the compiler version, it should be fine with adding it to the configuration text files.

edmont (approved):

PR #467, opened 2021-05-19
crypto: Adding option for no mutex when MULTITHREADING is off

frkv:
-Added additional source file that does not initialize RTOS mutexes
 in the CryptoCell runtime library
-Added build-asserts to protect against misuse/misconfiguration

ref: NCSDK-9650


nordic-krch (approved):

ioannisg (approved):

nvlsianpu:
introduced within #469

PR #468, opened 2021-05-20
nrf_802154: Serialize nrf_802154_time_get() for nRF53

jciupis:
This commit adds serialization of nrf_802154_time_get() function
to fix build for nRF5340, where it must be serialized in order to
be available.

Signed-off-by: Jedrzej Ciupis <jedrzej.ciupis@nordicsemi.no>

### DIFF
--- a/nrf_802154/serialization/src/nrf_802154_spinel_app.c
+++ b/nrf_802154/serialization/src/nrf_802154_spinel_app.c
@@ -279,6 +279,49 @@ bail:
     return error;
 }
 
+/**
+ * @brief Wait with timeout for time property to be received.
+ *
+ * @param[in]  timeout  Timeout in us.
+ * @param[out] p_time   Pointer to the time variable which needs to be populated.
+ *
+ * @returns  zero on success or negative error value on failure.
+ *
+ */
+static nrf_802154_ser_err_t time_await(uint32_t   timeout,
+                                       uint32_t * p_time)
+{
+    nrf_802154_ser_err_t              res;
+    nrf_802154_spinel_notify_buff_t * p_notify_data = NULL;
+
+    SERIALIZATION_ERROR_INIT(error);
+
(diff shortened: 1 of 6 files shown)


lmaciejonczyk (approved):

nrf_802154/serialization/src/nrf_802154_spinel_app.c
```diff
@@ -279,6 +279,49 @@ static nrf_802154_ser_err_t capabilities_await(uint32_t                    timeo
...
+                           bail);
+
+    res = nrf_802154_spinel_decode_prop_nrf_802154_time_get_ret(p_notify_data->data,
+                                                                p_notify_data->data_len,
+                                                                p_time);
+
+    SERIALIZATION_ERROR_CHECK(res, error, bail);
+
+    NRF_802154_SPINEL_LOG_BANNER_RESPONSE();
+    NRF_802154_SPINEL_LOG_VAR_NAMED("%ul", *p_time, "Time");
```
Damian-Nordic, line 314: Why 'l' at the end of the format string?

czeslawmakarski (approved):

Damian-Nordic (approved):

nrf_802154/serialization/src/nrf_802154_spinel_app.c
czeslawmakarski: Isn't it for `long`? `p_time` is the pointer to the `uint32_t`, which corresponds to the `long` one.

Damian-Nordic: So I guess it should be `%lu` in such a case? With the existing code, the `l` will be printed after the value.

lmaciejonczyk: Maybe using PRIu32 for print formatting uint32_t will be better solution?

nvlsianpu:
introduced within #469

PR #475, opened 2021-05-31
Bluetooth: Mesh: Switch default BLE controller for mesh to SDC

PavelVPV:
Run Bluetooth mesh in NCS with SoftDevice Controller by default.

Signed-off-by: Pavel Vasilyev <pavel.vasilyev@nordicsemi.no>

### DIFF
--- a/softdevice_controller/Kconfig
+++ b/softdevice_controller/Kconfig
@@ -32,7 +32,6 @@ config BT_CTLR_LLPM
 
 choice BT_LL_CHOICE
 	prompt "Bluetooth Link Layer Selection"
-	default BT_LL_SW_SPLIT if BT_MESH
 	default BT_LL_SOFTDEVICE
 
 config BT_LL_SOFTDEVICE


softdevice_controller/Kconfig
```diff
@@ -31,8 +31,8 @@ config BT_CTLR_LLPM
 
 choice BT_LL_CHOICE
 	prompt "Bluetooth Link Layer Selection"
-	default BT_LL_SW_SPLIT if BT_MESH
-	default BT_LL_SOFTDEVICE
+	default BT_LL_SOFTDEVICE if BT_MESH
+	default BT_LL_SW_SPLIT
```
joerchan, line 35: You would be changing the default for non-mesh. So if you are removing the mesh default to SPLIT it should look like this:
```suggestion
	default BT_LL_SOFTDEVICE
```

joerchan (changes requested):
Wrong default for non-mesh.

joerchan (approved):

Balaklaka (approved):

trond-snekvik (approved):

rugeGerritsen (approved):

PR #482, opened 2021-06-09
478 480 481 and 483

ioannisg:
#478 
#480 
#481 

and #483

PR #483, opened 2021-06-09
nrf_security: introducing real_mbedcrypto and real_mbedx509 targets

tejlmand:
The pelion-dm defines mbedcrypto and mbedx509 as interface libraries.

This results in duplicate library definitions as nRF Security creates
the same libraries.

This commit introduces an existence check for mbedcrypto library and in
case such library is already existing, the target library name in nRF
Security will be real_mbedcrypto and real_mbedx509 instead.

This commit can be reverted when the root cause is fixed in pelion-dm
repo.


pdunaj (approved):

frkv (approved):
LGTM

tejlmand:
closed, part of #482

PR #493, opened 2021-06-15
nrf_security: cmake: refactor library re-archiving and fix archive globbing expression

tejlmand:
nrf_security: use cmake script for TFM header copy on windows

Fixes: NCSDK-10008

CMake / Ninja generator will create a bat file for command invocation
when the command line execution exceeds a certain length.

Unfortunately SES-NE invokes said bat file wrongly, so instead we create
a CMake script to be invoked in windows for the header copy operation.

---------

nrf_security: cmake: refactor library re-archiving

Fixes: NCSDK-10008

SES-NE has a limitation on how long a POST_BUILD command from the
build.ninja it can handle.

When that length is exceeded then post build commands are truncated in
the middle and thus causing unexpected behavior.

This commit creates a <backend>_extract build target which can then be
set as a depedency to the static library. As the extract target is a
dependency it will be executed prior to the build of the static library.
This ensures that the post build step of adding objects to the static
library can now be safely done as the post build step is now within the
limits of SES-NE.

---------

 nrf_security: cmake: using archive script when archiving a folder

Fixes: NCSDK-10008

When archiving extracted objects from imported nRF security libraries
then a globbing expression in the form of: `ar q <library> <folder>/*`
is used.

Unfortunately, SES-NE treat the `*` as a filename and not a globbing
expression which results in `ar` to fail as there is no file named `*`.

This is handled by introducing archive_folder.cmake script which takes
the folder as an argument and then globs for all files in that folder.

The list of files is then passed directly to the `ar` command including
the library, and thus no globbing is needed directly on `ar`.

### DIFF
--- a/nrf_security/cmake/archive_folder.cmake
+++ b/nrf_security/cmake/archive_folder.cmake
@@ -1,0 +1,17 @@
+#
+# Copyright (c) 2021 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+#
+# Script used to add object files to archive.
+#
+# The script requires the following parameters to be set
+# CMAKE_AR : Archiver to use for extracting
+# LIBRARY  : Library to which objects from FOLDER should be added.
+# FOLDER   : Folder containing objects to add to archive.
+#
+file(GLOB files ${FOLDER}/*)
+
+execute_process(COMMAND ${CMAKE_AR} q ${LIBRARY} ${files})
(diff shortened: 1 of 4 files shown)


tejlmand:
This fixes one part of NCS-10008.

This means things are now working with SES-NE on MacOS and Linux, and on Windows after running `ninja tfm_mbedtls_headers_copy` manually.

Still looking into the tfm_mbedtls_headers_copy` manually dependency issue on Windows.

frkv (approved):
LGTM and should fix the SES-issues (except the header-file copy, as stated)

tejlmand:
> Still looking into the `tfm_mbedtls_headers_copy` manually dependency issue on Windows.

Fixed

frkv (approved):
LGTM

andvib (approved):

tejlmand:
closed, included in #498

PR #521, opened 2021-07-19
doc: fix Doxygen issues

gmarull:
There are numerous Doxygen issues in the nrfxlib code that need to be
addressed:

- Incorrect usage of C preprocessor
- Unbalanced Doxygen groups
- Missing/wrong parameter documentation
- Missing/wrong references

This change set is a WIP, includes a few examples that have been fixed.
The issues can be seen when building the nrfxlib docset in the NCS repository, or alternatively:
https://gist.github.com/gmarull/8fb0a5a19c00ca45a1e10b69bc87f626


FrancescoSer:
Please also check JIRA KRKNWK-10910, as it appears to be related to the issues solved here.

nrf_802154/driver/include/nrf_802154.h
```diff
@@ -102,7 +102,7 @@ void nrf_802154_deinit(void);
  */
 extern void nrf_802154_custom_part_of_radio_init(void);
 
-#if !NRF_802154_INTERNAL_RADIO_IRQ_HANDLING
+#if !defined(NRF_802154_INTERNAL_RADIO_IRQ_HANDLING)
```
pkwiek, line 105: Using `defined` and `#ifdef` instead of `#if` changes behavior of the driver.

pkwiek (changes requested):

nrf_802154/driver/include/nrf_802154.h
gmarull: Doxygen doesn't seem to be able to process `#if !NRF_802154_INTERNAL_RADIO_IRQ_HANDLING`, not sure what is the best solution here.

PR #524, opened 2021-07-26
sm_ipt: Add Shared Memory Inter-Processor Transport

dchat-nordic:
SM IPT is an OS independent shared memory
communication library, which works on shared memory
and provides os-porting templates.

Signed-off-by: Dominik Chat <dominik.chat@nordicsemi.no>

Co-authored-by: Dominik Kilian <Dominik.Kilian@nordicsemi.no>


carlescufi (dismissed):
NACKing for now until I learn more about the purpose of this library

doki-nordic:
Main purpose of this library is a backend for nRF RPC. Additionally, it can be later used as a backend for IPC Service.

Main benefits of this library (compared to current RPMsg solution):
* Can be easily ported to platforms other than Zephyr. One of the nRF RPC goals is to support multiple OSes or even baremetal. Currently there is only Zephyr/RPMsg backed which is not simply portable.
* Better performance. @dchat-nordic did preliminary measurements and RPC calls are ~30% faster.  It is achieved by:
  * Simpler implementation (it has less CPU overhead).
  * It uses zero-copy approach to reduce memory transfers.
* Less memory consumption:
  * zero-copy approach requires no additional buffers,
  * No need for additional receiving thread and its stack. Specific nRF RPC thread is triggered directly from the interrupt.
* Arriving data can be processed in any order. This makes nRF RPC less blocking.

doki-nordic:
@dchat-nordic, please update commit message, to follow [commit guidelines](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/contribute/index.html#commit-guidelines):
* Change commit author name to match "Signed-off-by" in the commit message
* Put single empty line after commit title and before commit details.

sm_ipt/template/sm_ipt_log_tmpl.h
```diff
@@ -0,0 +1,105 @@
...
+ */
+#define SM_IPT_INF(...)
+
+/**
+ * @brief Macro for logging a message with the severity level DBG.
+ *
+ * @param ... printf-style format string, optionally followed by arguments
+ *            to be formatted and inserted in the resulting string.
+ */
+#define SM_IPT_DBG(...)
```
doki-nordic, line 59: We are not using logs different than `ERR`. Maybe it will be better to add some more logs, e.g.:
* `DBG` with block number (or location) and size for each allocation, deallocation, send, receive, e.t.c.
* `DBG` when there is not enough memory and we are waiting on semaphore for more free memory.
* `DBG` on each handshake step.
* `INF` with one line of information that module has been initialized and maybe some memory sizes information.

doki-nordic (changes requested):

doki-nordic:
This PR requires high level documentation.

hubertmis:
Have I understood correctly that the purpose of this library is to replace RPMsg (OpenAMP) to get better performance?

> Can be easily ported to platforms other than Zephyr

RPMsg is available in other platforms than Zephyr (I'm aware of Linux)

> RPC calls are ~30% faster

> Less memory consumption

> This makes nRF RPC less blocking

These are examples of better performance of this library compared to RPMsg.

I'm thinking if it is the best way to maintain one more extension to Zephyr in NCS, instead of improving RPMsg performance. With two IPC transports (RPMsg upstream and SMIPT downstream) we would need to double verification effort for components like HCI over IPC, 802.15.4 serialization. From IPC user perspective, I would prefer to have the same transport available upstream and downstream, with good enough performance.

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,420 @@
...
+
+static bool free_mask_unset(struct sm_ipt_ctx *ctx, mask_t mask)
+{
+	uint32_t old;
+	uint32_t mask_part;
+
+	mask_part = (uint32_t)mask;
+	old = sm_ipt_os_atomic_and(&ctx->free_mask[0], ~mask_part);
+	if ((old & mask_part) != mask_part) {
+		sm_ipt_os_atomic_or(&ctx->free_mask[0], old & mask_part);
```
KAGA164, line 61: Is this some kind of fallback mechanism ?

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,420 @@
...
+		return false;
+	}
+	if (NUM_BLOCKS > 32) {
+		mask_part = (uint32_t)((uint64_t)mask >> 32);
+		old = sm_ipt_os_atomic_and(&ctx->free_mask[1], ~mask_part);
+		if ((old & mask_part) != mask_part) {
+			sm_ipt_os_atomic_or(&ctx->free_mask[1], old & mask_part);
+			sm_ipt_os_atomic_or(&ctx->free_mask[0], (uint32_t)mask);
```
KAGA164, line 69: No needed ?

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,420 @@
...
+	uint32_t old;
+	uint32_t mask_part;
+
+	mask_part = (uint32_t)mask;
+	old = sm_ipt_os_atomic_and(&ctx->free_mask[0], ~mask_part);
+	if ((old & mask_part) != mask_part) {
+		sm_ipt_os_atomic_or(&ctx->free_mask[0], old & mask_part);
+		return false;
+	}
+	if (NUM_BLOCKS > 32) {
```
KAGA164, line 64: Maybe looping over `free_mask` would be more efficient than if block

KAGA164 (changes requested):
I finished first round of a review. I will take another look after you will respond to my comments

hubertmis:
> What is the point of keeping backend-enabled IPC service and nRF RPC if we allow only one backend.

The point is to allow application developers using Nordic to use any backend in their product and to implement their own backends. However, we don't need to verify each backend. The one we recommend is enough.

> * I don't know how it can be ported to baremetal.

Is there a use case to create a bare metal port for IPC?

doki-nordic:
> Is there a use case to create a bare metal port for IPC?

Yes, this is one of the requirements that we get for nRF RPC.

sm_ipt/sm_ipt.c
dchat-nordic: Yes, if operation atomic_and was not done as expected - restore changes.

hubertmis:
We've discussed IPC transports during a meeting. I have no more concerns about maintenance cost for networks protocols serialization, because OpenAMP is going to remain default and recommended transport.

sm_ipt/sm_ipt.c
dchat-nordic: It is needed in case of 64 blocks:
If clearing high bits fails all bits must be restored to previous value.

sm_ipt/Kconfig
```diff
@@ -0,0 +1,42 @@
...
+config SM_IPT_NUM_BLOCKS_64
+	bool "64 shared memory blocks per channel"
+	help
+	  Selects 64 shared memory blocks per single input or output channel.
+	  This will increase allocatable memory granularity, but it will have
+	  some impact on performance and code size.
+
+endchoice
+
+config SM_IPT_PRIMARY
```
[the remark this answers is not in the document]
doki-nordic: Probably not:
https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-2-inclusive-language

sm_ipt/template/sm_ipt_os_tmpl.h
```diff
@@ -0,0 +1,114 @@
...
+	some_type other_os_context_data;
+};
+
+/** @brief Assertion function.
+ */
+#define SM_IPT_ASSERT(_expr) __ASSERT_FUNCTION(_expr)
+
+/** @brief Full memory barier.
+ */
+#define NRF_RPC_OS_MEMORY_BARIER() __sync_synchronize()
```
doki-nordic, line 49: I see that I did a misspelling here. Should be NRF_RPC_OS_MEMORY_BA***RR***IER.

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,420 @@
...
+
+#define ALLOCABLE_MULTIPLY (WORD_SIZE * NUM_BLOCKS)
+#define QUEUE_INDEX_MASK (NUM_BLOCKS - 1)
+
+#define memory_corrupted_error() sm_ipt_os_fatal(); \
+	SM_IPT_ERR("Shared memory corrupted"); \
+	SM_IPT_ASSERT(0)
+
+
+static inline void free_mask_set(struct sm_ipt_ctx *ctx, mask_t mask)
```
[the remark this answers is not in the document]
hubertmis: Actually, it makes very little difference. If I remember correctly, GCC obeys `inline` clue only for -O1. I don't think we use -O1 for any build :)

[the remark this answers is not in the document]
doki-nordic: hmm, I had a different experience. I have to check it.

sm_ipt/sm_ipt.c
doki-nordic: ... and return false.

This is needed, because I avoided locks, so in some rare cases, other thread may change `free_mask` after this thread calculated the `mask` and before `atomic_and`. In such case whole operation in repeated in current thread.

sm_ipt/sm_ipt.c
doki-nordic: No. Loop is done one level higher, because loop added here may wait for a long time for bits to be available again. Higher level function will recalculate the mask and try again immediately.

sm_ipt/sm_ipt.c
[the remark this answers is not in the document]
hubertmis: It seems now it inlines automatically even for -O1. Play with https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlJvoCqAZ0wAFAB4gA5AAYppAFZdSrZrVAB9LclIj2yAnjqVMtdAGFUrAK4BbWiADMnUqfQAZPLUwA5OwCNMYicATlIAB1QhQiNaSxt7JxdI6MM6T28/W0DgxzC9TANYhgJmYgJ4uwdnXUx9NNoSsoIM3wCg0N1S8srEmqFulq827I68gEpdVGtiZA4pAHoFgGoAFQBPcMxl9eniZbQsZYQgzFJlkmXWVGZ0ZbVlzHFmW3D2ADoAUhkAQW%2B/gBMji8yBsR0%2BjnMA3w9HeCAh2H%2BSKBWCow2WPgEAFkNAAhdwAeXMAGkGMtHADkY5UeiGDiAJKiVYaAkMDRY7BYgkAJQAmniftz6dhuRBxkjfgRNphUcsvAQKRoCMtbMwhABrJUQ3H/AbEawGZZCWwaPDhAgaAziZafADsOt%2ByydKrVmuVVGImEwGlVGs%2BAFZcZT/QARbVI21hxwOv6/ABuqDwd2NpvNGiiGmYRFseGQ6eIEF9buWACpwj7Xeci0qXRrxdGJT8Blnc3LaKx0Qmk22O94NB6vRWNRoRAQIHqDcqU2aLVbS1aq66a0Xxb87THnUaTTP00JM9nc/mIJ8AQA2K0AWgRA%2B9RYDuJkAZDi7r4bXMhCeCoywgmJx%2BKJUkbUcbByQBcYbXtf5N03ac0wzLNUBzPMSGPM9L2vT1b1de9OCfc4IGseVT24JVxiLYDERAsD6xjb4QjtMN3xCOkNEZZlWXZTkeX5XFBWFUVaIjJjYybUpDGQZYuzuG8h01UdxwIfVDTg2cCGtEsF1rYsVwjDdnVU3d9yQw9UJPc91KvEDZLvQNH1DF91SE5ivx/P88UJEkyQRGjIP0mCt1TC0EIPFCC3MjDrKwuTcPwn8iPoEiyIohEfIpZzYwYyMJRYhkmRZNkOS5PkBSFEUxTfP5svfH5pJdLwKrXKDHQMpTJ0Cnc5znCEQz89dH2y7KGxap1q2VFLHF6x9htEzcvF7b0bKXBSIvUxyMs3JbhxW9C1u0jLGKkSZWGkf15AcWR5FQaRzCNPY5htIFOHkAhpDkcZJnVEBuBkd4QgBD8AA5HFtW1Ac4M9T39FRpG4c63qu6R5CEEAZFIV7LsmOBYCQNBXjwdgyAoCA8fCAmOhYYhbE4TgZDRtFWAIIIUYgfwEdIfwvDKdZpGe0g8dsUwCAJdsecu0gsFVdR2HZ/BPSKONMBR8WnkKawmd5%2BR5TqdmO38YhucsLB2aUvBbE1yYqFUYAhAANTwTAAHcCS2C6%2Bf4RgWBlng%2BDoAhhDESRxaUFxVHUEAtEtFQ8H8FHIEmVBzViZWLwJPmCiKYwIDcXoHBcNxWiyHJlBSGI6FzkuojL2hC/aYIXAzhomh6KwqmURvikGWvRnrrpmgrhuu%2BGIuOk4SYhHujgx5hqQztIC65FIa6pEFLFlmAZBJJp95OB/XtrHECCIFwQhLhPGplksfHCce5wIPMF6EY%2B0gvoBf13ltbh/VPQGAUB///Q/RkLwE6Ug4bz3ZsvZGqN0ZP1INjRAIBpgEHCOrcglBSbk17jKU%2BvcPZMDYFPXgHsA4SHZo7A24QLYzzngvRGUgCTq1QcqVA35V7r03ssbeu8ID70Pj%2BK%2BZMb7nzHo/TGn0nCOHeI4GRsi5FyJnuAuhS8ka6BgRjd6x1pAAnhuLKBsDxGkEVsQaIxhuBAA

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,420 @@
...
+		/* create bit mask with blocks that will be used */
+		mask = calc_mask(blocks, free_index);
+		/* update masks */
+		unset_success = free_mask_unset(ctx, mask);
+		/* there is a small probability that unset will be unsuccessful */
+		if (!unset_success) {
+			/* give semaphore, because free_mask_unset may cause other thread
+			 * waiting before it reverted the changes
+			 */
+			SM_IPT_DBG("Failed to allocate shared memory");
```
doki-nordic, line 154:
```suggestion
			SM_IPT_DBG("Not enough free shared memory. Waiting.");
```

doki-nordic: This line should be moved after `sem_taken = false`, because comment above does not apply to this line.

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,420 @@
...
+{
+	uint32_t *mem_start = (uint32_t *)buf - 1;
+	uint32_t offset = (uint8_t *)mem_start - ctx->out.allocable;
+	uint32_t block_index = offset / ctx->out.block_size;
+	uint32_t allocated_size;
+	uint32_t allocated_blocks;
+	uint32_t blocks = (len + ctx->out.block_size - 1) / ctx->out.block_size;
+	uint32_t total_len = len + WORD_SIZE;
+
+	SM_IPT_DBG("Trying to send message of: %d bytes, at address: 0x%x", len, (uint32_t) buf);
```
doki-nordic, line 237: Do not use space in type casting. It should be treated as unary operator.

doki-nordic (changes requested):

doki-nordic:
I wrote core part of this code, so please add `Co-authored-by:` into the commit message.

sm_ipt/sm_ipt.c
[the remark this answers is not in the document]
doki-nordic: I played a little bit whit the code and I was able to observe different output for inline and non-inline function for -O1. I achieved that by adding two more functions that calls `free_mask_set()`. I wasn't able to do that for -O2 and -Os.

On the other hand, I now see that atomic `or` on Cortext-M33 is not one instruction long, but five, so inlining it does not make a lot of sense. Let the compiler to decide.

I agree to delete `inline`.

sm_ipt/doc/api.rst
```diff
@@ -0,0 +1,17 @@
...
+API documentation
+#################
+
+.. contents::
+   :local:
+   :depth: 2
+
+.. _sm_ipt_api_documentation:
+
+SM IPT API documentation
```
peknis, line 12:
```suggestion
SM IPT
```
You said already API documentation above.

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
+.. _sm_ipt_usage:
+
+Usage and Implementation
```
peknis, line 3:
```suggestion
Usage and implementation
```
We don't use title case

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+   :alt: SM IPT architecture layers
+   :align: center
+
+   SM IPT architecture layers
+
+Usage
+=====
+
+.. note::
+   Following examples are purely conceptual and simplified to show library functions. They should be rewritten to suit particular need and use OS modules like FIFO.
```
peknis, line 22:
```suggestion
   Following examples are purely conceptual and simplified to demonstrate the library functions.
   They should be rewritten to suit particular need and use OS modules like FIFO.
```
Every sentence must start from a new line. Everywhere.

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+		/* Process data */
+		some_data_processing_function();
+	}
+
+	int initialize(void)
+	{
+		return sm_ipt_init(&sm_ipt_ctx, sm_ipt_receive_handler);
+	}
+
+The above code uses the ``sm_ipt_init`` function to initialize ``struct sm_ipt_ctx`` SM IPT context data structure and assign the receive handler.
```
peknis, line 45:
```suggestion
The above code uses the :c:func:`sm_ipt_init` function to initialize the :c:struct:`struct sm_ipt_ctx` SM IPT context data structure and to assign the receive handler.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+		sm_ipt_alloc_tx_buf(&sm_ipt_ctx, &buffer, len);
+
+		/* Use buffer as any other allocated buffer */
+		some_data_preparation_function(buffer);
+
+		return sm_ipt_send(&sm_ipt_ctx, buffer, len);
+	}
+
+The above code uses the ``sm_ipt_alloc_tx_buf`` function to allocate transmit buffer in shared memory. Buffer is prepared by ``some_data_preparation_function`` and sent to another core via ``sm_ipt_send`` function.
```
peknis, line 87:
```suggestion
The above code uses the :c:func:`sm_ipt_alloc_tx_buf` function to allocate the transmit buffer in the shared memory.
The buffer is prepared by the :c:func:`some_data_preparation_function` and sent to another core through the :c:func:`sm_ipt_send` function.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+			return sm_ipt_send(&sm_ipt_ctx, buffer, len);
+		}
+		else{
+			sm_ipt_free_tx_buf(&sm_ipt_ctx, buffer);
+
+			return ERROR_ABORT;
+		}
+	}
+
+The above code uses the ``sm_ipt_alloc_tx_buf`` function to allocate transmit buffer in shared memory. Buffer is send only if send_choice is true. If send_choice is false the buffer is free using ``sm_ipt_free_tx_buf`` and send operation is aborted.
```
peknis, line 116:
```suggestion
The above code uses the :c:func:`sm_ipt_alloc_tx_buf` function to allocate the transmit buffer in the shared memory.
The buffer is sent only if ``send_choice`` is true. If ``send_choice`` is false the buffer is free using the :c:func:`sm_ipt_free_tx_buf` and the send operation is aborted.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+
+Data receive full-copy
+-------------------
```
peknis, line 120: The length of the underlining must match the number of characters in the title text

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+		/* Use buffer as any other allocated buffer */
+		some_data_preparation_function(buffer);
+
+		return sm_ipt_send(&sm_ipt_ctx, buffer, len);
+	}
+
+The above code uses the ``sm_ipt_alloc_tx_buf`` function to allocate transmit buffer in shared memory. Buffer is prepared by ``some_data_preparation_function`` and sent to another core via ``sm_ipt_send`` function.
+
+Data send abort
+-------------------
```
peknis, line 90: Match the underlining with the title text

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+		memcpy(buffer, data, len);
+
+		return sm_ipt_send(&sm_ipt_ctx, buffer, len);
+	}
+
+The above code uses the ``sm_ipt_alloc_tx_buf`` function to allocate transmit buffer in shared memory. Data is copied into said buffer and sent to another core via ``sm_ipt_send`` function.
+
+Data send no-copy
+-------------------
```
peknis, line 68: Match the underlining with the title text

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+	void data_processing_thread(void)
+	{
+		/* Wait for incoming data */
+		while(!data_received);
+
+		/*Process data */
+		some_data_processing_function(receive_buffer);
+	}
+
+The above code uses the ``memcpy`` function to copy shared memory buffer into local buffer. Next shared memory buffer is freed using ``sm_ipt_free_rx_buf``. Lastly the data is processed locally using a copy of buffer.
```
peknis, line 150:
```suggestion
The above code uses the :c:func:`memcpy` function to copy the shared memory buffer into the local buffer.
The next shared memory buffer is freed using the :c:func:`sm_ipt_free_rx_buf` function
 The data is processed locally using a copy of the buffer.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+
+Data receive no-copy
+-------------------
```
peknis, line 154: Match the underlining with the title text

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+		return sm_ipt_send(&sm_ipt_ctx, buffer, len);
+	}
+
+The above code uses the ``sm_ipt_alloc_tx_buf`` function to allocate transmit buffer in shared memory. Data is copied into said buffer and sent to another core via ``sm_ipt_send`` function.
+
+Data send no-copy
+-------------------
+
+The data is prepared after buffer allocation send to another core when data preparation (and usage) is finished.
```
peknis, line 70: This is very unclear. Please clarify.

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+=====
+
+.. note::
+   Following examples are purely conceptual and simplified to show library functions. They should be rewritten to suit particular need and use OS modules like FIFO.
+
+
+Initialization
+--------------
+
+The initialization function initializes context data and retrieves shared memory parameters from ``sm_ipt_os_init`` function.
```
peknis, line 28:
```suggestion
The initialization function initializes the context data and retrieves the shared memory parameters from the :c:func:`sm_ipt_os_init` function.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+	{
+		return sm_ipt_init(&sm_ipt_ctx, sm_ipt_receive_handler);
+	}
+
+The above code uses the ``sm_ipt_init`` function to initialize ``struct sm_ipt_ctx`` SM IPT context data structure and assign the receive handler.
+
+Data send full-copy
+-------------------
+
+The data is prepared beforehand, copied into allocated buffer and send to another core.
```
peknis, line 50:
```suggestion
The data is prepared beforehand, copied into the allocated buffer and sent to another core.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+		return sm_ipt_send(&sm_ipt_ctx, buffer, len);
+	}
+
+The above code uses the ``sm_ipt_alloc_tx_buf`` function to allocate transmit buffer in shared memory. Buffer is prepared by ``some_data_preparation_function`` and sent to another core via ``sm_ipt_send`` function.
+
+Data send abort
+-------------------
+
+There is possibility to free transmit buffer instead of sending it if the need arises.
```
peknis, line 92:
```suggestion
It is possible to free the transmit buffer instead of sending it.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+		}
+	}
+
+The above code uses the ``sm_ipt_alloc_tx_buf`` function to allocate transmit buffer in shared memory. Buffer is send only if send_choice is true. If send_choice is false the buffer is free using ``sm_ipt_free_tx_buf`` and send operation is aborted.
+
+
+Data receive full-copy
+-------------------
+
+When data is received it can be processed using full-copy mechanism. Buffer has to be copied to local memory. Shared memory rx buffer has to be freed and data processing is done on local copy of buffer.
```
peknis, line 122:
```suggestion
When data is received, it can be processed using the full-copy mechanism.
The buffer must be copied to the local memory.
The shared memory rx buffer must be freed and data processing takes place on the local copy of the buffer.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+		some_data_processing_function(receive_buffer);
+	}
+
+The above code uses the ``memcpy`` function to copy shared memory buffer into local buffer. Next shared memory buffer is freed using ``sm_ipt_free_rx_buf``. Lastly the data is processed locally using a copy of buffer.
+
+
+Data receive no-copy
+-------------------
+
+When data is received it can be processed using no-copy mechanism. The pointer to shared memory buffer has to be saved in order to use it later. Shared memory rx buffer has to be freed and data processing is done on shared memory buffer.
```
peknis, line 156:
```suggestion
When the data is received, it can be processed using the no-copy mechanism.
The pointer to the shared memory buffer must be saved for later use.
The shared memory rx buffer must be freed and data processing takes place on the shared memory buffer.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+.. note::
+   This approach increases risk of out of memory errors due to not freeing buffer right after receiving data.
```
peknis, line 159:
```suggestion
   This approach increases the risk of out-of-memory errors if the buffer is not freed right after receiving data.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+		while(!data_received);
+
+		/*Process data */
+		some_data_processing_function(receive_buffer);
+
+		/* Free receive buffer */
+		sm_ipt_free_rx_buf(&sm_ipt_ctx, packet);
+	}
+
+The above code copies pointer to shared memory buffer. Next data is processed using original buffer in share memory. Lastly shared memory buffer is freed using ``sm_ipt_free_rx_buf``.
```
peknis, line 187:
```suggestion
The above code copies the pointer to the shared memory buffer.
Next data is processed using the original buffer in the shared memory.
The shared memory buffer is freed using the :c:func:`sm_ipt_free_rx_buf` function.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+
+Lower layers
+============
+
+The lower layers of SM IPT are OS-dependent.
+They are responsible for memory management, synchronization and logging support.
+
+.. note::
+   Detailed knowledge about how the lower layers are implemented is not required when using the SM IPT API.
```
peknis, line 197:
```suggestion
   Details about the implementation of the lower layers are not required when using the SM IPT API.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+   However, this knowledge is required to port to a different operating system.
```
peknis, line 198:
```suggestion
   However, this knowledge is needed to port to a different operating system.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,219 @@
...
+
+Operating system abstraction
+----------------------------
+
+The operating system abstraction provides functionality for SM IPT that depends on the operating system.
```
peknis, line 203:
```suggestion
The operating system abstraction provides the SM IPT functionalities that depend on the operating system.
```

peknis:
Please, include also the original image (Visio file?). Someone may want to adjust it at some point. Also, pay attention to the colors and shapes as instructed in https://projecttools.nordicsemi.no/confluence/display/TECHDOC/Figure+guide.

peknis (changes requested):

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,450 @@
...
+	mask = (smask_t)mask >> (blocks - 1);        /* 111000000... (e.g. blocks = 3) */
+	mask = mask >> index;                        /* 000011100... (e.g. index = 4) */
+
+	return mask;
+}
+
+void sm_ipt_alloc_tx_buf(struct sm_ipt_ctx *ctx, uint8_t **buf, size_t len)
+{
+	/* Actual allocated memory: | 32-bit size | data | padding | */
+	size_t blocks = (len + (WORD_SIZE + ctx->out.block_size - 1)) / ctx->out.block_size;
```
KAGA164, line 131: You often use `WORD_SIZE`. In this case it looks like you reserve a word for some special data. Maybe you could use a define for it. I assume that you reserve 4 byte here for data length so maybe:
`#define DATA_LENGTH_FIELD_SIZE WORD_SIZE`

or something similar

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,450 @@
...
+
+			if ((total_size >= &ctx->in.shm_queue->allocable[ctx->in.allocable_size] -
+				(uint8_t *)mem_start) || (total_size < WORD_SIZE)) {
+				memory_corrupted_error();
+				break;
+			}
+
+			SM_IPT_DBG("Message received at: 0x%x, length: %d", (uint32_t)&mem_start[1],
+									(total_size - WORD_SIZE));
+			ctx->receive_handler((uint8_t *)&mem_start[1], total_size - WORD_SIZE);
```
KAGA164, line 370: Please see my previous comment about WORD_SIZE and using define

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,450 @@
...
+	SM_IPT_ERR("Shared memory corrupted"); \
+	SM_IPT_ASSERT(0)
+
+
+struct sm_ipt_shm_queue {
+	uint32_t queue_tx;
+	uint32_t queue_rx;
+	uint8_t queue[QUEUE_ITEMS_CNT];
+	uint8_t handshake;
+	uint16_t _padding; /* Care should be taken to align allocable to word boundary */
```
KAGA164, line 54: I might be wrong but wouldn't this padding will be added automatically by compiler ?

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,450 @@
...
+			blocks = (total_size + ctx->out.block_size - 1) / ctx->out.block_size;
+			free_mask_set(ctx, calc_mask(blocks, block));
+			sm_ipt_os_give(&ctx->out_sem);
+		} else {
+			mem_start = (uint32_t *)(&ctx->in.shm_queue->allocable[
+								block * ctx->in.block_size]);
+			total_size = mem_start[0];
+
+			if ((total_size >= &ctx->in.shm_queue->allocable[ctx->in.allocable_size] -
+				(uint8_t *)mem_start) || (total_size < WORD_SIZE)) {
```
KAGA164, line 363: Please add additional brackets
```
   if ((total_size >= (&ctx->in.shm_queue->allocable[ctx->in.allocable_size] -
                              (uint8_t *)mem_start)) || (total_size < WORD_SIZE)) {
```

sm_ipt/sm_ipt.c
KAGA164: Ok

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,450 @@
...
+
+	mem_start[0] = blocks * ctx->out.block_size;
+
+	*buf = (uint8_t *)(&mem_start[1]);
+	SM_IPT_DBG("Allocated memory at 0x%x", (uint32_t)*buf);
+}
+
+void sm_ipt_free_tx_buf(struct sm_ipt_ctx *ctx, uint8_t *buf)
+{
+	uint32_t *mem_start = (uint32_t *)buf - 1;
```
KAGA164, line 200: Why we need to subtract 1 ?

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,450 @@
...
+
+	sm_ipt_os_unlock(&ctx->out_mutex);
+
+	sm_ipt_os_signal(&ctx->os_ctx);
+}
+
+
+int sm_ipt_send(struct sm_ipt_ctx *ctx, uint8_t *buf, size_t len)
+{
+	uint32_t *mem_start = (uint32_t *)buf - 1;
```
KAGA164, line 257: I see a repeated pattern when you subtract 1. Is it a special value which can be defined ?

sm_ipt/include/sm_ipt.h
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * This function sends data stored at buf in shared memory.
+ *
+ * @param ctx         SM IPT Context.
+ * @param packet      Selected buffer.
+ * @param len         Length of message.
+ *
+ * @return            0 on success or negative error code.
+ */
+int sm_ipt_send(struct sm_ipt_ctx *ctx, uint8_t *buf, size_t len);
```
KAGA164, line 131: Can buf be const ?
```suggestion
int sm_ipt_send(struct sm_ipt_ctx *ctx, const uint8_t *buf, size_t len);
```

KAGA164 (changes requested):

sm_ipt/sm_ipt.c
dchat-nordic: Not necessarily. In this case uint8_t allocable needs to be aligned to word boundary. Allocable is of type uint8_t so compilers will not try to align it to word boundary.

sm_ipt/sm_ipt.c
dchat-nordic: When allocating buffer the first word is buffer size and the beginning of usable data starts at second word. Here (in free functions) The function needs to get the beginning of the buffer, not the usable data, so a decrement of one (word) is required.

sm_ipt/include/sm_ipt.h
doki-nordic: The function changes the content just before sending (the size of the allocated block), so it should be without `const`.

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,420 @@
...
+#define NUM_BLOCKS 64
+#define mask_clz sm_ipt_os_clz64
+#else
+#error Number of shared dynamic memory blocks is not configured.
+#endif
+
+#define ALLOCABLE_MULTIPLY (WORD_SIZE * NUM_BLOCKS)
+#define QUEUE_INDEX_MASK (NUM_BLOCKS - 1)
+
+#define memory_corrupted_error() sm_ipt_os_fatal(); \
```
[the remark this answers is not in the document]
dchat-nordic: I will change it to function in next push.

dchat-nordic:
> Please, include also the original image (Visio file?). Someone may want to adjust it at some point. Also, pay attention to the colors and shapes as instructed in https://projecttools.nordicsemi.no/confluence/display/TECHDOC/Figure+guide.

When I acquire Visio I will attach Visio file.

peknis:
> > Please, include also the original image (Visio file?). Someone may want to adjust it at some point. Also, pay attention to the colors and shapes as instructed in https://projecttools.nordicsemi.no/confluence/display/TECHDOC/Figure+guide.
> 
> When I acquire Visio I will attach Visio file.

It's a pity you cannot get the Visio file. However, the image looks a lot better now and is aligned with the figure guidelines. I think we can live without the native format so far.

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,228 @@
...
+		return sm_ipt_send(&sm_ipt_ctx, buffer, len);
+	}
+
+The above code uses the :c:func:`sm_ipt_alloc_tx_buf` function to allocate the transmit buffer in the shared memory.
+The data is copied into the said buffer and sent to another core through the :c:func:`sm_ipt_send` function.
+
+Data send no-copy
+-----------------
+
+It is possible to directly operate on the transmit buffer in order to avoid copying from the local buffer.
```
peknis, line 71:
```suggestion
It is possible to operate directly on the transmit buffer to avoid copying from the local buffer.
```

peknis (changes requested):

peknis (approved):

KAGA164 (approved):

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,228 @@
...
+----------------------
+
+When data is received, it can be processed using the full-copy mechanism.
+The buffer must be copied to the local memory.
+The shared memory rx buffer must be freed and data processing takes place on the local copy of the buffer.
+
+.. code-block:: c
+
+	volatile uint8_t receive_buffer[MAX_PACKET_SIZE];
+	volatile bool data_received;
```
doki-nordic, line 131: I am not sure that using volatiles int this sample is a good idea. Maybe use some generic functions names, e.g. 

```c
/* Write content of the buffer to some pipe. */
write_bytes_to_pipe(packet, len);

/* Free receive buffer */
sm_ipt_free_rx_buf(&sm_ipt_ctx, packet);
```

And then, `data_processing_thread` is not needed.

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,228 @@
...
+The pointer to the shared memory buffer must be saved for later use.
+The shared memory rx buffer must be freed and data processing takes place on the shared memory buffer.
+
+.. note::
+   This approach increases the risk of out-of-memory errors if the buffer is not freed right after receiving data.
+
+.. code-block:: c
+
+	volatile void *receive_buffer;
+	volatile bool data_received;
```
doki-nordic, line 171: The same comment as above. Maybe something like that:

```c
/* Put the buffer pointer to some FIFO */
put_buffer_ptr_to_fifo(packet, len);
```

in `data_processing_thread`:

```c
/* Wait for incoming data */
get_buffer_ptr_from_fifo(&packet, &len);

/*Process data */
some_data_processing_function(receive_buffer);

/* Free receive buffer */
sm_ipt_free_rx_buf(&sm_ipt_ctx, packet);

```

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,453 @@
...
+#define mask_clz sm_ipt_os_clz64
+#else
+#error Number of shared dynamic memory blocks is not configured.
+#endif
+
+#define ALLOCABLE_MULTIPLY (WORD_SIZE * NUM_BLOCKS)
+#define QUEUE_INDEX_MASK (NUM_BLOCKS - 1)
+
+/* Worst case scenario: NUM_BLOCKS count of messages waiting to be sent and NUM_BLOCKS count of
+ * buffers waiting to be freed + 1 item used as indicator of full queue
```
doki-nordic, line 39: It is not `indicator`. Maybe: `+ 1 item that must be always free in circular buffer`

sm_ipt/sm_ipt.c
doki-nordic: This is GCC specific, but we want to keep the code compiler-independent.

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,453 @@
...
+struct sm_ipt_message {
+	uint32_t size;
+	uint8_t buf[];
+};
+
+
+static void memory_corrupted_error(void)
+{
+	sm_ipt_os_fatal();
+	SM_IPT_ERR("Shared memory corrupted");
```
doki-nordic, line 62: Change order: first show log message, later report fatal error to OS.

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,453 @@
...
+	mask_t sh_mask;
+	bool unset_success;
+	mask_t mask;
+	size_t free_index;
+
+	SM_IPT_DBG("Trying to allocate %d bytes", len);
+
+	if ((blocks > NUM_BLOCKS) || (blocks == 0)) {
+		SM_IPT_ERR("Requested %d bytes, maximum is %d", len,
+				ctx->out.allocable_size - sizeof(uint32_t));
```
doki-nordic, line 151:
```suggestion
				ctx->out.allocable_size - sizeof(struct sm_ipt_message));
```

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,453 @@
...
+		unset_success = free_mask_unset(ctx, mask);
+
+		/* there is a small probability that unset will be unsuccessful */
+		if (!unset_success) {
+			/* give semaphore, because free_mask_unset may cause other thread
+			 * waiting before it reverted the changes
+			 */
+			sm_ipt_os_give(&ctx->out_sem);
+			sem_taken = false;
+			SM_IPT_DBG("Not enough free shared memory. Waiting.");
```
doki-nordic, line 188: Move this message to line 168, before `sm_ipt_os_take(&ctx->out_sem);`. We are not waiting here, but there.

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,453 @@
...
+
+	SM_IPT_ASSERT((uint32_t)ctx->out.shm_queue.allocable % WORD_SIZE == 0);
+	SM_IPT_ASSERT((uint32_t)ctx->in.shm_queue.allocable % WORD_SIZE == 0);
+
+	ctx->out.allocable_size = ((ctx->os_ctx.out_total_size - sizeof(struct sm_ipt_queue)) /
+							ALLOCABLE_MULTIPLY) * ALLOCABLE_MULTIPLY;
+	ctx->in.allocable_size = ((ctx->os_ctx.in_total_size - sizeof(struct sm_ipt_queue)) /
+							ALLOCABLE_MULTIPLY) * ALLOCABLE_MULTIPLY;
+	ctx->in.block_size = ctx->in.allocable_size / NUM_BLOCKS;
+	ctx->out.block_size = ctx->out.allocable_size / NUM_BLOCKS;
```
doki-nordic, line 433: Maybe this will make calculations more clear:

```c
ctx->in.block_size = (ctx->os_ctx.in_total_size - sizeof(struct sm_ipt_queue)) / NUM_BLOCKS / sizeof(uint32_t) * sizeof(uint32_t);
ctx->out.block_size = ...
ctx->in.allocable_size = ctx->in.block_size * NUM_BLOCKS;
ctx->out.allocable_size = ...
```

And `ALLOCABLE_MULTIPLY` is not needed any more.

doki-nordic (changes requested):

sm_ipt/sm_ipt.c
dchat-nordic: This solution is more readable, but I think WORD_SIZE should be used instead of sizeof(uint32_t).

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,214 @@
...
+
+It is possible to free the transmit buffer instead of sending it.
+
+.. code-block:: c
+
+	int send_if(size_t len)
+	{
+		uint8_t *buffer;
+		bool preparation_success;
+		size_t i;
```
doki-nordic, line 102: `i` is unused.

doki-nordic (changes requested):

doki-nordic:
@peknis, There are small changes in `usage.rst` since you last review (https://github.com/nrfconnect/sdk-nrfxlib/compare/129c3ec..390a82adb232061ab0d4212ecd074e0651649d07)

Can you look at it?

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,213 @@
...
+		} else {
+			sm_ipt_free_tx_buf(&sm_ipt_ctx, buffer);
+
+			return ERROR_ABORT;
+		}
+	}
+
+The above code uses the :c:func:`sm_ipt_alloc_tx_buf` function to allocate the transmit buffer in the shared memory.
+The buffer is sent only if ``preparation_success`` is true.
+If ``preparation_success`` is false the buffer is free using the :c:func:`sm_ipt_free_tx_buf` and the send operation is aborted.
```
peknis, line 119:
```suggestion
If ``preparation_success`` is false, the buffer is freed using the :c:func:`sm_ipt_free_tx_buf` function and the send operation is aborted.
```

sm_ipt/doc/usage.rst
```diff
@@ -0,0 +1,213 @@
...
+		get_buffer_ptr_from_fifo(&packet, &len);
+
+		/*Process data */
+		some_data_processing_function(packet, len);
+
+		/* Free receive buffer */
+		sm_ipt_free_rx_buf(&sm_ipt_ctx, packet);
+	}
+
+The above code puts the pointer to the shared memory buffer to the local fifo using :c:func:`put_buffer_ptr_to_fifo` function.
```
peknis, line 179:
```suggestion
The above code sets the pointer to the shared memory buffer to the local fifo using the :c:func:`put_buffer_ptr_to_fifo` function.
```

peknis (changes requested):

doki-nordic (approved):

peknis (approved):

sm_ipt/sm_ipt.c
```diff
@@ -0,0 +1,452 @@
...
+{
+	uint32_t tx;
+	uint32_t dst;
+
+	sm_ipt_os_lock(&ctx->out_mutex);
+
+	tx = ctx->out.shm_queue->queue_tx;
+	dst = tx;
+
+	if (dst >= QUEUE_ITEMS_CNT) {
```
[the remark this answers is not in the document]
dchat-nordic: Done

carlescufi:
Should we close this for now? It can always be reopened

PR #525, opened 2021-07-27
Nrf rpc sm_ipt backend

dchat-nordic:
Added support for sm_ipt backend
Updated templates, updated Kconfig
to support sm_ipt.

Depends on https://github.com/nrfconnect/sdk-nrfxlib/pull/524

Signed-off-by: Dominik Chat <dominik.chat@nordicsemi.no>

Co-authored-by: Dominik Kilian <Dominik.Kilian@nordicsemi.no>


nrf_rpc/include/nrf_rpc.h
```diff
@@ -28,6 +28,14 @@ extern "C" {
 /* Internal definition used in the macros. */
 #define _NRF_RPC_HEADER_SIZE 4
 
+#ifndef NRF_RPC_TR_MAX_HEADER_SIZE
+#define NRF_RPC_TR_MAX_HEADER_SIZE 0
```
doki-nordic, line 32: Don't define them here. Transport header file is responsible for defining it.
The same for `NRF_RPC_TR_AUTO_FREE_RX_BUF`.

nrf_rpc/nrf_rpc.c
```diff
@@ -10,9 +10,9 @@
 #include <stdbool.h>
 #include <string.h>
 
-#include "nrf_rpc.h"
-#include "nrf_rpc_tr.h"
-#include "nrf_rpc_os.h"
+#include <nrf_rpc.h>
+#include <nrf_rpc_tr.h>
+#include <nrf_rpc_os.h>
```
doki-nordic, line 15: Don't change this.

nrf_rpc/nrf_rpc.c
```diff
@@ -767,6 +768,7 @@ int nrf_rpc_init(nrf_rpc_err_handler_t err_handler)
 	return err;
 }
 
+
```
doki-nordic, line 771: Remove unnecessary line

nrf_rpc/sm_ipt_backend.c
```diff
@@ -0,0 +1,3 @@
+#include "sm_ipt_backend.h"
```
doki-nordic, line 1: Add copyright comment at the beginning of a file.

nrf_rpc/template/nrf_rpc_os_tmpl.h
```diff
@@ -43,10 +43,11 @@ typedef void (*nrf_rpc_os_work_t)(const uint8_t *data, size_t len);
  *
  * @param callback Work callback that will be called when something was send
  *                 to a thread pool.
+ * @param user_data Data pointer passed to ipm device callbacks
  *
  * @return         0 on success or negative error code.
  */
-int nrf_rpc_os_init(nrf_rpc_os_work_t callback);
+int nrf_rpc_os_init(nrf_rpc_os_work_t callback, void *user_data);
```
doki-nordic, line 50: `user_data` is not longer needed. Right?

nrf_rpc/Kconfig
```diff
@@ -26,6 +26,13 @@ config NRF_RPC_TR_RPMSG
 	help
 	  If enabled selects RPMsg as a transport layer for nRF PRC.
 
+config NRF_RPC_TR_SHMEM
+	bool "nRF RPC over shared memory"
+	select SM_IPT
+	help
+	  If enabled selects shared memory as a transport layer for nRF PRC.
+	  It requires additional API implemented in nrf_rpc_os.h file.
+
```
doki-nordic, line 35: We are using SM IPT module, so we can use its name instead of just `shared memory`.

nrf_rpc/include/nrf_rpc.h
```diff
@@ -230,9 +238,17 @@ struct nrf_rpc_err_report {
  *                     a newly allocated packet buffer.
  * @param[in]  _len    Requested length of the packet.
  */
+#ifdef nrf_rpc_tr_alloc_tx_buf
```
doki-nordic, line 241: Maybe it is better to explain this `#ifdef`, because it is not obvious.

```c
/* If nrf_rpc_tr_alloc_tx_buf is a macro, put it directly, because it may allocate memory on stack. */
```

nrf_rpc/nrf_rpc.c
```diff
@@ -95,6 +95,7 @@ static struct nrf_rpc_cmd_ctx *cmd_ctx_alloc(void)
 
 static void cmd_ctx_free(struct nrf_rpc_cmd_ctx *ctx)
 {
+	NRF_RPC_DBG("Command context %d free", ctx->id);
```
doki-nordic, line 98: This line can be deleted, it is not related to this PR.

doki-nordic (changes requested):

nrf_rpc/nrf_rpc.c
dchat-nordic: I tried to make it consistent with nrt_rpc.h

nrf_rpc/include/nrf_rpc.h
dchat-nordic: This is ifdef def, so if transport have not defined anything the values of macros are set to default.

nrf_rpc/nrf_rpc_cbor.c
```diff
@@ -227,6 +227,9 @@ void _nrf_rpc_cbor_proxy_handler(const uint8_t *packet, size_t len,
 
 void _nrf_rpc_cbor_prepare(struct nrf_rpc_cbor_ctx *ctx, size_t len)
 {
+#ifndef nrf_rpc_tr_alloc_tx_buf
```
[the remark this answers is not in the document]
dchat-nordic: Transport layer provides this function (nrf_rpc_tr.h). Here it is as a template (*_tmpl.h)

nrf_rpc/nrf_rpc.c
```diff
@@ -801,3 +801,17 @@ void nrf_rpc_err(int code, enum nrf_rpc_err_src src,
...
 	}
 }
+
+#ifndef nrf_rpc_tr_alloc_tx_buf
+
+uint8_t *_nrf_rpc_alloc(size_t len)
+{
+	uint8_t *packet;
+
+	NRF_RPC_ASSERT(len < 0x70000000);
```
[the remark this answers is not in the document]
dchat-nordic: Size limit check. Changed to define now.

nrf_rpc/nrf_rpc.c
```diff
@@ -801,3 +801,17 @@ void nrf_rpc_err(int code, enum nrf_rpc_err_src src,
...
+
+	nrf_rpc_tr_alloc_tx_buf(&packet, _NRF_RPC_HEADER_SIZE + len);
+	return &packet[_NRF_RPC_HEADER_SIZE];
```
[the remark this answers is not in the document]
dchat-nordic: Added

tejlmand (changes requested):
Why is this in `nrfxlib` and not in [sdk-nrf/lib](https://github.com/nrfconnect/sdk-nrf/tree/main/lib) ?

I don't see any prebuilt libraries in this PR, which is usually the case for nrfxlib libs.

doki-nordic:
> Why is this in `nrfxlib` and not in [sdk-nrf/lib](https://github.com/nrfconnect/sdk-nrf/tree/main/lib) ?
> 
> I don't see any prebuilt libraries in this PR, which is usually the case for nrfxlib libs.

@tejlmand, just FYI, the `nrfxlib` is not only for prebuild libraries. It is place for modules that are independent from Zephyr/NCS, so user can take some module from nrfxlib and use it e.g. in FreeRTOS or baremetal. Other example is `nrf_rpc` in this repo. If this PR will be reopen then it will be exactly for such case.

PR #527, opened 2021-07-27
doc: use kconfig role

gmarull:
References to Kconfig options require usage of the :kconfig: role.


doki-nordic (approved):

mbolivar-nordic:
Let's use #528 from my fork

PR #530, opened 2021-08-04
nrfxlib: add common nrf_errno.h file

lemrey:
Add a common nrf_errno.h file with errnos from IEEE Std 1003.1-2017.
The values are the same from this PR: https://github.com/zephyrproject-rtos/zephyr/pull/35291/files, the errno description is copy-pasted from the standard (there were differences from that PR).

Still not included by the build system, but each binary library can be updated to include it when it's released next, as necessary.
Some errnos are not in the standard, I grouped them at the bottom; what do you think?

### DIFF
--- a/include/nrf_errno.h
+++ b/include/nrf_errno.h
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/**
+ * @file
+ * @brief nRF error numbers
+ */
+
+/**
+ * @brief nRF error numbers
+ * A list of error codes defined by IEEE Std 1003.1-2017 found at:
+ * https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html
+ * @defgroup nrf_errno nRF error numbers
+ * @{
+ */
+
+#define NRF_EPERM 1             /**< Operation not permitted */
+#define NRF_ENOENT 2            /**< No such file or directory */
+#define NRF_ESRCH 3             /**< No such process */
+#define NRF_EINTR 4             /**< Interrupted function */
+#define NRF_EIO 5               /**< I/O error */
+#define NRF_ENXIO 6             /**< No such device or address */
+#define NRF_E2BIG 7             /**< Argument list too long */
+#define NRF_ENOEXEC 8           /**< Executable file format error */
+#define NRF_EBADF 9             /**< Bad file descriptor */
+#define NRF_ECHILD 10           /**< No child processes */
+#define NRF_EAGAIN 11           /**< Resource unavailable, try again */
+#define NRF_ENOMEM 12           /**< Not enough space */
+#define NRF_EACCES 13           /**< Permission denied */
+#define NRF_EFAULT 14           /**< Bad address */
+#define NRF_EBUSY 16            /**< Device or resource busy */
+#define NRF_EEXIST 17           /**< File exists */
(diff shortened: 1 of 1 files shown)


include/nrf_errno.h
```diff
@@ -0,0 +1,107 @@
...
+#define EWOULDBLOCK EAGAIN /**< Operation would block */
+
+/* Not in IEEE Std 1003.1-2017 */
+
+#define ENOTBLK 15          /**< Block device required */
+#define EPFNOSUPPORT 96     /**< Protocol family not supported */
+#define ESHUTDOWN 110       /**< Can't send after socket shutdown */
+#define EHOSTDOWN 117       /**< Host is down */
+#define ESOCKTNOSUPPORT 124 /**< Socket type not supported */
+#define ETOOMANYREFS 129    /**< Too many references: can't splice */
```
doki-nordic, line 103: I would prefer to put them together in the same list as it is in Zephyr. The error codes must be the same as in Zephyr, so keeping the file similar to Zephyr makes it easier to keep them aligned.

include/nrf_errno.h
```diff
@@ -0,0 +1,107 @@
+/*
```
doki-nordic, line 1: I am concern about the file name.

So, every module from nrfxlib, that needs errno, will include this file. If any c file will include this file and system's errno.h directly or indirectly, it will have two definition. This is ok for Zephyr, because we aligned all the numbers with it.

Problem is that nrfxlib is not only for Zephyr. I don't know about other modules that uses errno, but nRF RPC is system-independent and will allow calling remote procedures even from Linux (when serial backend will be finished). Linux errno.h is different than this, so nRF RPC will not compile on Linux if we switch to this solution.

My suggestion is to use name `errno.h`. Then, the build system must ensure that errno.h from system is used if available, and this file if not. I don't know how about binary modules. Does anyone allows to use them outside Zephyr, where system's `errno.h` may be incompatible? If yes, maybe we should consider some different approach?

doki-nordic (changes requested):

include/nrf_errno.h
doki-nordic: But it is just my opinion. I am not insisting on it.

include/nrf_errno.h
lemrey: Yes, there are binary libraries which are not necessarily tied to Zephyr. When using those, the errno definitions must stay the same as those that the library was compiled with. What happens right now is that errnos are glued for those libraries, and I don't see a way around that.

I updated the PR to prefix all the errnos with the `NRF_` prefix.
I think that source libraries or libraries that are Zephyr-only may include just `errno.h` and be happy.
For those _binary_ libraries that are OS-independent, we'll have to glue so that the application can compile with whatever errno.h definitions that they have.

doki-nordic: Yes, from nRF RPC perspective, this is acceptable.

doki-nordic: We should talk with verification team to consider some automated tests that checks if all defined symbols are aligned with Zephyr. Now, we are aware, but in a few years maybe someone will want to change this file or Zephyr's and make them incompatible.

Different approach for testing would be put `BUILD_ASSERT`s in the file. I don't remember if there is a definition that allows us detect if we are compiling in Zephyr, but let's assume for this example that it is `IS_ZEPHYR`.

```c
#ifdef IS_ZEPHYR
#include <kernel.h>
#include <errno.h>
BUILD_ASSERT(NRF_EPERM == EPERM, "Invalid NRF_EPERM value");
BUILD_ASSERT(NRF_ENOENT == ENOENT, "Invalid NRF_ENOENT value");
...
```

doki-nordic (approved):

PR #531, opened 2021-08-05
nrf_security: Make SHA_1 opt in for TF-M builds

Vge0rge:
-This disables the SHA-1 by default when building
 with TF-M. Since it is a legacy protocol now
 users that need it can just enable it manually.

 Ref: NCSDK-10177


tejlmand (approved):
personally I think we could always make it optionally, but doing it only for TF-M enabled builds is a start.

oyvindronningstad (approved):

hakonfam (approved):

PR #532, opened 2021-08-06
nrf_security: set config type of mbedcrypto to string

hakonfam:
The path to the generated mbedcrypto config file is not known at the
time of passing the arguments to the TF-M build system. Hence it is
only possible for us to pass the name of the file at this point. We
already know that the folder in which the file is generated is added
to the include path, so only the name is needed.

The variable passed 'TFM_MBEDCRYPTO_CONFIG_PATH' is by default of
type 'PATH' which results in its value being resolved as a relative
path, which is incorrect.

To get around this issue, we leverage the newly added option
'TFM_MBEDCRYPTO_CONFIG_TYPE' so that we can set it to type
'STRING'. This way there is no relative path resolving, and the
generated file is found by the TF-M build system.

Ref: NCSDK-10706

Signed-off-by: Håkon Øye Amundsen <haakon.amundsen@nordicsemi.no>


tejlmand (approved):

mbolivar-nordic:
Thanks @hakonfam. If you need this in the current upmerge, please send it to my branch: https://github.com/mbolivar-nordic/sdk-nrfxlib/tree/upmerge-2021-07-26

hakonfam:
https://github.com/mbolivar-nordic/sdk-nrfxlib/pull/3

Added DNM here to avoid it being merged straight to master

mbolivar-nordic:
> [mbolivar-nordic#3](https://github.com/mbolivar-nordic/sdk-nrfxlib/pull/3)
> 
> Added DNM here to avoid it being merged straight to master

Merged there now; thanks

PR #535, opened 2021-08-15
crypto: correcting typo

soburi:
mutex->flag is undefined symbol.
correcting to mutex->flags.


Vge0rge:
Thank you for the patch @soburi! The compliance failed because your name in the commit includes "," which is not supported, update it and we will see how that goes. If it continues to fail ping me ;)

soburi:
@Vge0rge, 

I re-pushed with correcting name.
If it was wronge, tell me please.

Vge0rge:
> @Vge0rge,
> 
> I re-pushed with correcting name.
> If it was wronge, tell me please.

Name is good now, now we need a code owner to approve.

Vge0rge:
@soburi Since it was just a small fixed I included it in the release then. Thanks for reporting!

PR #537, opened 2021-08-17
crypto: Fix oberon path when FPU is enabled

Vge0rge:
-This fixes a build issue when CONFIG_FPU is enabled
 with TF-M builds.

Ref: NCSDK-10877

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -11,6 +11,12 @@ nrfxlib_calculate_lib_path(lib_path)
 add_library(nrfxlib_crypto INTERFACE)
 
 if (CONFIG_NRF_OBERON OR CONFIG_BUILD_WITH_TFM)
+
+  # The oberon library doesn't have a softfp version so we replace its path with the soft-float one
+  if (CONFIG_FP_SOFTABI)
+        string(REGEX REPLACE "softfp-float" "soft-float"  lib_path ${lib_path})
+  endif()
+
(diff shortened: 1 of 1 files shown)


torsteingrindvik (approved):

magnev (approved):

SebastianBoe:
Does the rest of the FW also then use soft-float.

Or is the FPU still in use, even though the Oberon code specifically does not use it?

SebastianBoe (changes requested):
PR to nrf so that we get test coverage?

Vge0rge:
> Does the rest of the FW also then use soft-float.
> 
> Or is the FPU still in use, even though the Oberon code specifically does not use it?

 Since this is only happening when the CONFIG_FPU is enabled I assume that the rest of the firmware indeed is using softfp-float based on common.cmake: 
https://github.com/nrfconnect/sdk-nrfxlib/blob/master/common.cmake#L30

So the main difference here seems to be that oberon does not support it and so we need to use the soft-float version even if we have CONFIG_FPU enabled.

SebastianBoe (changes requested):
I wish to test this manually before it is merged.

SebastianBoe (commented):
LGTM.

Generally, this is dangerous.

If the oberon library had FPU instructions or references to soft-float library calls this would crash.

But it doesn't, so this works.

SebastianBoe (approved):

PR #541, opened 2021-08-25
doc: style fix to radio coexistence documentation

wiba-nordic:
Fixes abbreviated mentions of nRF Connect SDK.
Also removes reference to a non-public document.

This was supposed to be part of PR #534 but was causing it to fail compliance for some reason.

Signed-off-by: Wille Backman <wille.backman@nordicsemi.no>

### DIFF
--- a/mpsl/doc/cx.rst
+++ b/mpsl/doc/cx.rst
@@ -30,14 +30,14 @@ This experimental feature has several limitations:
 Implementation
 **************
 
-To use the radio coexistence feature, you can either select the implementation available in NCS or replace it with another implementation compatible with your PTA.
+To use the radio coexistence feature, you can either select the implementation available in nRF Connect SDK or replace it with another implementation compatible with your PTA.
 Each implementation exposes the same API to allow for an easier replacement that does not require modifications to the radio protocols.
 
-The radio coexistence implementation provided in NCS is the following:
+The radio coexistence implementation provided in nRF Connect SDK is the following:
 
-* *Thread Radio Coexistence*.
-  This PTA interface is compatible with the 3-wire PTA interface defined in the *Thread Radio Coexistence recommendations* document.
-  It is also compatible with any IEEE 802.15.4 protocol (including Thread), with |BLE|, and with other radio protocols.
+*Thread Radio Coexistence*
+   This PTA interface is compatible with the 3-wire PTA interface defined in the *Thread Radio Coexistence recommendations* document available to Thread Group members.
+   It is also compatible with any IEEE 802.15.4 protocol (including Thread), with |BLE|, and with other radio protocols.
 
 Configuration
 *************
@@ -64,7 +64,7 @@ The radio protocol must register a callback function in the radio coexistence mo
 Implementation of a new radio coexistence module
 ************************************************
 
-If the radio coexistence implementation provided in NCS is not compatible with the PTA used in your system, you can implement a software module compatible with your PTA that exposes the radio coexistence API, using the :file:`nrf/subsys/mpsl/cx/thread/mpsl_cx_thread.c` file as an example.
+If the radio coexistence implementation provided in nRF Connect SDK is not compatible with the PTA used in your system, you can implement a software module compatible with your PTA that exposes the radio coexistence API, using the :file:`nrf/subsys/mpsl/cx/thread/mpsl_cx_thread.c` file as an example.
 
 The radio coexistence implementation must implement the interface defined in the ``mpsl_cx_interface_t`` structure in the :file:`nrfxlib/mpsl/include/mpsl_cx_abstract_interface.h` file.
 Each function from this interface is described in details in the :file:`nrfxlib/mpsl/include/protocol/mpsl_cx_protocol_api.h` file.


mpsl/doc/cx.rst
```diff
@@ -30,13 +30,13 @@ This experimental feature has several limitations:
 Implementation
 **************
 
-To use the radio coexistence feature, you can either select the implementation available in NCS or replace it with another implementation compatible with your PTA.
+To use the radio coexistence feature, you can either select the implementation available in nRF Connect SDK or replace it with another implementation compatible with your PTA.
 Each implementation exposes the same API to allow for an easier replacement that does not require modifications to the radio protocols.
 
-The radio coexistence implementation provided in NCS is the following:
+The radio coexistence implementation provided in RF Connect SDK is the following:
```
rugeGerritsen, line 36:
```suggestion
The radio coexistence implementation provided in nRF Connect SDK is the following:
```

mpsl/doc/cx.rst
```diff
@@ -30,13 +30,13 @@ This experimental feature has several limitations:
...
 
 * *Thread Radio Coexistence*.
-  This PTA interface is compatible with the 3-wire PTA interface defined in the *Thread Radio Coexistence recommendations* document.
```
rugeGerritsen, line 39: @hubertmis , are you ok with these changes?

rugeGerritsen (changes requested):
Please fix the typos

mpsl/doc/cx.rst
wiba-nordic: Well, that's embarrassing.

mpsl/doc/cx.rst
wiba-nordic: To explain, I took out the reference because, to my knowledge, the document is non-public. In that case, referencing it seems odd.

hubertmis: Thread protocol specification has the same confidentiality level (it's available only for Thread Group members), but we mention it in Thread protocol specification. I don't think it's good to avoid mentioning which specification was used as a base of implementation, only because not everyone has access to the spec.

hubertmis (changes requested):
I'm reluctant to remove reference to the specification used as a base of implementation. We don't remove such references for other non-public specifications like Zigbee protocol.

Without this reference it's not clear how this implementation is intended to work.

mpsl/doc/cx.rst
wiba-nordic: Would an acceptable alternative then be:
"This PTA interface is compatible with the 3-wire PTA interface defined in *Thread Radio coexistence recommendations*."

This removes reference to the document, but keeps the source specification.

hubertmis: I think it's acceptable, but it makes this sentence ambiguous.

Let's say a Thread Group member reads cx.rst documentation. They understand "Thread Radio Coexistence" implementation is compatible with some _Thread Radio coexistence recommendations_. But what are these recommendations? Did documentation author mean the *Thread Radio Coexistence recommendations* document? Any other document? Or recommendations that were discussed by one of the working groups?

I would keep the original sentence, because it was clear.

mpsl/doc/cx.rst
```diff
@@ -30,14 +30,14 @@ This experimental feature has several limitations:
...
 Each implementation exposes the same API to allow for an easier replacement that does not require modifications to the radio protocols.
 
-The radio coexistence implementation provided in NCS is the following:
+The radio coexistence implementation provided in nRF Connect SDK is the following:
 
-* *Thread Radio Coexistence*.
-  This PTA interface is compatible with the 3-wire PTA interface defined in the *Thread Radio Coexistence recommendations* document.
-  It is also compatible with any IEEE 802.15.4 protocol (including Thread), with |BLE|, and with other radio protocols.
+*Thread Radio Coexistence*
+   This PTA interface is compatible with the 3-wire PTA interface  defined in the *Thread Radio Coexistence recommendations* document.
```
rugeGerritsen, line 39: Please remove additional spaces

mpsl/doc/cx.rst
wiba-nordic: Alright, I reverted the sentence to its original form, though I changed the preceding line from a bullet point to a term definition.

I still think it is an odd reference in the context that the reader does not already know of the existence of that specific document. For example, googling the document name gave me absolutely no related results. If the reader is not already familiar, they have no way of knowing where to gain more information. I gues it's a case of assuming a certain level of familiarity with Thread for anyone who reads the `cx.rst`.

hubertmis: We can clarify for users unfamiliar with this document, that it is available for Thread Group members, like a footnote or something

hubertmis: I can provide a link to the document. It leads to Thread Group portal and only registered users have access

wiba-nordic: Can the Thread Group be credited as an author for the document? That gives an indication of where to find more information. So "[...} document by the Thread Group."

rugeGerritsen (approved):

hubertmis (approved):
Looks very good :+1:

PR #550, opened 2021-09-06
nrf_security: symbol rename script now passes settings through script

tejlmand:
The symbol rename archive script used a generator expression to pass
all objects to the symbol_rename_archive_script.cmake among other
arguments.

When there are many objects in the library, then the command execution
string becomes very long.

On Windows, long execution commands are
wrapped inside a bat-file.

This might look like:
...\post-build.bat 7eb3ae9870a59d55

Resulting in SES-NE to try to execute the post-build.bat including the
following SHA as a single command, which fails.

This is fixed by generating a CMake file with required arguments to
symbol_rename_archive_script.cmake, and the pass this settings file as
argument to the script.

Fixes: NCSDK-10510

### DIFF
--- a/nrf_security/cmake/extensions.cmake
+++ b/nrf_security/cmake/extensions.cmake
@@ -649,15 +649,24 @@ function(nrf_security_target_embed_objects)
         ${CMAKE_CURRENT_BINARY_DIR}/symbol_rename_${backend_name}.txt
       )
 
+      set(symbol_rename_include"\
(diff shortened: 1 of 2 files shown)


SebastianBoe (approved):
Nice

osaether (approved):
Tested on Windows 10 with nRF Connect+Toolchain Manager and it works now. Thanks!

shantha-14 (approved):

SebastianBoe:
Waiting for manifest CI to run.

PR #566, opened 2021-10-06
cmake: Check for availability of softfp libararies

Vge0rge:
-When FPU and the SOFTABI is enabled the cmake code
 tries to use the softfp version of libraries. Many
 libraries in nrfxlib do not have a softfp version but
 they do have a soft-float version which is compatible
 with the SOFTABI. This adds a CMake function which checks
 for the availability of a softfp library and if it is not
 available it is using the soft-float version instead.
-The libraries where this function is applied to are: crypto,
 mpsl, nfc, nrf_802154, nrf_modem and softdevice_controller

Ref: NCSDK-11600

### DIFF
--- a/common.cmake
+++ b/common.cmake
@@ -41,6 +41,18 @@ function(nrfxlib_calculate_lib_path lib_path)
   set(${lib_path} "lib/${arch_soc_dir}/${float_dir}${short_wchar}" PARENT_SCOPE)
 endfunction()
 
+function(nrfxlib_check_softfp_path lib_name)
+  if (CONFIG_FP_SOFTABI)
+    set(lib_full_path ${${lib_name}})
+    string(REGEX REPLACE "softfp-float" "soft-float" soft_path ${lib_full_path})
+
+    if(NOT EXISTS ${lib_full_path} AND EXISTS ${soft_path})
+      message("There is no softfp version for library ${lib_name}, using the soft-float version.")
+      set(${lib_name} ${soft_path} PARENT_SCOPE)
+    endif()
+  endif()
+endfunction()
+
 function(get_mbedtls_dir ARM_MBEDTLS_PATH_ARG)
   if(EXISTS ${${ARM_MBEDTLS_PATH_ARG}})
   # Do nothing, just use the provided path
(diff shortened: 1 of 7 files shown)


SebastianBoe:
Relates to https://github.com/nrfconnect/sdk-nrfxlib/pull/537

SebastianBoe:
We have tested that this is safe when there is no FP, neither soft or hard, in the library. I am not sure about what happens if there is soft float in the library. Do we then link in the appropriate library to provide soft implementations of FP?

Does anyone know if their library definitely does or doesn't contain FP (soft or hard) ?

If all libraries touched here definitely don't contain any FP then this PR is safe.

rugeGerritsen (approved):

frkv:
> We have tested that this is safe when there is no FP, neither soft or hard, in the library. I am not sure about what happens if there is soft float in the library. Do we then link in the appropriate library to provide soft implementations of FP?
> 
> Does anyone know if their library definitely does or doesn't contain FP (soft or hard) ?
> 
> If all libraries touched here definitely don't contain any FP then this PR is safe.

I can only talk for the crypto library. We don't call any floating point operations. 

The only think I can think of with regards to floating point in generic libraries like this is through printf-calls (which may support floating points as input)....

frkv (approved):
LGTM for crypto

common.cmake
```diff
@@ -41,6 +41,18 @@ function(nrfxlib_calculate_lib_path lib_path)
   set(${lib_path} "lib/${arch_soc_dir}/${float_dir}${short_wchar}" PARENT_SCOPE)
 endfunction()
 
+function(nrfxlib_check_softfp_path lib_name)
+  if (CONFIG_FP_SOFTABI)
+    set(lib_full_path ${${lib_name}})
+    string(REGEX REPLACE "softfp-float" "soft-float" soft_path ${lib_full_path})
```
tejlmand, line 47: there is a minor risk that if someone has installed the workspace inside:
`/path/softfp-float/.../nrfxlib` then this will fail as all occurrences of `softfp-float` will be replaced.

Just as a general precaution, and because of the risk of such constructs might be copied elsewhere in the system, could you do the path replacing on the relative path you have, for example here:
https://github.com/nrfconnect/sdk-nrfxlib/blob/f25aa4ddee87cc525f66d9de3808426b2ef36afb/common.cmake#L41
or here:
https://github.com/nrfconnect/sdk-nrfxlib/blob/f25aa4ddee87cc525f66d9de3808426b2ef36afb/crypto/CMakeLists.txt#L18

before you make the path absolute ?

PR #570, opened 2021-10-26
crypto: Don't enable MD5 unless NET_TCP is enabled

SebastianBoe:
Instead of always enabling MD5 we enable it only when NET_TCP is
enabled.

This saves us a significant amount of flash.

We could make additional savings by only enabling it in certain types
of TCP builds, but it is not clear which option to depend on.


SebastianBoe:
This has  been cherry-picked into another PR.

PR #578, opened 2021-11-04
nrf_security: don't use downstream patch for setting config path

hakonfam:
The patch for adding a variable type for the mbedcyrpto config
patch cmake variable was not accepted upstream. Hence, we need
to provide the path to the configuration file relative to the
working directory of the TF-M build system (build/tfm).

Ref: NCSDK-12216

Signed-off-by: Håkon Øye Amundsen <haakon.amundsen@nordicsemi.no>


nrf_security/CMakeLists.txt
```diff
@@ -55,8 +55,7 @@ if(CONFIG_BUILD_WITH_TFM)
   set_property(TARGET zephyr_property_target
     APPEND PROPERTY TFM_CMAKE_OPTIONS
       -DNRF_SECURITY_SETTINGS=\"$<TARGET_PROPERTY:nrf_security_target,CMAKE_ARGS>\"
-      -DTFM_MBEDCRYPTO_CONFIG_TYPE=STRING
-      -DTFM_MBEDCRYPTO_CONFIG_PATH=$<TARGET_PROPERTY:nrf_security_target,MBEDTLS_CFG_FILE>
+      -DTFM_MBEDCRYPTO_CONFIG_PATH=secure_fw/partitions/crypto/mbedcrypto/include/$<TARGET_PROPERTY:nrf_security_target,MBEDTLS_CFG_FILE>
```
tejlmand, line 58: What about simply do:
```suggestion
      -DTFM_MBEDCRYPTO_CONFIG_PATH:STRING=$<TARGET_PROPERTY:nrf_security_target,MBEDTLS_CFG_FILE>
```

tejlmand (changes requested):
small suggestion that in theory should work.

nrf_security/CMakeLists.txt
hakonfam: Why didn't we do it like this in the first place? :D Great solution (y)

tejlmand (approved):
**hint** consider to take the PR out of draft.

frkv (approved):
LGTM, neat trick changing the type this way!

hakonfam:
https://github.com/nrfconnect/sdk-nrfxlib/pull/579

PR #583, opened 2021-11-10
nrf_security: Update mcuboot mbedcrypto config file

joerchan:
Update mcuboot mbedcrypto config file name. This was always the name,
but commit message that added it uses the other name, probably changed
during code-review in trusted-firmware-m.

Don't set the config unless BL2 is enabled. Otherwise this will produce
a cmake warning when the config is not used.

### DIFF
--- a/nrf_security/CMakeLists.txt
+++ b/nrf_security/CMakeLists.txt
@@ -56,8 +56,14 @@ else()
   set_property(TARGET zephyr_property_target
     APPEND PROPERTY TFM_CMAKE_OPTIONS
       -DTFM_MBEDCRYPTO_CONFIG_PATH:STRING=${CONFIG_MBEDTLS_CFG_FILE}
-      -DMCUBOOT_MBEDCRYPTO_CONFIG_PATH=${CONFIG_MBEDTLS_CFG_FILE}
   )
+
+  if(CONFIG_TFM_BL2)
+    set_property(TARGET zephyr_property_target
+      APPEND PROPERTY TFM_CMAKE_OPTIONS
+        -DMCUBOOT_MBEDCRYPTO_CONFIG_PATHPATH:STRING=${CONFIG_MBEDTLS_CFG_FILE}
+    )
+  endif()
 
   if(CONFIG_MBEDTLS_ENABLE_HEAP)
     # The ENGINE_BUF_SIZE holds the dynamic allocation buffer for


nrf_security/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ if(CONFIG_BUILD_WITH_TFM)
     APPEND PROPERTY TFM_CMAKE_OPTIONS
       -DNRF_SECURITY_SETTINGS=\"$<TARGET_PROPERTY:nrf_security_target,CMAKE_ARGS>\"
       -DTFM_MBEDCRYPTO_CONFIG_PATH:STRING=$<TARGET_PROPERTY:nrf_security_target,MBEDTLS_CFG_FILE>
-      -DMCUBOOT_MBEDCRYPTO_CONFIG_PATH=$<TARGET_PROPERTY:nrf_security_target,MBEDTLS_CFG_FILE>
+      -DMCUBOOT_MBEDCRYPTO_CONFIG_FILEPATH=$<TARGET_PROPERTY:nrf_security_target,MBEDTLS_CFG_FILE>
```
frkv, line 59: I recommend adding the same trick like we have done here: dbcb163, in case we add support for building MCUboot inside TF-M.

-DMCUBOOT_MBEDCRYPTO_CONFIG_FILEPATH:STRING=$<TARGET_PROPERTY:nrf_security_target,MBEDTLS_CFG_FILE>

joerchan: I saw that, but I don't understand why it is needed. Can you explain?

joerchan: explained ofline: it is to support relative filename instead of full path.

frkv (approved):
LGTM

SebastianBoe (approved):

SebastianBoe:
we need a manifest PR to regression test this

joerchan:
> we need a manifest PR to regression test this

I'll make one together with the related Crypto changes for sdk-zephyr.

For the record this only affects builds that enables BL2, so it is not a functional change.

joerchan:
@SebastianBoe @frkv Updated the PR to remove the warning. Since it is not in used it still produces the warning.

frkv (approved):
Still LGTM, pending passing tests

SebastianBoe (approved):

joerchan:
@SebastianBoe Redid the change after your recent changes.

carlescufi (approved):

PR #590, opened 2021-11-22
doc: nrf_security: Different configs on secure and non-secure

SebastianBoe:
Documented how to achieve different configuration of nrf_security on
the non-secure and secure side.

### DIFF
--- a/nrf_security/doc/different_tfm_config.rst
+++ b/nrf_security/doc/different_tfm_config.rst
@@ -3,23 +3,20 @@
 Different configurations in SPE and NSPE
 ########################################
 
-Normally in a TF-M application with a NSPE image and a SPE image the
-NSPE will invoke the PSA Crypto APIs to do crypto. And the feature set
-of the PSA Crypto API will be configured through nrf_security. So
-there will be no crypto implementations in the NSPE.
+When a TF-M application has both an NSPE (Non-secure Processing Environment) image and an SPE (Secure Processing Environment) image, the NSPE usually invokes the PSA Cryptography APIs to perform cryptographic operations.
+In this case, you can configure the feature set of the PSA Cryptography API through the Nordic Security Module (nrf_security).
+As such, there are no cryptographic implementations in the NSPE.
 
-Therefore in a normal TF-M application there will be two instances of
-nrf_security, one in the NSPE that just contains an API for PSA
-Crypto. And one in the SPE which contains a cryptocell driver and
-possibly other backends as well. Yet they will both use the same
-configuration.
+Therefore, in a typical TF-M application, there are two instances of the Nordic Security Module:
 
-But, if an application needs to have some crypto features implemented
-in the NSPE nrf_security instance and other features implemented in
-the SPE nrf_security instance the following guide can be followed to
-configure each instance independently:
+* One instance in the NSPE that contains just an API for the PSA Cryptography.
+* One instance in the SPE that contains an ARM Cryptocell driver and possibly other backends.
 
-1. Build the application for a board with TF-M.
+Both instances would still use the same configuration.
+
+If you want to independently configure the two Nordic Security Module instances so that they can have different features, follow the steps below:
+
+1. Build the application for a development kit with TF-M:
 
    .. code-block:: console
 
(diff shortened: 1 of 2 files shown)


nrf_security/doc/different_tfm_config.rst
```diff
@@ -0,0 +1,62 @@
...
+      mkdir -p $APP_DIR/secure/zephyr
+      cp $APP_DIR/build/zephyr/include/generated/autoconf.h $APP_DIR/secure/zephyr/
+
+#. Modify the application build scripts such that secure nrf_security
+   instance uses these copied files to configure itself instead of
+   using the same configuration as the non-secure instance.
+
+   .. code-block:: console
+
+      cat ~/ncs/nrf/samples/crypto/nrf_security_snippet.cmake >> $APP_DIR/CMakeLists.txt
```
tejlmand, line 51: Did you forget to add `nrf_security_snippet.cmake` to this PR ?

Also I think referring to `~/ncs/` is not optimal.
If we should refer to any default path, then it should probably be the default install path proposed when installing using  Toolchain manager for installing nRF Connect SDK.

nrf_security/doc/different_tfm_config.rst
```diff
@@ -0,0 +1,62 @@
...
+
+#. Copy the resulting configuration from the build directory to the
+   application source directory. This copy will be used to configure
+   the secure nrf_security instance.
+
+   .. code-block:: console
+
+      cp $APP_DIR/build/zephyr/.config $APP_DIR/secure_zephyr.config
+      mkdir -p $APP_DIR/secure/zephyr
+      cp $APP_DIR/build/zephyr/include/generated/autoconf.h $APP_DIR/secure/zephyr/
```
tejlmand, line 43: Could we provide a Kconfig option for this, and then use a `add_custom_command(OUTPUT <path>/secure/zephyr/autoconf.h)` without having a `DEPENDS`.

That way `add_custom_command()` will only copy the file if it is missing, but not overwrite an already existing / modified file.

Will save the reader some work, and at the same time allow local modifications.

Having such setting could potentially also remove the need for doing the `cat ... >> CMakeLists.txt` below.

tejlmand (commented):
I like the direction.

Clean and simple, and at the same time flexible enough for the user.

But I think we could automate some of the user steps.

nrf_security/doc/different_tfm_config.rst
```diff
@@ -0,0 +1,62 @@
+.. _nrf_security_different_tfm_config:
+
+Different configurations of secure and non-secure
+#################################################
+
+Normally in a TF-M application the non-secure app will invoke the PSA
+Crypto API to do crypto. And the feature set of the PSA Crypto API
```
frkv, line 7: `PSA Crypto APIs`

nrf_security/doc/different_tfm_config.rst
```diff
@@ -0,0 +1,62 @@
...
+1. Build the application for a non-secure board with TF-M.
+
+   .. code-block:: console
+
+      export APP_DIR=~/ncs/nrf/samples/crypto/psa_tls
+      cd $APP_DIR
+      west build -b nrf9160dk_nrf9160_ns
+
+#. Enable nrf_security, if it is not enabled already, and configure
+   the crypto functionality and backends as they should be configured
```
frkv, line 31: `the crypto functionality and backends` => `crypto functionality and enabled backends`

Also "the secure side" is probably not a term that is very helpful. Alternatives to this are "SPE", "Secure Partition Environment" or "the secure image". I don't know if we are using a consistent term for this, but we should...

nrf_security/doc/different_tfm_config.rst
tejlmand: it's in nrf repo.

nrf_security/doc/different_tfm_config.rst
tejlmand: discussed offline.

Can be enhanced later.

tejlmand (approved):

SebastianBoe:
Waiting for review by @frkv

frkv (approved):
LGTM

SebastianBoe:
The new nrf_security supports diffferent configs for secure and non-secure so this is no longer needed.

PR #594, opened 2021-11-29
nrf_security: Remove explicit link to trusted-firmware-m include folder

joerchan:
Remove explicit link to trusted-firmware-m include folder. Instead
rely on the build system to make the headers available.

### DIFF
--- a/nrf_security/src/zephyr/CMakeLists.txt
+++ b/nrf_security/src/zephyr/CMakeLists.txt
@@ -97,10 +97,6 @@ if(CONFIG_BUILD_WITH_TFM)
     add_dependencies(${mbedcrypto_target} tfm_mbedtls_headers_copy)
   endif()
 
-  zephyr_include_directories(
-    ${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/trusted-firmware-m/interface/include
-  )
-
 endif()
 
 zephyr_link_libraries(mbedtls_external)


frkv (approved):
LGTM (assuming the nrf PR also checks out)

nrf_security/src/zephyr/CMakeLists.txt
```diff
@@ -97,10 +97,6 @@ if(CONFIG_BUILD_WITH_TFM)
     add_dependencies(${mbedcrypto_target} tfm_mbedtls_headers_copy)
   endif()
 
-  zephyr_include_directories(
-    ${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/trusted-firmware-m/interface/include
```
tejlmand, line 101: > Instead rely on the build system to make the headers available.

Good principle.
I wonder if you could refer to a merged commit or PR that has added this so we are guaranteed that removing this line will work.

SebastianBoe: CI will guarantee it, no?

joerchan: @tejlmand Yes, I just haven't linked them together yet because the dependency is too long, going all the way to upstream TF-M. When I take this out of draft it will have the reference to the required zephyr change.

tejlmand: > CI will guarantee it, no?

most likely, but that doesn't tell what and where something was changed that makes this change safe.

SebastianBoe: Ah, I thought you meant in a comment.

Git commit message update would be good.

joerchan:
This didn't work out, it caused header collisions with MBedTLS.

PR #596, opened 2021-11-30
softdevice_controller: mpsl: Making debugversion compile work

martintv:
Moving mpsl under softdevice controller,
without this we get linker errors when linking with un-obfuscated softdevice-controller and mpsl libraries.

linker errors like:
"undefined reference to `mpsl_cx_granted_ops_get`

Signed-off-by: Martin Tverdal <martin.tverdal@nordicsemi.no>

Manifest PR https://github.com/nrfconnect/sdk-nrf/pull/10345

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,11 +11,11 @@ if (CONFIG_NFC_T2T_NRFXLIB OR
 	add_subdirectory(nfc)
 endif()
 
-add_subdirectory_ifdef(CONFIG_MPSL                    mpsl)
 add_subdirectory_ifdef(CONFIG_NRF_MODEM               nrf_modem)
 add_subdirectory_ifdef(CONFIG_NRFXLIB_CRYPTO          crypto)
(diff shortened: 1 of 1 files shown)


seko-nordic (approved):

rugeGerritsen (approved):

tejlmand (changes requested):
> Moving mpsl under softdevice controller, because this somehow works better when building with debugversion

So exactly why is that the case ?
Without knowing why things behave but just reorder is a very bad sign.

Looking at this code:
https://github.com/nrfconnect/sdk-nrfxlib/blob/fc361892ba4fb71e645edc3cdd7b8057ad4b8421/mpsl/CMakeLists.txt#L16-L18

I wonder, could there be some link ordering or include path that makes thing compile or link slightly different because of other libraries in `nrfxlib`.

If this occurs only when `mpsl` is included before `softdevice_controller`, then maybe ther is some relation there we should be aware about, and describe correctly in CMake.

Feel free to make direct ping, if we should investigate further.

PR #601, opened 2021-12-02
openthread: create rcp libraries in separated directory

lmaciejonczyk:
RCP libraries do not depend on security backend therefore keep them in one separated directory.
Strip and copy libraries which are linked only.


lmaciejonczyk:
merged into other PR

PR #606, opened 2021-12-13
nrf_dm: Adding the nordic distancing lib

Tschet1:
The nrf_dm librarary is a proprietary distance measurement library that
allows to estimate the distance between two nordic devices that support
this feature.

Signed-off-by: Jan Müller <jan.mueller@nordicsemi.no>

Corresponding PR in nrf: https://github.com/nrfconnect/sdk-nrf/pull/6428


eriksandgren (approved):

nrf_dm/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+if(CONFIG_SOC_NRF52840)
+	set(NRF_DM_PATH ${CMAKE_CURRENT_SOURCE_DIR}/lib/pca10056/hard-float)
+elseif(CONFIG_SOC_NRF52833)
+	set(NRF_DM_PATH ${CMAKE_CURRENT_SOURCE_DIR}/lib/pca10100/hard-float)
+else()
+	message(FATAL_ERROR "Unsupported SoC.")
+endif()
```
rugeGerritsen, line 13: You should consider unifying the path names to other libraries. This will also make it more future proof.
See the usage of `nrfxlib_calculate_lib_path` for other libs.

grochu: Good point @rugeGerritsen. If you decide not to fully align the _nrf_dm_ paths at least please consider getting rid of the "_pca10xxx_" naming as Zephyr and NCS don't use these codes. E.g. Zephyr defines build target `nrf52dk_nrf52832` where `nrf52dk` refers to the name of the DK.

nrf_dm/Kconfig
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2020 Nordic Semiconductor
```
grochu, line 2: Maybe 2021 for consistency with other files?

nrf_dm/README.rst
```diff
@@ -0,0 +1,161 @@
...
+* The distance is computed based on all information available to the device.
+* All the mathematics necessary to compute distance is performed inside the library.
+
+Background
+==========
+
+Multi-carrier phase difference
+------------------------------
+
+Multi-carrier phase ranging for distance measurement is based on measuring the phase difference between two devices as a function of frequency.
```
grochu, line 28: Just a suggestion, does it sound better?
```suggestion
Multi-carrier phase ranging for distance measurement is based on measuring the phase difference in RF signals between two devices as a function of frequency.
```

nrf_dm/include/nrf_dm.h
```diff
@@ -0,0 +1,310 @@
+/*
+ * Copyright (c) Nordic Semiconductor ASA
```
grochu, line 2: Please add year

nrf_dm/include/nrf_dm.h
```diff
@@ -0,0 +1,310 @@
...
+		.ant_comp = NRF_DM_ANTENNA_COMP_1_1,                                               \
+		.dist_offset_err = { .ifft = 0.0f, .phase_slope = 0.0f, .rssi_openspace = 0.0f },  \
+	})
+
+/**
+ * @brief Initialize the distance measurement procedure
+ *
+ * @param[in] ppi_cfg PPI channels to be used by the library. The required amount of PPI channels
+ * is defined by the maximum hamming distance between two successive patterns in @ref
+ * switch_pattern divided by two On the nrf53 series, only 2 DPPI channels are used. A minimum of
```
grochu, line 257: Should it be:
```suggestion
 * switch_pattern divided by two. On the nrf53 series, only 2 DPPI channels are used. A minimum of
```

nrf_dm/license.txt
```diff
@@ -0,0 +1,5 @@
+/*
+ * Copyright (c) Nordic Semiconductor ASA
```
grochu, line 2: Please add year

nrf_dm/include/nrf_dm_version.h
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) Nordic Semiconductor ASA
```
grochu, line 2: Please add year

nrf_dm/README.rst
```diff
@@ -0,0 +1,161 @@
...
+Overview
+********
+
+This library provides the following key features:
+
+* Easy-to-use interface for distance measurement.
+* The distance is computed based on all information available to the device.
+* All the mathematics necessary to compute distance is performed inside the library.
+
+Background
```
grochu, line 22: I am not familiar with the details of the ranging methods, but I've heard on some meetings that each method has its advantages in specific conditions: e.g. RTT might be more accurate for higher distances (I am not sure it's true, just an example). If any hint or characterization like this can be formulated, could you provide it here? It would help understand how to practically use this library.

grochu (commented):
Added a few comments/suggestions.

maje-emb:
Misspelling: `nrf_dm/lib/nNF52840/hard-float/libnrf_dm.a`?

nrf_dm/Kconfig
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2021 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config NRF_DM
+	bool "nRF DM (Distance Measurement) library [EXPERIMENTAL]"
+	select FPU
```
maje-emb, line 9: Maybe?
```suggestion
	select FPU if CPU_HAS_FPU
```
To consider an SoC without an FPU

nrf_dm/include/nrf_dm.h
```diff
@@ -0,0 +1,311 @@
...
+#endif
+
+/**
+ * @brief Quality of the procedure
+ */
+typedef enum {
+	NRF_DM_QUALITY_OK = 0,
+	NRF_DM_QUALITY_POOR = 1,
+	NRF_DM_QUALITY_DO_NOT_USE = 2,
+	NRF_DM_QUALITY_CRC_FAIL = 3,
```
maje-emb, line 33: Consider adding structure field comments, for consistency.

maje-emb (commented):
Adding some comments.

nrf_dm/include/nrf_dm.h
Tschet1: do you have a suggestion for the information to add?

maje-emb: Meybe:
```
/* Good measurement quality */
/* Poor measurement quality */
/* Measurement not for use */
```
I added this suggestion because almost all the structure fields in this file have comments.

Tschet1: Is this not already visible in the enum names?

maje-emb: Yes this is visible. It's just a suggestion.

Tschet1: ok, I will leave this as is, for now.

nrf_dm/README.rst
Tschet1: We agree that such information could be useful. At this point, we don't have all the information needed to provide such a statement with enough confidence, however. We would probably add more details about how to use the different modes as we approach the production release.

grochu: Sounds good, thanks for clarifying.

grochu (approved):

FrancescoSer (approved):

nrf_dm/doc/nrf_dm_usage.rst
```diff
@@ -9,82 +9,23 @@ Usage
...
-         float best;           /**< Best effort distance estimate */
-         } mcpd;
-
-         struct
-         {
-         float rtt;            /**< RTT: Distance estimate based on RTT measurement */
-         } rtt;
-      } nrf_dm_dist_estimates_t;
+#. Call the :c:func:`nrf_dm_report_get` function to get the results.
+   It returns a pointer to the :c:struct:`nrf_dm_report_t` structure which contains the the  :c:struct:`nrf_dm_dist_estimates_t` structure with an estimation of the distance.
```
FrancescoSer, line 31: @Tschet1 Please fix the " the the  "

nrf_dm/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+#
+# Copyright (c) 2021 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+include(${NRFXLIB_DIR}/common.cmake)
```
tejlmand, line 7: Why is this included here, when it's already loaded here:
https://github.com/nrfconnect/sdk-nrfxlib/blob/9869bf438ef973c68aaeac80c3431abb83b7caaa/CMakeLists.txt#L6

```suggestion
```

tejlmand (changes requested):

FrancescoSer (approved):

Tschet1:
rebased on current main

tejlmand (changes requested):
why are the files: `libnrf_dm.a`
called libraries and with a `.a` extensions when they are not archive files but object files ?

I suggest to either add them directly as object files or create a proper archive containing just the single object file with proper name.

```
$ file libnrf_dm.a 
libnrf_dm.a: ELF 32-bit LSB relocatable, ARM, EABI5 version 1 (SYSV), not stripped
```

Tschet1:
> why are the files: `libnrf_dm.a` called libraries and with a `.a` extensions when they are not archive files but object files ?
> 
> I suggest to either add them directly as object files or create a proper archive containing just the single object file with proper name.
> 
> ```
> $ file libnrf_dm.a 
> libnrf_dm.a: ELF 32-bit LSB relocatable, ARM, EABI5 version 1 (SYSV), not stripped
> ```

good point. fixed

tejlmand:
> good point. fixed

Why is the extension of the file inside the archive having `.elf` ?
It is a relocatable object file, as one can see:
```
$ ar t nrf_dm/lib/nRF52840/hard-float/libnrf_dm.a
libnrf_dm.elf
$ ar x nrf_dm/lib/nRF52840/hard-float/libnrf_dm.a
$ file libnrf_dm.elf 
libnrf_dm.elf: ELF 32-bit LSB relocatable, ARM, EABI5 version 1 (SYSV), not stripped
```

and not an executable.
An executable has another signature, like this:
```
<name>: ELF 32-bit LSB executable, ARM, ....
```

so calling this file `.elf` is messing up terminology.

Tschet1:
> so calling this file `.elf` is messing up terminology.
I see. My mistake.

@rugeGerritsen : I see that we are using ".elf" also for MPSL/SDC. Maybe we can update that, as well.

rugeGerritsen:
> > so calling this file `.elf` is messing up terminology.
> > I see. My mistake.
> 
> @rugeGerritsen : I see that we are using ".elf" also for MPSL/SDC. Maybe we can update that, as well.

Yes we can

tejlmand (approved):

PR #612, opened 2021-12-17
nrf_802154: fix build warning for nrf5340 APP

lmaciejonczyk:
The missleading warning is displayed: "This combination of SoC
and floating point ABI is not supported by the nrf_802154_sl lib."
for the case where nrf_802154_sl lib is not linked (it happens for
target: nrf5340dk_nrf5340_cpuapp when the radio serialization is
used instead).
This commit fixes it by exluding "sl" directory when the option
CONFIG_NRF_802154_RADIO_DRIVER is not active.


LuDuda (approved):

lmaciejonczyk:
@czeslawmakarski you review is needed to proceed, could you take a look?

jciupis:
@ankuns could you please have a look?

ankuns:
Causes trouble when applied to internal repo, looking at this. Marking as DNM until solution is found.

ankuns (approved):

MarekPorwisz (approved):

PR #620, opened 2022-01-17
mpsl: update mpsl_cx_register_callback description

piotrslawecki:
Updated the description of mpsl_cx_register_callback to match
the behavior as per https://github.com/nrfconnect/sdk-nrf/pull/6219

Signed-off-by: Piotr Sławęcki <piotr.slawecki@nordicsemi.no>


hubertmis (changes requested):
I think this patch should be propagated from origin MPSL repository while rendering MPSL to sdk-nrfxlib. It also requires an update in mpsl_cx_none.c implementation compiled in MPSL libraries.

PR #634, opened 2022-01-28
nrf_modem: v1.5.0

lemrey:
* The AT socket has been removed.
* The DFU socket has been removed.
* Added support for deferred processing of modem traces.
  Introduced the nrf_modem_trace_processed_callback() function
  that the application must call after it has processed a trace
  received in nrf_modem_os_trace_put.
* It is now possible to unset the AT notification handler
  by passing NULL to nrf_modem_at_notif_handler_set()
* The number of required semaphores is now exported in nrf_modem_os.h
* Fixed a bug where nrf_getsockopt() would not truncate the socket
  option as intended when the buffer provided was too small.
* Fixed a bug where closing a socket while another thread was in
  a recv() operation on the same socket would result in a crash.
* Fixed a bug in the delta DFU interface where the
  nrf_modem_delta_dfu_offset() call could return
  an unexpected error code in some cases.


nrf_modem/doc/ug_nrf_modem_porting_os.rst
```diff
@@ -134,10 +134,10 @@ nrf_modem_os_trace_put
 
 This function puts the trace string to the desired medium, typically UART.
 However, the medium used to forward and store the traces is up to the implementation and must be initialized correctly before using.
-Once the traces are processed/stored, the :c:func:`nrf_modem_trace_processed_callback` must be called.
-If you are not interested in traces you only need to ensure that :c:func:`nrf_modem_trace_processed_callback` is called for each received trace.
-Until the :c:func:`nrf_modem_trace_processed_callback` is called, the Modem library will not free up the memory allocated for that trace in the trace memory area.
-Since the modem uses this trace memory area to send traces, not calling :c:func:`nrf_modem_trace_processed_callback` may lead to lost modem traces.
+Once the traces are processed or stored, the :c:func:`nrf_modem_trace_processed_callback` must be called.
+If you want the traces, you need to ensure that :c:func:`nrf_modem_trace_processed_callback` is called for each received trace.
```
lemrey, line 138: We should rephrase this, it lost its meaning like this.
The idea is: regardless of whether the application is interested in processing the trace data, the application must call `nrf_modem_trace_processed_callback` for each received trace.

divipillai: fixed.

nrf_modem/include/nrf_modem.h
```diff
@@ -171,9 +171,27 @@ void nrf_modem_application_irq_handler(void);
...
+ * pointed to by @p buf. It is the application's responsibility to call this function with
+ * the same parameter values as received in the @ref nrf_modem_os_trace_put function.
+ * Calling this function with incorrect values leads to undefined behavior.
+ *
+ * @param buf Pointer to the memory buffer as received in @ref nrf_modem_os_trace_put
+ * @param len Length of memory buffer as received in @ref nrf_modem_os_trace_put
+ * @retval Zero on success.
+ * @retval -NRF_EINVAL @p buf is @c NULL or an invalid trace buffer
+ * @retval -NRF_EINVAL @p len is too large to be a valid trace length
+ * @retval -NRF_EAGAIN Resource temporarily unavailable. Try again with same parameters.
```
rlubos, line 191: Try again when, is there some indicator of when to retry?

lemrey: @balaji-nordic

balaji-nordic: There is no indicator. It should be treated similar to a `k_sem_take()` failure with `-EAGAIN` return.

divipillai (approved):
Reviewed and updated the document and figures as per the requirements.

rlubos (approved):

anhmolt (approved):

nrf_modem/doc/CHANGELOG.rst
```diff
@@ -9,6 +9,19 @@ Changelog
...
+nrf_modem 1.5.0
+***************
+
+* Added support for deferred processing of modem traces.
+  Introduced the :c:func:`nrf_modem_trace_processed_callback` function that the application must call after it has processed a trace received in :c:func:`nrf_modem_os_trace_put`.
+* It is now possible to unset the AT notification handler by passing NULL to :c:func:`nrf_modem_at_notif_handler_set`.
+* The number of required semaphores is now exported in :file:`nrf_modem_os.h`.
+* Removed the AT socket.
+* Removed the DFU socket.
+* Fixed a bug where :c:func:`nrf_getsockopt` do not truncate the socket option as intended when the buffer provided was too small.
```
MirkoCovizzi, line 21: Nit: should be "did" instead of "do"

PR #645, opened 2022-02-03
nrf_security: HMAC_DRBG fixes for TF-M minimal build

joerchan:
-Adds usage of nrf_cc3xx_platform_hmac_drbg APIs to ensure
 smaller TF-M image
-Disable unneccesary key derivation for TFM_MINIMAL
-Minor fixup using CONFIG_BUILD_WITH_TFM for random_ext.c addition

ref: NCSDK-13530


nrf_security/src/random_ext.c
```diff
@@ -7,18 +7,30 @@
...
         output_size,
         output_length);
+#elif defined(PSA_WANT_ALG_HMAC_DRBG)
+    ret = nrf_cc3xx_platform_hmac_drbg_get(
+        NULL,
+        output,
+        output_size,
+        output_length);
+#else
+    #error random_ext: Enable CONFIG_PSA_WANT_ALG_CTR_DRBG or CONFIG_PSA_WANT_ALG_HMAC_DRBG
```
SebastianBoe, line 32: No " ?

And please double-check that indenting preprocessor directives is OK with the coding standard.

frkv: The indenting in nrf_security is according to mbed TLS and TF-M coding standard

frkv: And missing quotes will be fixed

joerchan: Fixed the quotes

tejlmand: > The indenting in nrf_security is according to mbed TLS and TF-M coding standard

but this file is not copied from mbedTLS but a pure NCS file, from what I can tell.
In which I believe NCS / Zephyr coding guidelines should be followed.

tejlmand (approved):
lgtm.

Placed a comment regarding the coding guideline, but not blocking in this particular case, imho.

nrf_security/src/random_ext.c
frkv: We don't differentiate the coding standard inside nrf_security. All the changes and adaptions are on mbed TLS code and/or code that works together with mbed TLS

nrf_security/src/random_ext.c
```diff
@@ -7,18 +7,31 @@
...
+        output_length);
+#elif defined(CONFIG_PSA_WANT_ALG_HMAC_DRBG)
+    ret = nrf_cc3xx_platform_hmac_drbg_get(
         NULL,
         output,
         output_size,
         output_length);
+#else
+    #error "Enable CONFIG_PSA_WANT_ALG_CTR_DRBG or CONFIG_PSA_WANT_ALG_HMAC_DRBG"
+#endif
```
SebastianBoe, line 34: random_ext.c shouldn't need to know that nrf_cc3xx_platform_ctr_drbg_get is invoked when CONFIG_PSA_WANT_ALG_CTR_DRBG is enabled.

We should have a dedicated function that knows which get function to call based on PSA defines and this function should be invoked from here.

The same applies to the init functions.

SebastianBoe (changes requested):
changes requested

nrf_security/src/random_ext.c
joerchan: Discussed offline and we will leave this as is for now.

frkv (approved):
LGTM

And is following Mbed TLS coding standard as all other files in nrf_security

frkv (approved):
Still LGTM (with additional changes)

PR #660, opened 2022-02-15
crypto: Remove deprecated code

SebastianBoe:
Remove deprecated code that used to be used to ease migration. By this
point in time users should have migrated off of the nrf_cc310_* APIs.


tejlmand (approved):
always nice to see some cleanup :+1:

frkv (changes requested):
The files here are generated by a build of the runtime libraries and can be added as a task for a new version of the runtime-library. 

Rejecting this cleanup done directly in the code as it doesn't follow a runtime release

Added a task for it for the next runtime release: NCSDK-13917

PR #661, opened 2022-02-15
Remove duplicated code

SebastianBoe:


tejlmand (approved):

SebastianBoe:
The patches are touching "generated" (copied) files. This is not source code. The source code is elsewhere.

Closing.

Will send a new patch that fixes the actual source code.

PR #666, opened 2022-02-18
Multiple same-type endpoints conflict resolved

psychogenic:
Hi,
The Zboss implementation used with NCS won't let us create more than one endpoint of the same type.  For example, I can have a dimmable light and an outlet, but not two outlets.

This is because of the way ZB_DECLARE_SIMPLE_DESC and the resulting type depends only on the numbers of in/out clusters, and doesn't worry about if such a combination may have existed before.  

I detailed the problem and this fix at
https://devzone.nordicsemi.com/f/nordic-q-a/84787/zigbee-multiple-same-type-endpoints-conflict-and-proposed-fix
It is for the most part completely transparent to ZCL users and allows me to have a device with 2 outlets and 4 binary inputs without issue.  It'd be nice if this was the default.

### DIFF
--- a/zboss/development/include/ha/zb_ha_combined_interface.h
+++ b/zboss/development/include/ha/zb_ha_combined_interface.h
@@ -149,8 +149,8 @@
     @param out_clust_num - number of supported output clusters
 */
 #define ZB_ZCL_DECLARE_COMBINED_INTERFACE_SIMPLE_DESC(ep_name, ep_id, in_clust_num, out_clust_num) \
-  ZB_DECLARE_SIMPLE_DESC(in_clust_num, out_clust_num);                                             \
-  ZB_AF_SIMPLE_DESC_TYPE(in_clust_num, out_clust_num) simple_desc_##ep_name =                      \
+  ZB_DECLARE_SIMPLE_DESC(ep_name,in_clust_num, out_clust_num);                                             \
+  ZB_AF_SIMPLE_DESC_TYPE(ep_name, in_clust_num, out_clust_num) simple_desc_##ep_name =                      \
   {                                                                                                \
     ep_id,                                                                                         \
     ZB_AF_HA_PROFILE_ID,                                                                           \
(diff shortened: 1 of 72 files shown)


psychogenic:
Sidenote: I signed the CLA but it's still saying I haven't (even after "recheck")--if there's something else I must do, lemme know.

mike-scofield:
I had the same problem.  My solution was far less elegant:

```
ZB_HA_DECLARE_ON_OFF_OUTPUT_EP(
        relay_0_on_off_ep,
        EP_RELAY_0,
        relay_0_on_off_clusters);

// HACK: defining two of the same endpoints causes a redefinition
#undef ZB_DECLARE_SIMPLE_DESC
#define ZB_DECLARE_SIMPLE_DESC(...)

ZB_HA_DECLARE_ON_OFF_OUTPUT_EP(
        relay_1_on_off_ep,
        EP_RELAY_1,
        relay_1_on_off_clusters);
```

Seems like it would be a fairly common use case so it would be nice to get this fixed upstream.

bmangodt:
Hi @psychogenic!

I am a complete noob at this zigbee stuff and I need to make the multiple same-type endpoints work for a PoC here at work. This is the first time we have done anything like that and our dev team are scratching their heads on it.
I pulled your fix from your repo, put the files in ~/ncs/v2.0.0/nrfxlib but when I build the project in VSCode with the nrfConnect plugin, I still get build errors about implicit declarations/undeclared variables that were previously not an issue. Is there a rebuilt I need to do of something? And if so, what are the steps to successfully build it?

Thanks in advance and with kind regards,
Beau

psychogenic:
Hello @bmangodt,  
I submitted this patch half a year ago, and didn't get any love.  It seems things have changed while they didn't merge, and now it looks borken.  I have to update my nRF SDK and, unless they actually got this pretty obvious use case working in the code, will have to find a way to re-patch.  This should be in the coming weeks and I'll update you should I get it working.  Would appreciate if you did same.  Cheers,
PatD

bmangodt:
> Hello @bmangodt, I submitted this patch half a year ago, and didn't get any love. It seems things have changed while they didn't merge, and now it looks borken. I have to update my nRF SDK and, unless they actually got this pretty obvious use case working in the code, will have to find a way to re-patch. This should be in the coming weeks and I'll update you should I get it working. Would appreciate if you did same. Cheers, PatD

Hi @psychogenic,
Thanks for your quick reply. I don't understand why your fix wasn't added a long time ago. From what I can see using the latest released nRF SDK, your fix or any derivative of it for this use case has been added. What a shame!
I will have a look at fixing the compile errors with my limited knowledge of the SDK and see what I can do. I'll keep you posted.
Thanks anyway for your work trying to fix this obvious issue.

Kind regards,
Beau

martelmy:
Hi @bmangodt and @psychogenic,
When moving to ZCL8 the HA device definitions were removed due to incompatibilities, so most of the files under /include/ha/ do not exist anymore in v2.0.0. See the [nrfxlib ZBOSS changelog](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/2.0.2/nrfxlib/zboss/CHANGELOG.html#removed) for more info.
Instead, these files are added to the specific samples and applications. There are no issues with the other files in your fix, only the HA device files, so just removing these files from /include/ha/ and adding the changes for multiple same-type endpoints to the HA device file for the sample/application instead, e.g. https://github.com/nrfconnect/sdk-nrf/blob/v2.0.0/samples/zigbee/light_bulb/include/zb_dimmable_light.h, fixes the issue. 
Again thanks for the fix, @psychogenic. I will check with the development team again to see if they will apply it.

PR #672, opened 2022-02-22
Upmerge 22-03-03

gmarull:
The new Kconfig extension requires usage of :kconfig:option: role, since
it uses a language domain.


leewkb4567 (approved):
Gazell OK.

tejlmand (approved):
lgtm

PR #678, opened 2022-03-07
nrf_modem: Make the irq priorities configurable via KConfig

SvenGrandcentrix:
When zero-latency-interrupts are enabled, the current values for the interrupt priority are no longer valid. I introduced the interrupt priorities as parameter to the KConfig. The defaut values are selected so that applications start without an assert from the irq system.  

Signed-off-by: Sven Hädrich <sven.haedrich@grandcentrix.net>

### DIFF
--- a/nrf_modem/Kconfig
+++ b/nrf_modem/Kconfig
@@ -11,6 +11,14 @@ config NRF_MODEM_LINK_BINARY
 
 endchoice
 
+config NRF_MODEM_NETWORK_IRQ_PRIORITY
+	int "Modem network layer communication interrupt priority"
+	default 1
+
+config NRF_MODEM_APPLICATION_IRQ_PRIORITY
+	int "Modem application layer communication interrupt priority"
+	default 4
+
 config NRF_MODEM_LOG
 	bool "Enable library logging"
 
--- a/nrf_modem/include/nrf_modem_platform.h
+++ b/nrf_modem/include/nrf_modem_platform.h
@@ -35,13 +35,13 @@ extern "C" {
 #define NRF_MODEM_NETWORK_IRQ IPC_IRQn
 
 /**@brief Interrupt priority used on interrupt for communication with the network layer. */
-#define NRF_MODEM_NETWORK_IRQ_PRIORITY 0
+#define NRF_MODEM_NETWORK_IRQ_PRIORITY CONFIG_NRF_MODEM_NETWORK_IRQ_PRIORITY
 
 /**@brief Interrupt used for communication with the application layer. */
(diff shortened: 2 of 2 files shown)


rlubos:
The commit message body has also a line length limit (72 char/line).

@lemrey Can you comment if it's ok to set those new IRQ priorities in general?

SvenGrandcentrix:
fixed line break of commit message

rlubos:
Still too long (from checkpatch resutls):
> Line exceeds max length (92>72):

SvenGrandcentrix:
re-formated the commit message again

lemrey:
We can't have KConfig-generate values in exported library headers, because it should be possible to use the library without NCS.

rlubos:
> We can't have KConfig-generate values in exported library headers, because it should be possible to use the library without NCS.

Good point. Maybe we could wrap the values here witihn `ifndef NRF_MODEM_APPLICATION_IRQ_PRIORITY` and override at `sdk-nrf` level? @tejlmand Any ideas how to approach the issue?

lemrey:
I don't think these values should be exported to begin with, to be honest.
The IPC IRQ prio is fed into the library during initialization and the EGU IRQ prio belongs entirely to the application.
They belong to the integration layer, as KConfig, and should default to values fetched from DT.

rlubos:
> I don't think these values should be exported to begin with, to be honest. The IPC IRQ prio is fed into the library during initialization and the EGU IRQ prio belongs entirely to the application. They belong to the integration layer, as KConfig, and should default to values fetched from DT.

Ok, that sounds like a larger refactoring though, is this on schedule for libmodem?

lemrey:
Yes, we have scheduled some work in this direction.

SvenGrandcentrix:
Still I would like to combine nrf_modem with ZLI. I see two options:
1. modify IRQ priorities regardless of ZLI usage
2. change `nrf_modem_platform.h` into something like:
```
**@brief Interrupt priority used on interrupt for communication with the network layer. */
#ifdef CCONFIG_ZERO_LATENCY_IRQS
#define NRF_MODEM_NETWORK_IRQ_PRIORITY 0
#else
#define NRF_MODEM_NETWORK_IRQ_PRIORITY 1
#endif
```
What is your opinion?

lemrey:
> What is your opinion?

There is a third option, the IRQ priority becomes a user-configurable parameter and the integration layer changes the defaults automatically in case ZLI are enabled.

lemrey:
Modem library only uses the IPC IRQ now. The priority is configurable via DTS or Kconfig.

PR #686, opened 2022-03-15
adding .github/test-spec.yml

DatGizmo:
Signed-off-by: Sebastian Wezel <sebastian.wezel@nordicsemi.no>

This PR is meant to be a starting point to create the .github/test-spec.yml
Which will work similar to the [sdk-nrf/.github/test-spec.yml](https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml)

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -1,0 +1,25 @@
+# This is the Jenkins ci variant of the .github/labler.yaml
+
+CI-ble-test:
+  - 'softdevice_controller/**/*'
+  - 'mpsl/**/*'
+
+CI-zigbee-test:
+  - 'mpsl/**/*'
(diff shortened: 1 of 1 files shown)


.github/test-spec.yml
```diff
@@ -0,0 +1,25 @@
+# This is the Jenkins ci variant of the .github/labler.yaml
+
+CI-ble-test:
+  - 'softdevice_controller/**/*'
```
guwa, line 4: Can we update the path to:
'softdevice_controller/include/**/*'
'softdevice_controller/lib/**/*'

The doc update don't need to trigger the downstream CI

Same for the path under other test labels

guwa: Sure

PR #703, opened 2022-03-30
nrf_security: Fix TF-M FPU flags and TF-M relink dependency

joerchan:
Add TF-M compile flags required when enabling FPU in TF-M.
TF-M documentation on FPU support:
"Secure and non-secure libraries are compiled with COMPILER_CP_FLAG and
linked with LINKER_CP_OPTION for different FP ABI types. All those
libraries shall be built with COMPLIER_CP_FLAG"

NCSDK-14513

Revert "crypto: Add workaround for liboberon PSA API prefix

The workaround is no longer needed since the Oberon library no
longer calls psa_generate_random.

The workaround had some cmake dependency issues and the command
was always run, causing dependent targets to be re-run.

This reverts commit https://github.com/joerchan/nrfxlib/commit/be4f031dcf7617e6b18a49a58a18fd37221e4671.

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -77,16 +77,16 @@ if (CONFIG_NRF_OBERON OR CONFIG_BUILD_WITH_TFM OR CONFIG_OBERON_BACKEND OR CONFI
       # Generate a new library with this symbol redefined.
       set(OBERON_MBEDCRYPTO__PSA_LIB ${CMAKE_CURRENT_BINARY_DIR}/liboberon_mbedcrypto__psa_${OBERON_VER}.a)
 
-      add_custom_target(mbedcrypto_oberon_mbedcrypto__psa_rename
+      add_custom_command(
+        OUTPUT ${OBERON_MBEDCRYPTO__PSA_LIB}
         COMMAND ${CMAKE_OBJCOPY}
-        "${OBERON_PSA_LIB}"
-        "${OBERON_MBEDCRYPTO__PSA_LIB}"
-        --redefine-sym psa_generate_random=mbedcrypto__psa_generate_random
-        BYPRODUCTS
-        "${OBERON_MBEDCRYPTO__PSA_LIB}"
+          ${OBERON_PSA_LIB}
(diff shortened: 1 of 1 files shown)


frkv (approved):
LGTM

frkv (approved):
LEBTM (Looks Even Better To Me)

frkv (commented):
The addition of "mbedcrypto__" in the naming seems a bit strange to me

I would rather name the renamed library to include the terms "renames" and/or TF-M (or tfm or tf-m) to state that it is manipulated (and to tell by whom)

joerchan:
> The addition of "mbedcrypto__" in the naming seems a bit strange to me
> 
> I would rather name the renamed library to include the terms "renames" and/or TF-M (or tfm or tf-m) to state that it is manipulated (and to tell by whom)

That is the rename that is done though. Rename all `psa_` APIs to `mbedcrypto__psa`, since that is how TF-M wants to build mbedcrypto.

frkv:
> > The addition of "mbedcrypto__" in the naming seems a bit strange to me
> > I would rather name the renamed library to include the terms "renames" and/or TF-M (or tfm or tf-m) to state that it is manipulated (and to tell by whom)
> 
> That is the rename that is done though. Rename all `psa_` APIs to `mbedcrypto__psa`, since that is how TF-M wants to build mbedcrypto.

Yes. but this is a renamed library. `mbedcrypto` isn't a very helpful term as it is fairly under-expressed what it really means. My point is that a library that is named  `liboberon_psa_renamed_tfm` would be easier to locate. The symbol prefix scheme in TF-M is just a "magic prefix" that happens to reuse the mbedcrypto term. There are multiple term they could have used for prefixed renaming of APIs, like for example  "tfm_"...

joerchan:
@frkv The name doesn't really have anything to do with the change in the PR though. I'm just fixing the dependency here, the name is the same as before.
This is a temporary workaround and should be removed once oberon no longer calls psa_generate_random anymore, is this important enough to warrant changing the PR?

Vge0rge (approved):
Looks good!

PR #715, opened 2022-04-19
nrf_security: Add compile flags when built in TF-M with FPU enabled

joerchan:
Add TF-M compile flags required when enabling FPU in TF-M.
TF-M documentation on FPU support:
"Secure and non-secure libraries are compiled with COMPILER_CP_FLAG and
linked with LINKER_CP_OPTION for different FP ABI types. All those
libraries shall be built with COMPLIER_CP_FLAG"

NCSDK-14513


joerchan:
This commit was part of https://github.com/nrfconnect/sdk-nrfxlib/pull/703 but was dropped by accident during a rebase.

magnev (approved):

PR #716, opened 2022-04-21
KRKNWK-13402: Align ZBOSS addons with ZCL8

tomchy:


tomchy:
Pulled into: https://github.com/nrfconnect/sdk-nrfxlib/pull/719

PR #731, opened 2022-05-10
nrf_security: Remove setting of crypto partition heap to mbedtls config

joerchan:
Remove crypto partition heap size set based on the mbedtls heap size
configuration.

### DIFF
--- a/nrf_security/cmake/config_to_tf-m.cmake
+++ b/nrf_security/cmake/config_to_tf-m.cmake
@@ -34,12 +34,3 @@ if(CONFIG_TFM_BL2)
       -DMCUBOOT_MBEDCRYPTO_CONFIG_FILEPATH:STRING=${CONFIG_MBEDTLS_CFG_FILE}
   )
 endif()
-
-if(CONFIG_MBEDTLS_ENABLE_HEAP)
-  # The ENGINE_BUF_SIZE holds the dynamic allocation buffer for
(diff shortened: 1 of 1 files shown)


Vge0rge:
What is the intention for this? The reason that we included this was that it seems more natural compared to ignoring the mbedtls heap size kconfig and having to set the cypto_engine_buf_size in the cmake code manually.

joerchan:
> What is the intention for this? The reason that we included this was that it seems more natural compared to ignoring the mbedtls heap size kconfig and having to set the cypto_engine_buf_size in the cmake code manually.

This PR matches the changes here:
https://github.com/nrfconnect/sdk-nrf/pull/7511/commits/3f7fefda232c1d2e72a05f046fc6a87aa291182c

The intention is to allow the TF-M crypto partition heap size to be set independently of the application mbedtls heap size.
I was getting build errors in TF-M with TFM_MINIMAL enabled because of SRAM usage. Even though the crypto partition in this case is set to only have `psa_generate_random` available.

frkv (approved):
LGTM

The configuration is moved to sdk-nrf instead, where we can use PSA_WANT_ALG_XXXX to tune he heap requirements (if needed)

Vge0rge:
Ok, I didn't notice that we exposed this configuration in sdk-nrf. Then that looks good to me as well!

Vge0rge (approved):

PR #737, opened 2022-05-12
nrf_security: Add EITS config for TFM

Vge0rge:
-This adds an option to enable EITS in TFM (nrf9160/nrf5340) targets
    
 Ref: NCSDK-14747
   

crypto/nrf_cc312_platform/include/nrf_cc3xx_platform.h
```diff
@@ -90,6 +90,24 @@ bool nrf_cc3xx_platform_rng_is_initialized(void);
...
 void CRYPTOCELL_IRQHandler(void);
 
+
+/** @brief Function to get the nonce seed used for encrypted ITS usage
+ *
+ * @param[in]	buffer 	Buffer to fill the nonce seed generated during boot.
+ *
+ * @return Zero on success, otherwise a non-zero error code.
+ */
+int nrf_cc3xx_platform_get_nonce_seed(uint8_t buffer[8]);
```
oyvindronningstad, line 100: The buffer size here (and for `nrf_cc3xx_plaform_get_boot_seed()`) should have a define IMO, so it's easy to create correctly-sized arrays.

Vge0rge: This will need changes in the runtime library so I will create a task for the next version. The 0.9.14 is now merged.

nrf_security/Kconfig.psa
```diff
@@ -78,6 +78,14 @@ config MBEDTLS_USE_PSA_CRYPTO
 	help
 	  Corresponds to MBEDTLS_USE_PSA_CRYPTO setting in mbed TLS config file.
 
+config PSA_EITS_BACKEND_TFM
```
frkv, line 81: The naming "BACKEND" is unfortunate as the term is used for PSA_NATIVE_ITS and this configurations is not grouped within. 

I propose maybe PSA_EITS or PSA_ITS_ENCRYPTED to normalize this. If we state PSA_EITS/PSA_ITS_ENCRYPTED we can potentially reuse this for non TF-M usage as well.

crypto/nrf_cc312_platform/include/nrf_cc3xx_platform.h
frkv: Agreed, this was an oversight on my utilizing a magic number. Will be rectified in the next version of the runtime library

nrf_security/Kconfig.psa
Vge0rge: Adding something as generic as PSA_ITS_ENCRYPTED/PSA_EITS  when we have the NATIVE_ITS I think that is more confusing to the users. Adding a generic PSA_ITS_ENCRYPTED/PSA_EITS as a top level for every use case and independent use cases for backends makes more sense to me. What do you think?

frkv: The TFM config for this is currently TFM_ITS_ENCRYPT, which I'm proposing is changed to TFM_ITS_ENCRYPTED. This would in my opinion not be difficult to understand. It is also possible to make dependent defaults:

`config XXXX_ITS_ENCRYPTED
    default Y if YYYY_IS_ENCRYPTED`

To keep the internal consistency between PSA and TFM configurations. 

The term backend is a legacy term that we are gravitating away from because of the addition of the term "driver" in PSA APIs.

Vge0rge: That's a valid point, this looks better in the Kconfigs for sure. I will update this.

nrf_security/Kconfig.psa
```diff
@@ -78,6 +78,14 @@ config MBEDTLS_USE_PSA_CRYPTO
 	help
 	  Corresponds to MBEDTLS_USE_PSA_CRYPTO setting in mbed TLS config file.
 
+config PSA_ITS_ENCRYPTED
```
oyvindronningstad, line 81: Curious: Why isn't this a TFM_ option?

Vge0rge: Mainly because we will support EITS outside of TF-M later. We already have some PRs with the EITS being implemented by the zephyr settings backend. This is intended to be used as a high level Kconfig option which will remove the requirement of TF-M when EITS based on zephyr is merged.

nrf_security/cmake/config_to_tf-m.cmake
```diff
@@ -34,3 +34,10 @@ if(CONFIG_TFM_BL2)
       -DMCUBOOT_MBEDCRYPTO_CONFIG_FILEPATH:STRING=${CONFIG_MBEDTLS_CFG_FILE}
   )
 endif()
+
+if(CONFIG_PSA_ITS_ENCRYPTED)
```
oyvindronningstad, line 38: Maybe there should be a separate TFM_ option for doing this, for easier discoverability. That option could also select the PSA_ITS_ENCRYPTED option.

Vge0rge: Let me see if I get it straight. You propose to have TFM_ITS_ENCRYPTED which selects PSA_ITS_ENCRYPTED?

oyvindronningstad: Yes, so that users are able to find this config among the TFM configs.

Vge0rge: Isn't it better to have the other way around? Meaning to have the PSA_ITS_ENCRYPTED select the TFM_ITS_ENCRYPTED? Otherwise  platforms will have to actively enable a different configurations based on what they support if I am not mistaken.

frkv: I think that TFM_ITS_ENCRYPTED is the standardized config and PSA_ITS_ENCRYPTED is our convenience for handling non-TF-M

I would suggest TFM_ITS_ENCRYPTED selecting PSA_ITS_ENCRYPTED

Vge0rge: Should be done now ;)

frkv (approved):
LGTM

oyvindronningstad (approved):

greg-fer:
Adding @b-gent , as I have no clue why this has doc-required...

PR #746, opened 2022-05-23
nrf_modem: Use IRQ_PRIO_LOWEST as NRF_MODEM_APPLICATION_IRQ_PRIORITY

anangl:
... instead of the hard-coded value of 6 and this way allow successful
compilation also with zero-latency IRQs enabled (6 is not a valid
priority to be used in IRQ_CONNECT() then and a build assertion fails
when it is used).

Signed-off-by: Andrzej Głąbek <andrzej.glabek@nordicsemi.no>

### DIFF
--- a/nrf_modem/include/nrf_modem_platform.h
+++ b/nrf_modem/include/nrf_modem_platform.h
@@ -40,7 +40,7 @@ extern "C" {
 #define NRF_MODEM_APPLICATION_IRQ EGU1_IRQn
 
 /**@brief Interrupt priority used on interrupt for communication with the application layer. */
-#define NRF_MODEM_APPLICATION_IRQ_PRIORITY 6
+#define NRF_MODEM_APPLICATION_IRQ_PRIORITY IRQ_PRIO_LOWEST
 
 /**@} */
 


anangl:
> What build is failing?

The problem was reported here: https://devzone.nordicsemi.com/f/nordic-q-a/87533/regression-in-irq_direct_connect-behavior-from-ncs-1-7-to-1-9-1.

lemrey (changes requested):
This would be changing the priority of the EGU interrupt used by all nRF9160 applications; I do not think we should change it now. I suggest this be a known issue instead.

anangl:
> See, #678. I understand `IRQ_PRIO_LOWEST` is defined in `nrfx` which would be okay, but I am not confident in this change at this point during the release.

It is not defined in nrfx, it is defined in Zephyr, see [exc.h](https://github.com/nrfconnect/sdk-zephyr/blob/b821eb680689b7b05b756d1df4744f04ec512c80/include/zephyr/arch/arm/aarch32/exc.h#L54).

> This would be changing the priority of the EGU interrupt used by all nRF9160 applications; I do not think we should change it now. I suggest this be a known issue instead.

No. When ZLIs are not enabled, this macro will evaluate to 6, so nothing will actually change in this case. The difference is that with this change it will be possible to successfully compile `nrf_modem_os.c` in applications where `CONFIG_ZERO_LATENCY_IRQS=y` (without it, [this](https://github.com/nrfconnect/sdk-zephyr/blob/b821eb680689b7b05b756d1df4744f04ec512c80/include/zephyr/arch/arm/aarch32/irq.h#L96-L99) build assertion will fail [here](https://github.com/nrfconnect/sdk-zephyr/blob/b821eb680689b7b05b756d1df4744f04ec512c80/include/zephyr/arch/arm/aarch32/irq.h#L127) for [this](https://github.com/nrfconnect/sdk-nrf/blob/d1d8457cd9ef719725e28efba9da5107599528e2/lib/nrf_modem_lib/nrf_modem_os.c#L363-L365) call to `IRQ_DIRECT_CONNECT`).

lemrey:
Ok, then it's a problem because you are changing a file exported by this library to include a macro that is defined in Zephyr, but this library is not tied to Zephyr and it is supposed to be able to run on a different OS or bare metal.

lemrey:
Perhaps we could follow my suggestion in the other PR and let the application configure the priority of that interrupt in `nrf_modem_os.c` based on whether ZLI are enabled.

anangl:
> Ok, then it's a problem because you are changing a file exported by this library to include a macro that is defined in Zephyr, but this library is not tied to Zephyr and it is supposed to be able to run on a different OS or bare metal.

Oh, now I see. Then the change I proposed does not make sense, indeed.

But I don't really understand that "exporting" of this value by the library. It seems that the priority is something that the application should configure according to its needs/circumstances. What's the point then in defining it in a file that comes from the library? Does the library at all use or need to know this value? Am I missing something again?

anangl:
It's not a right approach to solve the problem.

lemrey:
> But I don't really understand that "exporting" of this value by the library. It seems that the priority is something that the application should configure according to its needs/circumstances. What's the point then in defining it in a file that comes from the library? Does the library at all use or need to know this value? Am I missing something again?

No, I agree with you 100% in fact we are working toward removing this :) just not for this release yet.

PR #757, opened 2022-05-27
nrf_security: Add definition of appmem partition for mbedtls library

anangl:
... to make it possible to build with `CONFIG_USERSPACE` enabled
(there is such definition in userspace.c in Zephyr but it is compiled
only when `CONFIG_MBEDTLS` is enabled, what is not the case here).

Ref.: NCSDK-15305


carlescufi (approved):

Vge0rge (approved):
LGTM

frkv (approved):
lgtm

PR #766, opened 2022-06-10
mpsl: Kconfig: Remove Kconfig from nrfxlib

rugeGerritsen:
It is added to sdk-nrf in
https://github.com/nrfconnect/sdk-nrf/pull/7939

Signed-off-by: Rubin Gerritsen <rubin.gerritsen@nordicsemi.no>


rugeGerritsen:
Added DNM as it requires the PR to sdk-nrf to be merged first

rugeGerritsen:
Rebased on main

tejlmand (changes requested):
I thought it was agreed that nrfxlib should still be the base for providing the library selection, as to be aligned with other libraries and their Kconfig in nrfxlib.

It was accepted to have the library choice available also in `sdk-nrf/subsys` as a user convenience, but not that this setting should be removed in nrfxlib.

Remember, the location where this setting is used in CMake is also in this repo:
https://github.com/nrfconnect/sdk-nrfxlib/blob/dc28a4f6addb944403c872d69de6f6dc428a8b31/mpsl/CMakeLists.txt#L7-L9

PR #781, opened 2022-07-18
nrf_security/openthread: Disable ECDH legacy ctx

Vge0rge:
-Disable the MBEDTLS_ECDH_LEGACY_CONTEXT from all
 mbedtls configuration files. This has no effect
 right now since it is undefined later by the
 ecdh.h header file in sdk-mbedtls because the
 MBEDTLS_ECP_RESTARTABLE is not set in our builds.
 So removing it should have not effect and it will
 make our configuration more clear.

Ref: NCSDK-15697


openthread/lib/cortex-m33/hard-float/v1.2/ftd/oberon/nrf_security_mbedtls_configuration.h
```diff
@@ -3272,13 +3272,6 @@ it is (2^48 - 1), our restriction is :  (int - 0xFFFF - 0xF).*/
  */
 //#define MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED
 
-/**
- * Nordic added
- *
- * Adding back the legacy ECDH context for binary compatibility
- */
-#define MBEDTLS_ECDH_LEGACY_CONTEXT
```
edmont, line 3280: Hi, I don't think we should modify this, the files are copied together with the Thread libraries for the customers to see what configuration was used to build them. We'll provide new libraries with the next release, so most likely the config will be gone when the script copies the files.

Vge0rge: Thanks for letting me know, I was not aware of this. I will revert all of these then and will only include the nrf_security one.

edmont (approved):

canisLupus1313 (approved):

PR #791, opened 2022-08-02
openthread: add information about GCC version

edmont:
When building OpenThread libraries, add information about the
GCC version that was used to the configuration text file.

Signed-off-by: Eduardo Montoya <eduardo.montoya@nordicsemi.no>

KRKNWK-14765

### DIFF
--- a/openthread/cmake/extensions.cmake
+++ b/openthread/cmake/extensions.cmake
@@ -83,7 +83,14 @@ function(openthread_libs_configuration_write CONFIG_FILE)
     list(INSERT OPENTHREAD_SETTINGS 0 "OpenThread_commit=${git_describe}")
   endif()
 
-  FILE(WRITE ${CONFIG_FILE} ${OPENTHREAD_SETTINGS})
+  # Store compiler and Zephyr SDK version
+  execute_process(COMMAND ${CMAKE_C_COMPILER} --version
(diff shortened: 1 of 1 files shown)


rlubos (approved):

openthread/cmake/extensions.cmake
```diff
@@ -83,7 +83,14 @@ function(openthread_libs_configuration_write CONFIG_FILE)
     list(INSERT OPENTHREAD_SETTINGS 0 "OpenThread_commit=${git_describe}")
   endif()
 
-  FILE(WRITE ${CONFIG_FILE} ${OPENTHREAD_SETTINGS})
+  # Store compiler and Zephyr SDK version
```
Damian-Nordic, line 86: Note there's also `gcc -dumpfullversion` if only the version is needed (without the build details).

Damian-Nordic (approved):

openthread/cmake/extensions.cmake
edmont: Yes, but `--version` shows the Zephyr SDK version as well:
`arm-zephyr-eabi-gcc (Zephyr SDK 0.14.1) 10.3.0`

Damian-Nordic: Right, so that's indeed better.

PR #807, opened 2022-08-22
nrf_modem: doc: Improve OS porting documentation

fnawratil:
Add requirement note for semaphores
Improve IPC interrupt docs
Update implementation template

Signed-off-by: Fabian Nawratil <fabian.nawratil@nordicsemi.no>

### DIFF
--- a/nrf_modem/doc/ug_nrf_modem_porting_os.rst
+++ b/nrf_modem/doc/ug_nrf_modem_porting_os.rst
@@ -128,11 +128,6 @@ This function is called by the Modem library when the library wants to clear IRQ
 
 * Clear the low priority scheduling IRQ of the Modem library using OS primitives or NVIC functions.
 
-nrf_modem_application_irq_handler()
-===================================
-
-This function is implemented in the Modem library and must be called upon the low priority IRQ handler of the Modem library, which is triggered by the :c:func:`nrf_modem_os_application_irq_set` function.
-
 nrf_modem_os_event_notify()
 ===========================
 
@@ -157,6 +152,11 @@ This function is called by the library to allocate and initialize a semaphore.
 * Allocate and initialize a semaphore.
 * If an address of an already allocated semaphore is provided as an input, the allocation part is skipped and the semaphore is only reinitialized.
 
+.. note::
+
+   Semaphores are not required if multithreaded access to modem functionality is not needed.
+   In this case, the function can blindly return ``0``.
+
 nrf_modem_os_sem_give()
 =======================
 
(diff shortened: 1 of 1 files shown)


nrf_modem/doc/ug_nrf_modem_porting_os.rst
```diff
@@ -270,119 +278,222 @@ Reference template for the nrf_modem_os.c file
...
-        // If enabled, initialize the trace task and mediums.
+    #include <nrf_modem.h>
+    #include <nrf_modem_os.h>
+    #include <nrf_errno.h>
+    #include <nrf.h>
+    #include "errno.h"
+
+
+    /**
+    * @brief Initialize the glue layer and required peripherals
```
eivindj-nordic, line 292: Comment indentation seem off
```suggestion
    /**
     * @brief Initialize the glue layer and required peripherals
```

divipillai: The API documentation (doxygen) are always generated from the header file. 
c files are internal to the developer and its up to developer to add it as a normal comment or in doxygen format. I think its always the standard comments that I have seen in the c files (I may be wrong).

fnawratil: Alright, changed comments to standard comments.

nrf_modem/doc/ug_nrf_modem_porting_os.rst
```diff
@@ -270,119 +278,121 @@ Reference template for the nrf_modem_os.c file
...
 
-	void NRF_MODEM_APPLICATION_IRQ_HANDLER(void) {
-		nrf_modem_application_irq_handler();
-	}
+    int nrf_modem_os_sem_init(void **sem, unsigned int initial_count, unsigned int limit) {
+        // The function shall allocate and initialize a semaphore and return its address
+        // as an output. If an address of an already allocated semaphore is provided as
+        // an input, the allocation part is skipped and the semaphore is only reinitialized.
+        return 0;
+    }
```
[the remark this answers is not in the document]
fnawratil: Removing the second * here would change the meaning. 🙂 
A single asterisk denotes a pointer, while two denote a pointer to a pointer or "double pointer".
(This can be extended to even more asterisks as well.)

Also, since this is within a code-block, it shouldn't show up as bold in the docs.

divipillai (approved):

eivindj-nordic (approved):
Good job! 
And thank you for your patience with my nitpicking :)

lemrey (commented):
Looks good, but the code snippets should follow the coding style of NCS imo.
Use C style `/* */` comments and break like so:
```
int foo(void) 
{
}
```
instead of 
```
int foo(void) {
}
```

PR #810, opened 2022-08-24
crypto: doc: Document that the key material can not be in flash

SebastianBoe:
Document that the function nrf_cc3xx_platform_identity_key_store can
not accept a pointer to a key that is stored in flash.

### DIFF
--- a/crypto/nrf_cc310_platform/include/nrf_cc3xx_platform_identity_key.h
+++ b/crypto/nrf_cc310_platform/include/nrf_cc3xx_platform_identity_key.h
@@ -24,8 +24,9 @@
  *
  * @note    This API assumes no format of the data stored.
  *
- * @note    The key material stored using this API should be kept secret, for instance
- *          by generating it directly or the device or securely provisioning it.
+ * @note The key material stored using this API may not be in flash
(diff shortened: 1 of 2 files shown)


crypto/nrf_cc310_platform/include/nrf_cc3xx_platform_identity_key.h
```diff
@@ -24,8 +24,9 @@
  *
  * @note    This API assumes no format of the data stored.
  *
- * @note    The key material stored using this API should be kept secret, for instance
- *          by generating it directly or the device or securely provisioning it.
```
SebastianBoe, line 28: Further improvements are needed.

What are we trying to say when we say

"by generating it directly or the device or securely provisioning it."

SebastianBoe: should it be "on the device" ?

Vge0rge: It should be on the device, yes. Seems like a typo

SebastianBoe:
These docs are not maintained here.

They are copied into this repo from somewhere else.

PR #820, opened 2022-08-31
test: update test-spec.yml

lemrey:
Narrow the pattern that triggers Gazelle, NFC and Find My tests.

At least NFC and Find My tests were triggered by manifest updates
to sdk-nrf that only updated the Modem library, to which
they are completely unrelated.

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -85,21 +85,20 @@
     - "!openthread/doc/**/*"
 
 "CI-nfc-test":
-  - "**/*"
-  - "!softdevice_controller/include/**/*"
-  - "!softdevice_controller/lib/**/*"
-  - "!mpsl/include/**/*"
-  - "!mpsl/lib/**/*"
+  - "nfc/**/*"
 
 "CI-matter-test":
   - "mpsl/include/**/*"
   - "mpsl/lib/**/*"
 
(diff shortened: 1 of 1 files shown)


.github/test-spec.yml
```diff
@@ -85,21 +85,17 @@
...
-  - "!mpsl/lib/**/*"
+  - "nfc/**/*"
 
 "CI-matter-test":
   - "mpsl/include/**/*"
   - "mpsl/lib/**/*"
 
 "CI-find-my-test":
-  - "**/*"
+  - "!nrf_modem/**/*"
```
lemrey, line 95: @kapi-no could you advise a better pattern?

kapi-no: sure

kapi-no, line 95:
```suggestion
  - "crypto/**/**"
  - "nfc/**/**"
  - "nrf_security/**/**"
  - "softdevice_controller/**/**"
```

lemrey: Thanks, updated

thst-nordic (approved):

lemrey:
@shantha-14 this would reduce the load on the build servers, I have added it to the v2.1.0 milestone.

kapi-no (approved):

PR #823, opened 2022-09-05
Add fix for GNU compilation for ssize_t (again) + Add fix for restrict

tothero:
Include of nrf_socket.h prevented successful compilation under GNU:

a) Double declaration of "ssize_t" - check from v1.5.1 has been removed? Why?
b) restrict keyword broke compilation if included as extern "C" in a C++ file.

### DIFF
--- a/nrf_modem/include/nrf_socket.h
+++ b/nrf_modem/include/nrf_socket.h
@@ -17,7 +17,22 @@
 extern "C" {
 #endif
 
+#if !defined(__GNUC__) || (__GNUC__ == 0)
 typedef int32_t ssize_t;
+#else
+#include <sys/types.h>
+#ifdef __SES_ARM
+typedef int32_t ssize_t;
+#endif
+#endif
+
+#ifdef __cplusplus
+#   ifdef __GNUC__
+#       define restrict __restrict__ // G++ has restrict
+#   else
+#       define restrict // C++ in general doesn't
+#   endif
+#endif
 
 /**@addtogroup nrf_socket_address_resolution
  *@{


nrf_modem/include/nrf_socket.h
```diff
@@ -17,7 +17,22 @@
...
 typedef int32_t ssize_t;
+#else
+#include <sys/types.h>
+#ifdef __SES_ARM
+typedef int32_t ssize_t;
+#endif
+#endif
+
+#ifdef __cplusplus
+#   ifdef __GNUC__
```
lemrey, line 30: Should this not be `__GNUG__` ? `__GNUC__` is the C compiler.

nrf_modem/include/nrf_socket.h
```diff
@@ -17,7 +17,22 @@
 extern "C" {
 #endif
 
+#if !defined(__GNUC__) || (__GNUC__ == 0)
```
lemrey, line 20: Why check for `__GNUC__ == 0` ?
If we only need to check for the C compiler, we don't need to check that `__GNUC__` is equal to zero.

nrf_modem/include/nrf_socket.h
```diff
@@ -17,7 +17,22 @@
...
 typedef int32_t ssize_t;
+#else
+#include <sys/types.h>
+#ifdef __SES_ARM
+typedef int32_t ssize_t;
+#endif
```
lemrey, line 26: How can we have `__SES_ARM` defined when `__GNUC__` is defined?

nrf_modem/include/nrf_socket.h
```diff
@@ -17,7 +17,22 @@
 extern "C" {
 #endif
 
+#if !defined(__GNUC__) || (__GNUC__ == 0)
 typedef int32_t ssize_t;
+#else
+#include <sys/types.h>
```
lemrey, line 23: I am not sure about adding a dependency on a POSIX header.
What about checking for `_SSIZE_T_DEFINED` instead?

lemrey (commented):
Hi @tothero, thanks for the patch. I left some questions

nrf_modem/include/nrf_socket.h
tothero: Okey. I basically took this part of code from the history of nrf_modem (v1.5.1). I agree that the !defined part would be enough.

nrf_modem/include/nrf_socket.h
tothero: I assumed some author of this code in this repository (see repro at modem v1.5.1) knows the answer. I may me removed, I agree.

nrf_modem/include/nrf_socket.h
lemrey: It's not clear to me how the C compiler can be used to compile C++ code. `g++` is the GNU C++ compiler and `gcc` is the GNU C compiler. They are both part of GCC but they are different compilers. The C++ compiler defines `__GNUG__` and the C compiler defines `__GNUC__`, as far as I understand.

So I think here `__GNUG__` is correct. The fact that it works with `__GNUC__` is perhaps because the `g++` compiler supports `restrict`.

tothero: Please see: https://gcc.gnu.org/onlinedocs/gcc-3.3/cpp/Common-Predefined-Macros.html
a)
__GNUC__
__GNUC_MINOR__
__GNUC_PATCHLEVEL__
These macros are defined by all GNU compilers that use the C preprocessor: C, C++, and Objective-C. 

b) __GNUG__
The GNU C++ compiler defines this. Testing it is equivalent to testing (__GNUC__ && __cplusplus).

So if we use GNUG we clean up the code, I agree. But in general GNUC is also set by GNU C++ compiler too, meaning the initial commit solution is correct and would explain why it works in our code base. We use c++ in a FreeRTOS environment without Zephyr yet.

tothero: I personally like making the combination of C++ (outer ifdef) and GNUC more explicit more. Because you see that the explicit keyword is only defined away when using C++ on non GNU compilers.

lemrey: Thanks for helping me look into this; I agree with you that it's a bit clearer with two defines.

nrf_modem/include/nrf_socket.h
lemrey: Do you think this would work @tothero? Does your `sys/types.h` file define this macro?

lemrey:
Should be fixed in `libmodem v2.2.0`.

PR #831, opened 2022-09-15
nrf_modem: v1.2.3

MirkoCovizzi:
* Fixed a bug in `nrf_recv()` where the socket context would
crash during receive when closed by another thread.
* Fixed a bug in the IP interface where a race condition would interfere
with the processing of packets read from the modem.
* Replaced the error code `NRF_EINTR` with `NRF_ECANCELED` in
`nrf_recv()`.
    
Signed-off-by: Mirko Covizzi <mirko.covizzi@nordicsemi.no>
Signed-off-by: Emanuele Di Santo <emdi@nordicsemi.no>


divipillai (approved):

lemrey:
You can drop the commit where you add the version `1.2.2`

MirkoCovizzi:
> You can drop the commit where you add the version `1.2.2`. I would add the changes in the `v1.2.2` changelog to the list of changes for `v1.2.3` instead.

Wouldn't that be confusing compared to having the two commits separate (and preserving the original one for 1.2.2)?

lemrey:
What do you think would be confusing, jumping from `v1.2.1` to `v1.2.3`? I don't think that's particularly confusing.

I would like to avoid releasing two libraries for the only purpose of having both. Releasing `v1.2.2` does not have any practical advantage, since we release `v1.2.3` immediately after, which contains all the commits in the previous version anyway. So I don't see the need to have both.

divipillai (approved):

eivindj-nordic (approved):
LGTM
Library build version SHAs are correct for soft- and hardfloat.

lemrey (approved):

nrf_modem/doc/CHANGELOG.rst
```diff
@@ -9,6 +9,19 @@ Changelog
...
+
+* Fixed a bug in :c:func:`nrf_recv()` where the socket context would crash during receive when closed by another thread.
+* Fixed a bug where, in rare cases, the :c:func:`nrf_recv()` function would crash.
+* Replaced the error code ``NRF_EINTR`` with ``NRF_ECANCELED`` in :c:func:`nrf_recv()`.
+* Fixed a memory leak in :c:func:`nrf_recv()` when reading many packets quickly.
+* Fixed a bug in :c:func:`nrf_getaddrinfo()` where the function was not returning the proper protocol suggested by the hints.
+* Fixed a bug in :c:func:`nrf_getaddrinfo()` where specifying ``NRF_AF_UNSPEC`` would incorrectly return an error.
+* Fixed a bug in :c:func:`nrf_setsockopt()` where the option ``NRF_SO_HOSTNAME`` would incorrectly return an error when the hostname was NULL and optlen was 0.
+* Fixed a bug in :c:func:`nrf_modem_gnss_init()` where calling the function would lead to field accuracy speed to always be 0 and to the new GNSS events not working.
+  This issue would occur when GNSS is not enabled in %XSYSTEMMODE and modem functional mode is not online.
```
divipillai, line 23:
```suggestion
  This issue would occur when GNSS is not enabled in ``%XSYSTEMMODE`` and modem functional mode is not online.
```
This could fix the doc build issue in the manifest PR.

MirkoCovizzi: Thank you @divipillai

rlubos (approved):

PR #839, opened 2022-10-07
crypto: Support for building with out of tree cc3xx libaries

mswarowsky:
This enables to easily build with out of tree version of the cc3xx runtime library, by setting the env variable CUSTOM_CC3XX_BUILD_PATH It should point to the folder that includes a nrf_{CC3XX_ARCH}_platform and nrf_{CC3XX_ARCH}_mbedcrypto folder with all the files normally provided by nrfxlib.

Ref: NCSDK-17366

Signed-off-by: Markus Swarowsky <markus.swarowsky@nordicsemi.no>


vili-nordic (approved):

crypto/CMakeLists.txt
```diff
@@ -142,7 +142,14 @@ if (CONFIG_NRF_CC3XX_PLATFORM)
     set (CC3XX_ARCH cc310)
   endif()
 
-  set (NRF_CC3XX_PLATFORM_BASE ${CMAKE_CURRENT_SOURCE_DIR}/nrf_${CC3XX_ARCH}_platform)
+  if (DEFINED ENV{CUSTOM_CC3XX_BUILD_PATH})
```
tejlmand, line 145: we should not suddenly check for an environment variable that suddenly change the code path.

This setting is not documented anywhere, and if we start using this principle here then other part of the NCS codebase will soon copy this principle to other places, making it much harder to understand how to use / configure NCS.

Please ping directly for discussion of alternative solutions depending on exact needs.

tejlmand (changes requested):

mswarowsky:
Based on the feedback for Torsten, gonna close this and work on a solution that is more streamlined with the rest of nrfxlib

PR #845, opened 2022-10-17
nrf_modem: version 2.2.0

lemrey:
Modem library release v2.2.0.

The biggest change is the removal of the EGU usage for rescheduling the IPC interrupt.
See CHANGELOG for details.


rlubos (approved):

nrf_modem/doc/CHANGELOG.rst
```diff
@@ -9,6 +9,19 @@ Changelog
...
+nrf_modem 2.2.0
+***************
+
+* Added `timeout` parameter to `nrf_modem_trace_get()`.
+* Fixed an issue when compiling the :file:`nrf_modem.h` header in C++.
+* The Delta DFU interface (:file:`nrf_modem_delta_dfu.h`) is now thread safe.
+* Fixed possible race conditions in the `nrf_modem_init()` and `nrf_modem_shutdown()` functions.
+* Fixed a bug in `nrf_listen()` that let the queue of incoming connection requests be of size one.
+* The ``NRF_MODEM_GNSS_EVT_BLOCKED`` event is now sent only when the GNSS stack does not get any runtime due to LTE activity, while earlier it could be sent also when the GNSS stack average runtime was too short.
+* Removed the application EGU interrupt. The library uses only the IPC peripheral interrupt now.
```
divipillai, line 21:
```suggestion
* Fixed possible race conditions in the :c:func:`nrf_modem_init` and :c:func:`nrf_modem_shutdown` functions.
* Fixed a bug in the :c:func:`nrf_listen` function that allows the queue of incoming connection requests to be of size one.
* The :c:data:`NRF_MODEM_GNSS_EVT_BLOCKED` event is now sent only when the GNSS stack does not get any runtime due to LTE activity, whereas earlier it could be sent when the GNSS stack average runtime was too short.
* Removed the EGU interrupt of the application. The library uses only the IPC peripheral interrupt now.
```

nrf_modem/doc/ug_nrf_modem_porting_os.rst
```diff
@@ -148,13 +131,10 @@ OS abstraction layer
...
-=======================
-
 An OS abstraction layer implementation for the Zephyr RTOS is also available in the |NCS|.
 See :file:`nrf_modem_os.c` for more details.
-The implementation details for each function are shown in the section following and in the header file:
+
+The implementation details for each function are shown in the following section and in the header file:
 
 nrf_modem_os_init()
 -------------------
```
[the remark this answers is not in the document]
lemrey: OK, I have updated it by re-adding the heading above to fix the level.

nrf_modem/doc/CHANGELOG.rst
```diff
@@ -9,6 +9,19 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+nrf_modem 2.2.0
+***************
+
+* Added `timeout` parameter to `nrf_modem_trace_get()`.
```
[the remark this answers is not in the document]
lemrey: Ok

nrf_modem/doc/CHANGELOG.rst
lemrey: why remove _also_ in `[...] it could be sent also when the GNSS`? I think it emphasizes that earlier it could have been sent in both cases, now only in one.

not sure about changing _let_ into _allow_ `a bug .. that let the queue .. be of size one` ->
`a bug .. that allows the queue .. to be of size one`

The queue can be of size one under normal operation. It's always allowed to be of size one, in a sense, if one is the number of pending incoming connections. The bug made it be of size one always.

eivindj-nordic (approved):

nrf_modem/doc/CHANGELOG.rst
divipillai: Thanks for the explanation, and after going through this, I would stick by your initial statements. I have updated the comments accordingly.

lemrey: Ok, regarding `Removed the EGU interrupt of the application` I have rephrased it as
`Removed the usage of the application software interrupt.`

lemrey: In theory, it could have been a peripheral other than the EGU to generate the software interrupt, so I think that's clearer.

divipillai (approved):

PR #838, opened 2022-10-03
Upmerge 2022.09.27

anangl:
Updates needed for NCS upmerge 2022.09.27.

*treewide: Update Zephyr include paths*

- Add `zephyr/` prefix where needed
- Replace deprecated `<zephyr/zephyr.h>` with `<zephyr/kernel.h>`

*nrf_rpc: Fix array out-of-bounds warnings*

The newest zephyr-sdk contains GCC version that started
to complain about array out-of-bounds. This is false positive
because the actual content of the array is added by the linker.

PR #846, opened 2022-10-19
nrf_rpc: Fix array out-of-bounds warnings

doki-nordic:
The newest zephyr-sdk contains GCC version that started
to complain about array out-of-bounds. This is false positive
because the actual content of the array is added by the linker.


stephanosio (approved):

anangl:
Commit 0cb1767d0f3170d8f9e006242a063aa93e89ace8 was cherry-picked into #838.

PR #863, opened 2022-11-14
nrf_security: Fix mbedlts Kconfig issues for MD5 and MbedTLS

Vge0rge:
This fixes two issues that we found while trying to build some zephyr samples.
1) When CONFIG_MBEDTLS and CONFIG_USERSPACE are enabled the mbedtls_partition is defined
    by zephyr so nrf_security should not define it. Redefinition of the partition is not allowed and is causing 
    a build failure.

2) The CONFIG_MBEDTLS_MAC_MD5_ENABLED was not defined in nrf_security before. This was
    originally defined in: zephyr/modules/mbedtls/Kconfig.tls-generic. But we have a patch on this file which 
    does not allow this Kconfig to be enabled when CONFIG_NRF_SECURITY is not enabled. 
    The addition here allows the Kconfig to be enabled since
    it is used by some network protocols and it is supported by nrf_security. 

Ref: NCSDK-17995

### DIFF
--- a/nrf_security/src/zephyr/CMakeLists.txt
+++ b/nrf_security/src/zephyr/CMakeLists.txt
@@ -64,8 +64,17 @@ if(DEFINED src_zephyr)
     $<TARGET_PROPERTY:${mbedcrypto_target},INTERFACE_COMPILE_DEFINITIONS>
   )
 
-  zephyr_library_sources_ifdef(CONFIG_USERSPACE mbedtls_partition.c)
+  # We need the k_mbedtls_partition when userspace is
+  # used and nrf_security is enabled. However when CONFIG_MBEDTLS
+  # is enabled the partition is defined by Zephyr and therefore
+  # we cannot redefine it here.
+  # (see zephyr/doc/kernel/usermode/memory_domain.rst)
+  if((NOT CONFIG_MBEDTLS) AND CONFIG_USERSPACE)
+    zephyr_library_sources(mbedtls_partition.c)
+  endif()
+
   zephyr_library_app_memory(k_mbedtls_partition)
+
 endif()
 
 # Add configuration/options from zephyr interface libraries
--- a/nrf_security/Kconfig
+++ b/nrf_security/Kconfig
@@ -182,6 +182,9 @@ config MBEDTLS_MAC_SHA256_ENABLED
 	select PSA_WANT_ALG_SHA_256
 	select PSA_WANT_ALG_HMAC
 
+config MBEDTLS_MAC_MD5_ENABLED
+	bool "MD5 hash algorithm"
+
 config MBEDTLS_CTR_DRBG_ENABLED
 	bool "Enable the CTR_DRBG AES-256-based random generator"
 	select PSA_WANT_ALG_CTR_DRBG


frkv (approved):
lgtm

nrf_security/src/zephyr/CMakeLists.txt
```diff
@@ -64,8 +64,13 @@ if(DEFINED src_zephyr)
     $<TARGET_PROPERTY:${mbedcrypto_target},INTERFACE_COMPILE_DEFINITIONS>
   )
 
-  zephyr_library_sources_ifdef(CONFIG_USERSPACE mbedtls_partition.c)
-  zephyr_library_app_memory(k_mbedtls_partition)
+  # When CONFIG_MBEDLTS is enabled Zephyr defines the MbedTLS
+  # partition in zephyr/kernel/userpace.c
+  if(NOT CONFIG_MBEDTLS)
+    zephyr_library_sources_ifdef(CONFIG_USERSPACE mbedtls_partition.c)
```
joerchan, line 70: I don't follow the logic here at all. If USERSPACE is enabled then userspace.c is compiled in, which defines a partition if CONFIG_MBEDTLS is enabled. So the partition is enabled if MBEDTLS && USERSPACE.
If we have a collision shouldn't it be if NOT (MEBDETLS && USERSPACE)?

NOT MBEDTLS && USERSPACE does not make sense to me.

Vge0rge: They had the ifdef with userspace, so my assumption is that this needs to be enabled only when userspace is enabled so  If I do :
if NOT (MBEDTLS && USERSPACE)

Then we will define this partition when:
CONFIG_MBEDTLS=y and CONFIG_USERSPACE=n

Isn't it?

SebastianBoe: You talk about when it should not be added by us (MBEDTLS && USERSPACE). But when is it required?

SebastianBoe: The comment says that Zephyr adds this partition, so why do we do so as well?

Vge0rge: @SebastianBoe  is the comment more clear now?

nrf_security/src/zephyr/CMakeLists.txt
```diff
@@ -64,8 +64,19 @@ if(DEFINED src_zephyr)
     $<TARGET_PROPERTY:${mbedcrypto_target},INTERFACE_COMPILE_DEFINITIONS>
   )
 
-  zephyr_library_sources_ifdef(CONFIG_USERSPACE mbedtls_partition.c)
+  # When CONFIG_MBEDLTS is enabled Zephyr defines the MbedTLS
```
SebastianBoe, line 67:
```suggestion
  # When CONFIG_MBEDLTS and CONFIG_USERSPACE is enabled Zephyr defines the MbedTLS
```

nrf_security/src/zephyr/CMakeLists.txt
```diff
@@ -64,8 +64,19 @@ if(DEFINED src_zephyr)
...
+  # partition in zephyr/kernel/userpace.c
+  # When nrf_security is enabled it provides an implementation
+  # of the the MbedTLS library. Since the MbedTLS library requires
+  # the k_mbedtls_partition when the userspace is enabled
+  # (see /doc/kernel/usermode/memory_domain.rst)
+  # we need to define it here.
+  if(NOT CONFIG_MBEDTLS AND CONFIG_USERSPACE)
```
SebastianBoe, line 74: Based on the above comment, shouldn't this be

if ( NOT (CONFIG_MBEDTLS AND CONFIG_USERSPACE))

When nrf_security is enabled we always need something to provide this partition.

When CONFIG_MBEDTLS AND CONFIG_USERSPACE is enabled this is provided by zephyr, otherwise we provide it.

nrf_security/src/zephyr/CMakeLists.txt
SebastianBoe: (I checked the CMake)

nrf_security/src/zephyr/CMakeLists.txt
SebastianBoe: Comment is so clear, that the bug is now obvious :)

nrf_security/src/zephyr/CMakeLists.txt
Vge0rge: I think that the comment in Zephyr is clear that this is used when the userspace is enabled. What you are proposing enables it even if USERSPACE is disabled. So I will not follow this.

nrf_security/src/zephyr/CMakeLists.txt
Vge0rge: I rewrote it, please check again.

SebastianBoe (approved):

nrf_security/src/zephyr/CMakeLists.txt
Vge0rge: I added the parenthesis to make it more clear

nrf_security/src/zephyr/CMakeLists.txt
SebastianBoe: no bug, comment was unclear :( Clear now though.

joerchan (approved):
Thanks for the clarification, makes sense now.

PR #865, opened 2022-11-16
nrf_security: Remove NRF_SECURITY_RNG

Vge0rge:
This removes NRF_SECURITY_RNG since we now
fully support the Zephyr RNG model for TF-M and
secure only targets. It also selects the PSA RNG
driver from Zephyr for TF-M targets.

Ref: NCSDK-14509


joerchan (approved):

frkv (approved):
LGTM

PR #879, opened 2022-11-30
nrf_security: Set default MPI window size according to nrf_cc3xx

frkv:
-Convert the constant configuration of MBEDTLS_MPI_WINDOW_SIZE to
 Kconfig and change the default to match with the configuration used
 when building the CryptoCell runtime libraries.

ref: NCSDK-18406

### DIFF
--- a/nrf_security/Kconfig.legacy
+++ b/nrf_security/Kconfig.legacy
@@ -254,6 +254,15 @@ config OBERON_ONLY_ENABLED
 
 menu "Legacy mbed TLS crypto APIs"
 
+config MBEDTLS_MPI_WINDOW_SIZE
+	int
+	default 6
+	range 1 6
+	help
+	  Maximum window size used for modular exponentiation.
+	  The default value of 6 is set to match with the configuration used
+	  when building the CryptoCell runtime libraries.
+
 config MBEDTLS_MPI_MAX_SIZE
 	int
 	default 256 if CRYPTOCELL_CC310_USABLE || !CRYPTOCELL_USABLE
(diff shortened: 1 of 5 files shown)


nrf_security/Kconfig.legacy
```diff
@@ -254,6 +254,15 @@ config OBERON_ONLY_ENABLED
 
 menu "Legacy mbed TLS crypto APIs"
 
+config MBEDTLS_MPI_WINDOW_SIZE
+	int
+	default 6
```
Vge0rge, line 259: Maybe we can try to enforce the 6 if CC3XX is enabled. Just to avoid misconfigurations. So something like range 6 6 if CC3XX_BACKEND

nrf_security/Kconfig.legacy
```diff
@@ -254,6 +254,15 @@ config OBERON_ONLY_ENABLED
 
 menu "Legacy mbed TLS crypto APIs"
 
+config MBEDTLS_MPI_WINDOW_SIZE
+	int
```
SebastianBoe, line 258: Do we wish for it to be user-configurable what the value is when CC3X_BACKEND is disabled? If so we should have a prompt that is conditional on CC3X_BACKEND.

If we never want this to be user-configurable we should make it clearer to the user that this is a constant by hardcoding 6 in the mbedtls header file.

SebastianBoe (changes requested):

SebastianBoe (approved):

Vge0rge (approved):

edmont (approved):

PR #883, opened 2022-12-02
nrf_security:  Return code  fixes + minor cleanups for PSA driver wrapper

Vge0rge:
This fixes an issue with the import key function of the psa_crypto_driver_wrappers.c
in which we returned PSA_ERROR_INVALID_ARGUMENT when only one driver
was enabled (CC3XX/Oberon) and it returned PSA_ERROR_NOT_SUPPORTED.
We now return the error code of the driver which matches what the PSA
tests are expecting.

There are also some minor cleanups for the file which doesn't affect functionality
at all.

Ref: NCSDK-17707

### DIFF
--- a/nrf_security/src/psa_crypto_driver_wrappers.c
+++ b/nrf_security/src/psa_crypto_driver_wrappers.c
@@ -687,6 +687,8 @@ psa_status_t psa_driver_wrapper_import_key(
                                               data, data_length,
                                               key_buffer, key_buffer_size,
                                               key_buffer_length, bits ) );
+#else
+            return ( PSA_ERROR_NOT_SUPPORTED );
 #endif /* !MBEDTLS_PSA_BUILTIN_HAS_KEY_TYPE */
 #if defined(PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT)
 #endif /* PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT */
@@ -811,6 +813,8 @@ psa_status_t psa_driver_wrapper_export_public_key(
                                                     data,
                                                     data_size,
                                                     data_length ) );
+#else
+            return ( PSA_ERROR_NOT_SUPPORTED );
 #endif /* !MBEDTLS_PSA_BUILTIN_HAS_KEY_TYPE */
         default:
             /* Key is declared with a lifetime not known to us */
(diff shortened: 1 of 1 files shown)


frkv (approved):
LGTM

nrf_security/src/psa_crypto_driver_wrappers.c
```diff
@@ -688,13 +684,14 @@ psa_status_t psa_driver_wrapper_import_key(
                                               key_buffer, key_buffer_size,
                                               key_buffer_length, bits ) );
 #endif /* !MBEDTLS_PSA_BUILTIN_HAS_KEY_TYPE */
-#if defined(PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT)
-#endif /* PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT */
+            break;
```
joerchan, line 687: This looks like a mistake, break after return.

joerchan: It's hard to trace the valid return code here, but from I can tell we would return PSA_ERROR_CORRUPTION_DETECTED, isn't it more correct to return PSA_ERROR_NOT_SUPPORTED instead?

joerchan: If everything is ifdeffed out here I mean.

joerchan:
```
#if defined(MBEDTLS_PSA_BUILTIN_HAS_KEY_TYPE)
            return( psa_import_key_into_slot( attributes,
                                              data, data_length,
                                              key_buffer, key_buffer_size,
                                              key_buffer_length, bits ) );
#else
            return PSA_ERROR_NOT_SUPPORTED;
#endif /* !MBEDTLS_PSA_BUILTIN_HAS_KEY_TYPE */
```

Based on the other functions this is consistent, and I believe the right thing to do also.
I think `psa_driver_wrapper_export_public_key` should also be updated to follow this pattern.

joerchan (changes requested):

nrf_security/src/psa_crypto_driver_wrappers.c
SebastianBoe: IMHO the root of the bug is the unnecessary complexity from doing

```
switch(a) {
  case b:
   default:
            return( PSA_ERROR_INVALID_ARGUMENT );
}
```

instead of just

```
if(a != b) {
  return PSA_ERROR_INVALID_ARGUMENT;
}
```

SebastianBoe: switch-cases with a single case should be avoided, as that is what an if-statement is.

SebastianBoe: **https://github.com/nrfconnect/sdk-nrfxlib/pull/883/files#diff-8838a2c3e7a631a1a8115f848672dfdad351e3769fd43502620a44ec6b964844R815**

here status is initialized with INVALID_ARGUMENT instead of CORRUPT, is it intentional to behave differently in these two functions?

SebastianBoe (approved):
There are other problems here, but +1 because this is an improvement.

joerchan (approved):

PR #896, opened 2023-01-09
doc: nrf_modem: document how to calculate 1PPS pulse time

tokangas:
Added formula for calculating the current UTC time at the time of an 1PPS pulse.

### DIFF
--- a/nrf_modem/doc/gnss_interface.rst
+++ b/nrf_modem/doc/gnss_interface.rst
@@ -476,6 +476,30 @@ In cases where GNSS is not running continuously, it may be beneficial to change 
 
    1PPS is only supported by modem firmware v1.3.0 or later.
 
+Resolving the UTC time of 1PPS pulse occurrence
+===============================================
+
+As the time of the pulse (aligned to the top of an UTC second) is calculated from the previous valid PVT fix, the latest PVT fix notification needs to be used when resolving the UTC time of the 1PPS pulse.
+While the 1PPS pulse does not have any time delay, the PVT fix notification will always have some delay, both from the PVT solution calculation (approximately 100 ms) and the notification message delivery.
+Therefore, the 1PPS pulse may come before the PVT notification that was used to calculate the exact pulse time.
+
+The UTC time of the 1PPS pulse can be calculated using the following formula:
+
+.. math::
+
+   round\ to\ nearest\ second\ (t_{(PVT,GPST)}+∆t+100 ms)
+
+t\ :sub:`(PVT,GPST)` \ is the GPS time stamp in the previous PVT notification and |delta| t is the time difference between the 1PPS pulse and the reception of previous PVT notification:
+
+.. |delta| unicode:: 0x394 .. capital delta sign
+   :rtrim:
+
+.. math::
+
+   ∆t=t_P-t_{PVT}
+
+Thus, |delta| t is always positive.
+
 .. _gnss_int_agps_data:
 
 A-GPS data


nrf_modem/doc/gnss_interface.rst
```diff
@@ -476,6 +476,30 @@ In cases where GNSS is not running continuously, it may be beneficial to change
 
    1PPS is only supported by modem firmware v1.3.0 or later.
 
+Resolving the UTC time of 1PPS pulse occurrence
+===============================================
+
+As the next pulse time (beginning of an UTC second) is calculated from the previous valid PVT fix, the latest PVT fix notification needs to be used when resolving the UTC time of the 1PPS pulse.
```
hiltunent, line 482:
```suggestion
As the time pulse (aligned to the top of an UTC second) is calculated from the previous valid PVT fix, the latest PVT fix notification needs to be used when resolving the UTC time of the 1PPS pulse.
```

nrf_modem/doc/gnss_interface.rst
```diff
@@ -476,6 +476,30 @@ In cases where GNSS is not running continuously, it may be beneficial to change
...
+While the 1PPS pulse does not have any time delay, the PVT fix notification will always have some delay, both from the PVT solution calculation (approximately 100ms) and the notification message delivery.
```
peknis, line 483:
```suggestion
While the 1PPS pulse does not have any time delay, the PVT fix notification will always have some delay, both from the PVT solution calculation (approximately 100 ms) and the notification message delivery.
```

nrf_modem/doc/gnss_interface.rst
```diff
@@ -476,6 +476,30 @@ In cases where GNSS is not running continuously, it may be beneficial to change
...
+Therefore, the 1PPS pulse may come before the PVT notification that was used to calculate the exact pulse time.
+
+The UTC time of the 1PPS pulse can be calculated using the following formula:
+
+.. math::
+
+   round\ to\ nearest\ second\ (t_{(PVT,GPST)}+∆t+100 ms)
+
+t\ :sub:`(PVT,GPST)` \ is the GPS time stamp in the previous PVT notification and |delta| t is the time difference between the 1PPS pulse and the reception of previous PVT notification:
```
peknis, line 492:
```suggestion
t\ :sub:`(PVT,GPST)` \ is the GPS time stamp in the previous PVT notification and |delta| ``t`` is the time difference between the 1PPS pulse and the reception of previous PVT notification:
```

nrf_modem/doc/gnss_interface.rst
```diff
@@ -476,6 +476,30 @@ In cases where GNSS is not running continuously, it may be beneficial to change
...
+
+.. |delta| unicode:: 0x394 .. capital delta sign
+   :rtrim:
+
+.. math::
+
+   ∆t=t_P-t_{PVT}
+
+Thus, |delta| t is always positive.
```
peknis, line 501:
```suggestion
Thus, |delta| ``t`` is always positive.
```

peknis (changes requested):

peknis:
Does this deserve a changelog entry?

nrf_modem/doc/gnss_interface.rst
tokangas: Or maybe "time of the pulse"?

Can you say "aligned to the **top** of a second"? @peknis?

peknis: I have to say it sounds a bit weird. Can we somehow go around this?

lemrey:
We don't usually document changes to this documentation in the `CHANGELOG`.

lemrey:
We don't have an intermediate release-notes file either, so where would this be added? there is no entry for `v2.3.0` and it'd be wrong to add it to a previous entry's list of changes.

nrf_modem/doc/gnss_interface.rst
tokangas: "aligned to the beginning of", "aligned to the boundary of". Any other ideas?

tokangas:
> We don't have an intermediate release-notes file either, so where would this be added? there is no entry for `v2.3.0` and it'd be wrong to add it to a previous entry's list of changes.

Well, looks like we might skip the changelog entry then. 😄

nrf_modem/doc/gnss_interface.rst
hiltunent: This top of the second thing is basically reflecting u-blox timing parlance. I always assumed it is correct way to say it, but I don't know if I've actually seen it being used elesewhere.

hiltunent: Okay, top of a second seems to be common expression in the domain of timing. If you google "top of a second" & UTC you'll find plenty of hits

tokangas: Seems that you are right, this just didn't ring a bell for me. @peknis, is "aligned to the top of an UTC second" then fine for you?

And do we go with "the time pulse"? To me "time of the pulse" sounds better, because that's what we're actually calculating.

peknis: To be honest, I'm not really familiar with the topic. If you say the "top of a second" is a common expression, so be it. Same with the "time of the pulse".

nrf_modem/doc/gnss_interface.rst
tokangas: @peknis This gets rendered like this, is this what we want?

image

peknis: I thought we might want "t" to stand out. If it bothers you, take it out. It is not code, after all.

lemrey:
If you want to mention this somewhere you could mention it in the release notes for `ncs v2.3.0`, and add it to `release-notes-changelog.rst` in `sdk-nrf`.

nrf_modem/doc/gnss_interface.rst
tokangas: I think it looks better if both `delta` and `t` are the same color and font. I tried the one below, but that gets rendered like this:
image
```suggestion
Thus, ``|delta| t`` is always positive.
```
So maybe we'll go with the original...

peknis (approved):

hiltunent (approved):

PR #901, opened 2023-01-18
doc: nrf_modem: add section about PDN-related error handling

lemrey:
Add section on how to handle PDN-related errors in networking sockets.

### DIFF
--- a/nrf_modem/doc/at_interface.rst
+++ b/nrf_modem/doc/at_interface.rst
@@ -286,10 +286,10 @@ Receiving AT notifications
 **************************
 
 The Modem library can dispatch incoming AT notifications from the modem to a user-provided callback function set by :c:func:`nrf_modem_at_notif_handler_set`.
-Only one callback function must be registered with the Modem library.
+Only one callback function can be registered with the Modem library.
 Registering a new callback function will override any callback previously set.
-Multiple parts of an application might require receiving of AT notifications, thus AT notifications need to be dispatched as necessary.
-In |NCS|, :ref:`at_monitor_readme` library takes care of the dispatching of notifications.
+The callback function can be unset by setting ``NULL`` as the callback.
+If multiple parts of your application need to receive AT notifications, you must dispatch them from the callback function that you registered.
 
 The following snippet shows how to setup an AT notification handler:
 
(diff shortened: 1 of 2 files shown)


nrf_modem/doc/sockets.rst
```diff
@@ -274,6 +274,19 @@ The following code shows how to route a DNS query to the PDN with ID 1:
 
    nrf_getaddrinfo("example.com", "pdn1", &hints, &result);
 
+Handling PDN errors on sockets
+==============================
+
+During operation, an active PDN connection may be deactivated due to loss of connectivity or other reasons.
+When a socket operation is attempted on a socket that no longer has an active PDN connection, the operation will return ``-1`` and set ``errno`` to ``NRF_ENETDOWN``.
+If the socket is being polled, :c:func:`nrf_poll` will set the ``POLLERR`` flag and set ``errno`` to `NRF_ENETDOWN`; the error can be retrieved asynchronously using the ``NRF_SO_ERROR`` socket option.
```
divipillai, line 282:
```suggestion
If the socket is being polled, the :c:func:`nrf_poll` function will set the ``POLLERR`` flag and set ``errno`` to ``NRF_ENETDOWN``, the error can be retrieved asynchronously using the ``NRF_SO_ERROR`` socket option.
```

nrf_modem/doc/at_interface.rst
```diff
@@ -312,11 +312,16 @@ The following snippet shows how to setup an AT notification handler:
...
+The user is responsible for rescheduling the processing of AT notifications as appropriate.
+
+In |NCS|, the :ref:`at_monitor_readme` library takes care of dispatching notifications to different parts of the application.
 
-   The callback is invoked in an interrupt context.
-   The user is responsible for rescheduling the processing of AT notifications as appropriate.
-   In |NCS|, the :ref:`at_monitor_readme` library and the :ref:`nrf_modem_lib_readme` takes care of the rescheduling.
+.. important::
+   In NCS applications, many libraries use the :ref:`at_monitor_readme` library to register their own callback with the Modem library using the :c:func:`nrf_modem_at_notif_handler_set` function.
+   If you are building an NCS application, do not use the :c:func:`nrf_modem_at_notif_handler_set` function to register your callback.
```
trantanen, line 322: Can we add something like this into the description of `nrf_modem_at_notif_handler_set`?
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrfxlib/nrf_modem/doc/api.html#c.nrf_modem_at_notif_handler_set

lemrey: I would rather not add it there, because that header comes from libmodem and it shouldn't contain things which relate to NCS since the library could be used without NCS. In that case the documentation would not apply :/

nrf_modem/doc/at_interface.rst
```diff
@@ -312,11 +312,16 @@ The following snippet shows how to setup an AT notification handler:
...
+   Instead, use the :ref:`at_monitor_readme` library to dispatch AT notifications to where you need them in your application, and to ensure compatibility with other NCS libraries.
```
eivindj-nordic, line 323: nit:
```suggestion
   Instead, use the :ref:`at_monitor_readme` library to dispatch AT notifications to where you need them in your application, to ensure compatibility with other NCS libraries.
```

nrf_modem/doc/at_interface.rst
trantanen: I understand and then I don't :) I mean generally people develop modules that could be used without its intended main "client" but in practice it'll never be used without it. And we try to keep things general, which prevents us from adding the necessary documentation that we'd like to add. But that's the way it goes :)

trantanen (approved):

divipillai (approved):

didrikr (approved):

PR #911, opened 2023-02-13
nrf_security: Update to Mbed TLS 3.2.1 and TF-M 1.7.0

joerchan:


mswarowsky (approved):

nrf_security/tfm/CMakeLists.txt
```diff
@@ -71,6 +71,13 @@ set(CONFIG_MBEDTLS_USE_PSA_CRYPTO                   False)
 # complete build with all libraries and a full mbedcrypto library for linking.
 set(CONFIG_BUILD_WITH_TFM                           False)
 
+# In TF-M 1.7.0 we required mbedtls_hkdf function in HUK derivation.
+if (CONFIG_TFM_PROFILE_TYPE_MINIMAL)
+set(CONFIG_MBEDTLS_HKDF_C                           False)
+else()
+set(CONFIG_MBEDTLS_HKDF_C                           True)
```
SebastianBoe, line 78: indentation

nrf_security/src/CMakeLists.txt
```diff
@@ -184,6 +189,8 @@ append_with_prefix(src_crypto_base ${ARM_MBEDTLS_PATH}/library
   platform_util.c
   version.c
   constant_time.c
+
+  hkdf.c
```
SebastianBoe, line 193: Shouldn't this only be added when MBEDTLS_HKDF_C is enabled?

I understand that we always need it when TF-M is enabled, but this will add hkdf.c even for non-TFM builds wont it?

joerchan: Will investigate this.

SebastianBoe (approved):

nrf_security/src/CMakeLists.txt
joerchan: Moved the addition of this source file to a more appropriate location, and added if guards for the special case needed.

nrf_security/src/legacy/CMakeLists.txt
```diff
@@ -46,6 +46,13 @@ if(CONFIG_MBEDTLS_LEGACY_CRYPTO_C OR
   )
 endif()
 
+if(CONFIG_MBEDTLS_PSA_CRYPTO_SPM AND CONFIG_MBEDTLS_HKDF_C)
```
frkv, line 49: Complete this with || MBEDTLS_PSA_BUILTIN_ALG_HKDF so we have a way to remove CONFIG_MBEDTLS_HKDF_C more cleanly when legacy configuration is dropped?

nrf_security/configs/psa_crypto_config.h.template
```diff
@@ -385,4 +385,9 @@
 #cmakedefine MBEDTLS_MPI_WINDOW_SIZE       @MBEDTLS_MPI_WINDOW_SIZE@ /**< Maximum window size used. */
 #cmakedefine MBEDTLS_MPI_MAX_SIZE          @MBEDTLS_MPI_MAX_SIZE@ /**< Maximum number of bytes for usable MPIs. */
 
+/* TF-M has dependency on mbedtls_hkdf function in order to avoid a recursion problem.
+ * See tfm_builtin_key_loader.c for explanation.
+ */
+#cmakedefine MBEDTLS_HKDF_C                @MBEDTLS_HKDF_C@
```
frkv, line 391: Nit: We probably could standardize on the naked #defines for legacy configurations. It is wrong that I've started to use the #define SOMTHING 1 for these.

You can end up with "already defined"-issues...

nrf_security/src/legacy/CMakeLists.txt
frkv: Ah. we aren't using the BUILTIN config, but rather relying on legacy configuration for the recursion problem. Resolving myself

frkv (approved):
LGTM

PR #914, opened 2023-02-15
nrf_security: Add missing configuration settings for legacy.

joerchan:
MBEDTLS_USE_PSA_CRYPTO
MBEDTLS_PSA_CRYPTO_STORAGE_C
MBEDTLS_PSA_CRYPTO_CLIENT
Was not correctly set in the legacy crypto configuration header.

Default these to disabled when legacy configuration is enabled
so that this does not cause any changes to the build.

Remove conditional check for SPM. This is removed so this is always
true.

### DIFF
--- a/nrf_security/Kconfig.psa
+++ b/nrf_security/Kconfig.psa
@@ -52,7 +52,7 @@ config PSA_CRYPTO_SECURE
 
 config MBEDTLS_PSA_CRYPTO_CLIENT
 	bool
-	default y
+	default y if !MBEDTLS_LEGACY_CRYPTO_C
 
 rsource "src/drivers/Kconfig"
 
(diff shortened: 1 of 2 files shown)


Vge0rge (changes requested):
In this state this breaks the default built of the PSA samples. 
I  checked the configurations and the main issue with this is that now we enable:
MBEDTLS_PSA_CRYPTO_STORAGE_C in the nrf-config.h of the application but it doesn't have all the requirements for it. We need to investigate what to do with the storage in the APP scope.

joerchan:
Will deal with this at a later time. Putting it into draft until then.

joerchan:
Closing as we don't have a need for this change.

PR #930, opened 2023-03-02
nrf_security: Enable support for MBEDTLS debug logs

vili-nordic:
Adds support for CONFIG_MBEDTLS_DEBUG, so that debug log callbacks and log level is set up for MBEDTLS.

### DIFF
--- a/nrf_security/src/CMakeLists.txt
+++ b/nrf_security/src/CMakeLists.txt
@@ -129,7 +129,11 @@ append_with_prefix(src_tls ${ARM_MBEDTLS_PATH}/library
     ssl_tls13_generic.c
 )
 
-
+if (CONFIG_MBEDTLS_DEBUG)
+  append_with_prefix(src_tls ${ARM_MBEDTLS_PATH}/library
+    ssl_debug_helpers_generated.c
+  )
+endif()
 
 if (COMPILE_PSA_APIS)
   # Add files for PSA build
@@ -280,6 +284,16 @@ if (CONFIG_MBEDTLS_TLS_LIBRARY)
     PUBLIC
       mbedcrypto_common
   )
+
+  if (CONFIG_MBEDTLS_DEBUG)
+    # Add Zephyr's mbedlts module include path because we need rely on their headers for debug logs.
+    target_include_directories(mbedcrypto_common
+        INTERFACE
+          ${ZEPHYR_BASE}/modules/mbedtls/include
+    )
+    zephyr_library_sources(${ZEPHYR_BASE}/modules/mbedtls/debug.c)
+    zephyr_library_sources(debug_init.c)
+  endif()
 endif()
 
 # Include EITS in Zephyr build
--- a/nrf_security/configs/nrf-config.h
+++ b/nrf_security/configs/nrf-config.h
@@ -396,6 +396,12 @@ extern "C" {
 #define MBEDTLS_PSA_BUILTIN_MAC
 #endif
 
+#if defined(CONFIG_MBEDTLS_DEBUG)
+#define MBEDTLS_ERROR_C
+#define MBEDTLS_DEBUG_C
+#define MBEDTLS_SSL_DEBUG_ALL
+#endif
+
 #ifdef __cplusplus
 }
 #endif
--- a/nrf_security/src/debug_init.c
+++ b/nrf_security/src/debug_init.c
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/init.h>
+#include <mbedtls/debug.h>
+
+static int _mbedtls_init(const struct device *device)
+{
+	ARG_UNUSED(device);
+
+#if defined(CONFIG_MBEDTLS_DEBUG_LEVEL)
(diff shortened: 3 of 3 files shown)


nrf_security/configs/legacy_crypto_config.h.template
```diff
@@ -3279,6 +3279,11 @@ it is (2^48 - 1), our restriction is :  (int - 0xFFFF - 0xF).*/
 #define MBEDTLS_ECP_FIXED_POINT_OPTIM      0 /**< Enable fixed-point speed-up */
 #endif
 
+#if defined(CONFIG_MBEDTLS_DEBUG)
+#define MBEDTLS_ERROR_C
+#define MBEDTLS_DEBUG_C
+#define MBEDTLS_SSL_DEBUG_ALL
+#endif
```
joerchan, line 3286: AFAIK you don't need this if you have it in nrf-config.h.

nrf_security/include/mbedtls/platform.h
```diff
@@ -401,6 +401,10 @@ int mbedtls_platform_setup( mbedtls_platform_context *ctx );
  */
 void mbedtls_platform_teardown( mbedtls_platform_context *ctx );
 
+#ifdef CONFIG_MBEDTLS_DEBUG
+void zephyr_mbedtls_debug(void *ctx, int level, const char *file, int line, const char *str);
```
joerchan, line 405: Since you already have a patch to zephyr I don't think we need this change.
see comment: https://github.com/nrfconnect/sdk-zephyr/pull/1087#pullrequestreview-1332255733

nrf_security/src/CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ append_with_prefix(src_tls ${ARM_MBEDTLS_PATH}/library
     ssl_ticket.c
     ssl_tls.c
     ssl_tls13_keys.c
+    ssl_debug_helpers_generated.c
```
joerchan, line 125: I can't see this file.

I think this patch needs to be rebased after the Mbed TLS 3.2.1 update.

joerchan (changes requested):

nrf_security/src/CMakeLists.txt
vili-nordic: I need this to have `mbedtls_ssl_states_str` defined

joerchan (approved):

nrf_security/src/CMakeLists.txt
```diff
@@ -127,6 +127,7 @@ append_with_prefix(src_tls ${ARM_MBEDTLS_PATH}/library
     ssl_tls13_server.c
     ssl_tls13_client.c
     ssl_tls13_generic.c
+    ssl_debug_helpers_generated.c
 )
```
SebastianBoe, line 131:
```suggestion
)

if (CONFIG_MBEDTLS_DEBUG)
append_with_prefix(src_tls ${ARM_MBEDTLS_PATH}/library
ssl_debug_helpers_generated.c
)
endif()
```

Unneccessary source files slow down the build and create noise in the logs.

The whitespace in my proposed patch is not correct.

NB: I know you are following prior-art fairly well. The prior-art is also noisy and slow.

nrf_security/src/CMakeLists.txt
```diff
@@ -279,6 +280,13 @@ if (CONFIG_MBEDTLS_TLS_LIBRARY)
     PUBLIC
       mbedcrypto_common
   )
+
+  target_include_directories(mbedcrypto_common
+      INTERFACE
+        ${ZEPHYR_BASE}/modules/mbedtls/include
+  )
+  zephyr_library_sources_ifdef(CONFIG_MBEDTLS_DEBUG ${ZEPHYR_BASE}/modules/mbedtls/debug.c)
+  zephyr_library_sources_ifdef(CONFIG_MBEDTLS_DEBUG ${ZEPHYR_BASE}/modules/mbedtls/zephyr_init.c)
```
SebastianBoe, line 289: zephyr_init does more than call mbedtls_debug_set_threshold. It can also call mbedtls_memory_buffer_alloc_init for some configurations.

I assume we have our own mbedtls dynamic memory management somehow.

Are we certain a user cannot accidentally configure his build to also invoke zephyr_init's mbedtls_memory_buffer_alloc_init in a way that is incompatible with our own system?

SebastianBoe (changes requested):
See review

nrf_security/src/CMakeLists.txt
```diff
@@ -279,6 +283,16 @@ if (CONFIG_MBEDTLS_TLS_LIBRARY)
...
+
+  if (CONFIG_MBEDTLS_DEBUG)
+    # Add Zephyr's mbedlts module include path because we need rely on their headers for debug logs.
+    target_include_directories(mbedcrypto_common
+        INTERFACE
+          ${ZEPHYR_BASE}/modules/mbedtls/include
+    )
+  endif()
+  zephyr_library_sources_ifdef(CONFIG_MBEDTLS_DEBUG ${ZEPHYR_BASE}/modules/mbedtls/debug.c)
+  zephyr_library_sources_ifdef(CONFIG_MBEDTLS_DEBUG debug_init.c)
```
SebastianBoe, line 295: Now that we have the above ifdef we can do:

zephyr_library_sources(
  debug_init.c
  ${ZEPHYR_BASE}/modules/mbedtls/debug.c
)

SebastianBoe (approved):

PR #933, opened 2023-03-09
nrf_security: Allow import key from PSA core

Vge0rge:
This allows the psa_import_key to use the
implementation for symmetric key types.
This is useful because we can later remove the
import key from the different drivers and reduce
the code size and maintenance of the drivers. The
effect that this will have in current builds is
very minimal.


vili-nordic (approved):

Vge0rge:
We decided that we will handle this later.

PR #938, opened 2023-03-16
nrf_modem: include: add dfu to nrf_modem_init return values

eivindj-nordic:
When performing DFU, the nrf_modem_init() return value will contain the DFU result. This was missing from the documentation when splitting the normal and bootloader mode into separate functions.


junqingzou:
@eivindj-nordic need to update [here](https://github.com/nrfconnect/sdk-nrf/blob/main/include/modem/nrf_modem_lib.h#L59), too?

PR #948, opened 2023-03-24
nrf_security: Use default import_key from PSA core

Vge0rge:
This makes sure that we always call the function
psa_import_key_into_slot when no crypto accelerator import_key functionality is available. This is needed because our Oberon library does not support import_key for symmetric keys but it supports doing operations with symmetric keys. The psa_import_key_into_slot function always includes the logic for symmetric keys, which has minor effects in increasing the code size. The support for asymmetric keys on the other hand is configurable which is a good for us since Oberon does support
import_key with asymmetric keys.

### DIFF
--- a/nrf_security/src/psa_crypto_driver_wrappers.c
+++ b/nrf_security/src/psa_crypto_driver_wrappers.c
@@ -716,15 +716,15 @@ psa_status_t psa_driver_wrapper_import_key(
                return( status );
 #endif /* PSA_CRYPTO_DRIVER_HAS_ACCEL_KEY_TYPES_OBERON */
 #endif /* PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT */
-            /* Fell through, meaning no accelerator supports this operation */
-#if defined(MBEDTLS_PSA_BUILTIN_HAS_KEY_TYPE)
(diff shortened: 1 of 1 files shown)


nrf_security/src/psa_crypto_driver_wrappers.c
```diff
@@ -716,15 +716,11 @@ psa_status_t psa_driver_wrapper_import_key(
                return( status );
 #endif /* PSA_CRYPTO_DRIVER_HAS_ACCEL_KEY_TYPES_OBERON */
 #endif /* PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT */
-            /* Fell through, meaning no accelerator supports this operation */
-#if defined(MBEDTLS_PSA_BUILTIN_HAS_KEY_TYPE)
```
frkv, line 720: Why not add enabled nrf_oberon PSA Crypto driver as a matching criteria? This detail gets lost in a commit in history...

Vge0rge: I prefer to have that as default in the core in all cases. We can try to remove it from the CC3XX driver as well as it doesn't add any value I think.

frkv (approved):
LGTM

PR #963, opened 2023-04-27
nrf_security: add openthread config for mbedtls debug

pjasinski990:
Adds config that makes it possible to enable mbedtls debug configs with openthread. Currently it's impossible to select MBEDTLS_DEBUG if not using CUSTOM_OPENTHREAD_SECURITY (as MBEDTLS is promptless in default openthread configuration).

### DIFF
--- a/nrf_security/Kconfig.tls
+++ b/nrf_security/Kconfig.tls
@@ -439,3 +439,9 @@ endmenu
 
 
 endif # MBEDTLS_TLS_LIBRARY
+
+config OPENTHREAD_MBEDTLS_DEBUG
+	bool "MbedTLS logs for OpenThread"
+	select MBEDTLS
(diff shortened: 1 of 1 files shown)


nrf_security/Kconfig.tls
```diff
@@ -439,3 +439,9 @@ endmenu
 
 
 endif # MBEDTLS_TLS_LIBRARY
+
+config OPENTHREAD_MBEDTLS_DEBUG
+	bool "MbedTLS logs for OpenThread"
+	select MBEDTLS
```
frkv, line 445: Maybe this should rather be depends on MBEDTLS?

pjasinski990: This makes sense, but this config is disabled in our configuration (and is promptless). I believe we need to select it.

PR #965, opened 2023-05-01
TF-M: Remove usage of HEAP

joerchan:
Remove usage of HEAP in platform.c.
The functions free and calloc pulls in heap from the C standard library.
We always call mbedtls_platform_set_calloc_free to set alternative
heap management functions so this is unneccesary.

NCSDK-19561

TF-M uses the mbedtls_platform_set_printf to set 'printf' as the
mbedtls logging function. This is unneccary as we already have set
'printf' as the default function in platform.h, but we don't want to
modify the TF-M source code unless we have to.The functions free and calloc pulls in heap from the C standard library.
We always call mbedtls_platform_set_calloc_free to set alternative
heap management functions so this is unneccesary.


frkv (approved):
LGTM

SebastianBoe (approved):

vili-nordic (approved):

PR #966, opened 2023-05-01
nrf_security: Remove usage of CONFIG_SPM configuration

joerchan:
Remove usage of CONFIG_SPM configuration.
SPM has been removed so the if-statement is always true.


frkv (approved):
LGTM

PR #977, opened 2023-05-09
nrf_security: CFB->OFB for CC3XX PSA driver

Vge0rge:
The CC3XX PSA crypto driver does not support the AES CFB algorithm but it does support the AES OFB. This change replaces the AES CFG configuration with the AES OFB to match what the driver supports.

Ref: NCSDK-21022
Ref: NCSDK-20664


SebastianBoe (approved):

frkv (approved):
LGTM

mia-ko (approved):
Is this something that should be mentioned in the changelog?

Vge0rge:
> 

I don't considering worth mentioning but let me know if you disagree.

PR #981, opened 2023-05-11
ci: Add Wi-Fi crypto dependency to test-spec

krish2718:
Wi-Fi crypto has dependencies on MbedTLS, so, add an entry to the test-spec to catch any breaking changes.

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -130,3 +130,7 @@
   - "nrf_security/**/*"
   - "crypto/**/*"
   - "nrf_rpc/**/*"
+
+"CI-wifi":
+  - "crypto/**/*"
+  - "nrf_security/**/*"


.github/test-spec.yml
```diff
@@ -130,3 +130,6 @@
   - "nrf_security/**/*"
   - "crypto/**/*"
   - "nrf_rpc/**/*"
+
+"CI-wifi":
+  - "crypto/**/*"
```
joerchan, line 135: Crypto just has the base libraries, nrf_security has most of the complexity of building the crypto libraries with mbedtls and PSA drivers.
```suggestion
  - "crypto/**/*"
  - "nrf_security/**/*"
```

torsteingrindvik (approved):
Same logic as when crypto/TFM is triggered, so should be ok.

joerchan (approved):

krish2718:
can we please merge this?

PR #990, opened 2023-05-18
softdevice_controller: Do not use header files when SDC build from src

ppryga-nordic:
If softdevice controller is build from sources as a submodule it should not use header files exported to nrfxlib. These header files may be not changed in SDC used to build. That could end with strange runtime errors that are hard to track.

Signed-off-by: Piotr Pryga <piotr.pryga@nordicsemi.no>


ppryga-nordic:
The PR verified by https://github.com/nrfconnect/sdk-nrf/pull/11261.

rugeGerritsen (approved):

rlubos (approved):

PR #995, opened 2023-05-24
nrf_security: Fix regression when using builtin implemented PSA crypto

frkv:
There is a regression where legacy APIs for:
 - ECP (ECDH/ECDSA)
 - AES
 - AEAD
 - RSA
 - Hash 
are pulled in from nrf_cc3xx library or nrf_oberon library regardless if nrf_cc3xx or nrf_oberon enables said feature in PSA crypto driver configurations. This commit enforces stricter rules to how _ALT is enabled based on enabled PSA crypto driver configurations for said drivers.

Note: Oberon PSA core is unaffected by this as it doesn't rely on legacy Mbed TLS APIs and _ALT configuration.


nrf_security/Kconfig.legacy
```diff
@@ -76,81 +76,102 @@ config MBEDTLS_AES_DECRYPT_ALT
...
+	depends on OBERON_BACKEND || (PSA_CRYPTO_DRIVER_OBERON    && \
+		   PSA_CRYPTO_DRIVER_ALG_CHACHA20_POLY1305_OBERON)|| \
+		   CC3XX_BACKEND || (PSA_CRYPTO_DRIVER_CC3XX      && \
+		   PSA_CRYPTO_DRIVER_ALG_CHACHA20_POLY1305_CC3XX)
 	default y
 
 config MBEDTLS_CHACHAPOLY_ALT
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || (PSA_CRYPTO_DRIVER_CC3XX && \
```
joerchan, line 120: As a simplification: All PSA_CRYPTO_x_CC3XX symbols depend on PSA_CRYPTO_DRIVER_CC3XX.
So you could simplify these && statements to only check PSA_CRYPTO_x_CC3XX feature instead.

nrf_security/Kconfig.legacy
```diff
@@ -76,81 +76,102 @@ config MBEDTLS_AES_DECRYPT_ALT
 
 config MBEDTLS_AES_ALT
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || (PSA_CRYPTO_DRIVER_CC3XX || \
```
joerchan, line 79: Why ||? I can't see that this changes anything, as PSA_CRYPTO_DRIVER_HAS_CIPHER_SUPPORT_CC3XX depends on PSA_CRYPTO_DRIVER_CC3XX , so in effect only PSA_CRYPTO_DRIVER_CC3XX  is needed.

Was this supposed to be &&?

joerchan (commented):
I still don't understand the consequences of this regression.

nrf_security/Kconfig.legacy
frkv: Simplification has merit, will be changed to that

nrf_security/Kconfig.legacy
frkv: Should be &&, will fi

PR #1003, opened 2023-05-30
nrf_security: Only build the oberon PSA driver sources that are used

SebastianBoe:
Build only the Oberon PSA Crypto driver sources that are enabled.

This reduces build time and noise in the logs.


vili-nordic (approved):

nrf_security/src/drivers/nrf_oberon/CMakeLists.txt
```diff
@@ -18,38 +18,38 @@ if(CONFIG_MBEDTLS_PSA_CRYPTO_C)
...
-    oberon_key_pair.c
-    oberon_mac.c
-    oberon_rsa.c
-  )
-
-  append_with_prefix_ifdef(CONFIG_PSA_CRYPTO_DRIVER_ALG_CTR_DRBG_OBERON
-    src_crypto_oberon ${OBERON_PSA_CRYPTO_PATH}/oberon/drivers/
-      oberon_hmac_drbg.c
-  )
+macro(add_oberon_source kconfig filename)
```
joerchan, line 23: Use the `_ifdef` convention for the macro name?

nrf_security/src/drivers/nrf_oberon/CMakeLists.txt
```diff
@@ -18,38 +18,38 @@ if(CONFIG_MBEDTLS_PSA_CRYPTO_C)
...
+  add_oberon_source(HAS_CIPHER_SUPPORT  oberon_cipher.c)
+  add_oberon_source(HAS_RSA_SUPPORT     oberon_rsa.c)
+  add_oberon_source(HAS_HASH_SUPPORT    oberon_hash.c)
+  add_oberon_source(ALG_ECDH            oberon_ecdh.c)
+  add_oberon_source(ALG_ECDSA           oberon_ecdsa.c)
+  add_oberon_source(ALG_CTR_DRBG        oberon_ctr_drbg.c)
+  add_oberon_source(ALG_HMAC_DRBG       oberon_hmac_drbg.c)
 
   # Add builtin functionality (if enabled)
   if(DEFINED src_crypto_oberon)
```
joerchan, line 48: oberon_helpers only added to the sources if atleast one more source is defined.
I find this, confusing, although it could be correct.

Maybe removed the check for list defined (`if (DEFINED src_crypto_oberon)`).
Or split this into two statements?

nrf_security/src/drivers/nrf_oberon/CMakeLists.txt
SebastianBoe: The macro doesn't follow the _ifdef convention.

It constructs a variable from the kconfig and then ifdef's based on that.

In the _ifdef convention you take the variable as-is and ifdef on it.

SebastianBoe: But I could maybe rename it to something more appropriate?

Like

maybe_add_oberon_source
conditionally_add_oberon_source

?

nrf_security/src/drivers/nrf_oberon/CMakeLists.txt
SebastianBoe: I assumed oberon_helpers.c existed to help the other drivers here so the semantics would be desired.

"Maybe removed the check for list defined (if (DEFINED src_crypto_oberon))."

I assumed the if-statement was there for good reason, e.g. that it is possible to execute this code and end up with an empty list. Should I check this assumption and see if we can remove the if-statement?

"Or split this into two statements?"
This proposal I did not understand :(

nrf_security/src/drivers/nrf_oberon/CMakeLists.txt
```diff
@@ -18,38 +18,38 @@ if(CONFIG_MBEDTLS_PSA_CRYPTO_C)
...
+endmacro()
 
-  append_with_prefix_ifdef(CONFIG_PSA_CRYPTO_DRIVER_ALG_HMAC_DRBG_OBERON
-    src_crypto_oberon ${OBERON_PSA_CRYPTO_PATH}/oberon/drivers/
-      oberon_ctr_drbg.c
-  )
+if (COMPILE_PSA_APIS)
+  # expands to
+  # CONFIG_PSA_CRYPTO_DRIVER_HAS_AEAD_SUPPORT_OBERON
+  add_oberon_source(HAS_AEAD_SUPPORT    oberon_aead.c)
```
joerchan, line 35: add_oberon_source_ifdef(FULL_KCONFIG source).
For grepping.

nrf_security/src/drivers/nrf_oberon/CMakeLists.txt
frkv: One can argue that the macro is narrow enough in scope that the small breakage on convention isn't as problematic. _ifdef looks very nice as a suffix

joerchan (approved):

SebastianBoe:
These commits are now in https://github.com/nrfconnect/sdk-nrfxlib/pull/1009

PR #1008, opened 2023-06-01
doc: Use same phrasing in legacy and PSA API docs

SebastianBoe:
We document the RSA key sizes two places. One for the legacy APIs and one for the PSA APIs.

But they use different phrasings and AFAIK they should not.

Change the legacy phrasing to match the PSA phrasing.


SebastianBoe:
will be fixed as a part of a larger effort later.

PR #1010, opened 2023-06-01
Move "Oberon PSA Crypto" into nrf_security

SebastianBoe:
Move "Oberon PSA Crypto" into nrf_security

PR #1014, opened 2023-06-02
Refactoring

SebastianBoe:
Misc. refactoring of CMakeLists.txt to follow best practices.


nrf_security/src/CMakeLists.txt
```diff
@@ -7,12 +7,14 @@
...
-
-set(OBERON_PSA_PATH ${NRFXLIB_DIR}/crypto/nrf_oberon)
-
+# Can be executed by both the NCS and TF-M build systems.
 #
+# For TF-M the nrf_security/tfm/CMakeLists.txt file will execute this file.
+# For NCS  the nrf_security/CMakeLists.txt     file will execute this file.
 #
 
+set(OBERON_PSA_PATH ${NRFXLIB_DIR}/crypto/nrf_oberon)
```
Vge0rge, line 16: The variable is set here as well but not the (helpful) comments above. Clean that up please. 
https://github.com/nrfconnect/sdk-nrfxlib/pull/1010/commits/4920b1088c8c772656a23e05a38c52599f92f440

nrf_security/src/CMakeLists.txt
```diff
@@ -221,6 +196,31 @@ endif()
 
 # Library for TLS protocol support
 if (CONFIG_MBEDTLS_TLS_LIBRARY)
+  # TLS files
```
Vge0rge, line 199: Nothing wrong with this line, I just want note that it's better to merge this commit with the previous one: 3e27e70898dcd92b3aef4f520640ecce5922e145

They have identical names and they do the same thing. No need to be separate I think.

nrf_security/src/core/nrf_oberon/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ target_include_directories(mbedcrypto_common
 target_include_directories(mbedcrypto_common
   INTERFACE
     # Nordic PSA headers
-    ${NRF_SECURITY_ROOT}/include/
+    ${NRF_SECURITY_ROOT}/include
```
Vge0rge: No issue with this line, this and the next commits already exist in https://github.com/nrfconnect/sdk-nrfxlib/pull/1010/commits.
So better remove them from here.

crypto/CMakeLists.txt
```diff
@@ -96,15 +93,15 @@ if (CONFIG_NRF_CC310_BL)
   target_link_libraries(nrfxlib_crypto INTERFACE ${CC310_BL_LIB})
 endif()
 
-if (CONFIG_NRF_CC3XX_PLATFORM)
-  if(CONFIG_SOC_NRF5340_CPUAPP)
-    set (CC3XX_ARCH cc312)
-  else()
-    set (CC3XX_ARCH cc310)
-  endif()
+if(CONFIG_SOC_NRF5340_CPUAPP)
```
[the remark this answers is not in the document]
SebastianBoe: I didn't consider it.

I'm not familiar with it, should I check it out and then use it, or do you trust it enough that I can just use it? :)

nrf_security/CMakeLists.txt
```diff
@@ -9,20 +9,20 @@
...
 if(TARGET mbedcrypto)
-  set(mbedcrypto_target real_mbedcrypto)
-  set(mbedx509_target   real_mbedx509)
-  set(mbedtls_target    real_mbedtls)
-else()
-  set(mbedcrypto_target mbedcrypto)
-  set(mbedx509_target   mbedx509)
-  set(mbedtls_target    mbedtls)
+  set(mbedcrypto_target real_${mbedcrypto_target})
+  set(mbedx509_target   real_${mbedcrypto_target})
```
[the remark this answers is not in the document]
SebastianBoe: oops

nrf_security/src/core/nrf_oberon/CMakeLists.txt
SebastianBoe: 1010 is now rebased on top of 1014 so there are no duplicates in the branches (although github shows the same commits up for review in several PRs).

nrf_security/src/CMakeLists.txt
SebastianBoe: done

Vge0rge (approved):

nrf_security/src/core/CMakeLists.txt
```diff
@@ -5,9 +5,27 @@
...
+  include(builtin/config.cmake)
+elseif(CONFIG_PSA_CORE_OBERON)
+  include(nrf_oberon/config.cmake)
 endif()
 
-if (CONFIG_PSA_CORE_OBERON)
-  add_subdirectory(nrf_oberon)
+if (CONFIG_PSA_CORE_BUILTIN OR CONFIG_PSA_CORE_OBERON)
+  # If either of the above were enabled
+  target_sources(${mbedcrypto_target}
```
frkv, line 15: Shouldn't this be inside COMPILE_PSA_APIS?

Or for simplication: Shouldn't everything inside this file be wrapped in COMPILE_PSA_APIS

nrf_security/src/core/CMakeLists.txt
```diff
@@ -5,9 +5,27 @@
...
 if (CONFIG_PSA_CORE_BUILTIN)
-  add_subdirectory(builtin)
+  include(builtin/config.cmake)
+elseif(CONFIG_PSA_CORE_OBERON)
+  include(nrf_oberon/config.cmake)
 endif()
 
-if (CONFIG_PSA_CORE_OBERON)
-  add_subdirectory(nrf_oberon)
+if (CONFIG_PSA_CORE_BUILTIN OR CONFIG_PSA_CORE_OBERON)
```
frkv, line 13: Related to the comment on subsequent lines: Can't this if-sentence be skipped. You are either using the builtin Mbed TLS core or Oberon PSA core

nrf_security/src/core/CMakeLists.txt
SebastianBoe: I don't know, I'm not familiar with that variable.

Perhaps Joakim knows.

frkv: PSA core is only compiled in secure-only builds and inside the TF-M build. In both cases COMPILE_PSA_APIS is set

When you build "client support" for PSA crypto APIs, the PSA core can't be built. In this case you shouldn't have either builtin or Oberon PSA core

Client support is equivalent with Non-secure (zephyr) image, in this case PSA crypto APIs are provided from TF-M build

crypto/CMakeLists.txt
```diff
@@ -96,15 +93,15 @@ if (CONFIG_NRF_CC310_BL)
...
+  set(CC3XX_ARCH cc312)
+else()
+  set(CC3XX_ARCH cc310)
```
joerchan, line 99: This change is not really a refactor, CC3XX_ARCH will be more broadly defined by this change.

I think instead of else doing `if (CONFIG_HAS_HW_NRF_CC310)` is a more logical check (or elseif if you prefer).
then we don't end up with CC3XX_ARCH being defined as cc310 on non-cryptocell devices, or possibly being the default when a new cc3xx arch is added.

nrf_security/tfm/CMakeLists.txt
```diff
@@ -96,8 +96,6 @@ if (${PROJECT_NAME} STREQUAL "Bootloader")
   set(CONFIG_MBEDTLS_PSA_CRYPTO_SPM False)
 endif()
 
-set(COMPILE_PSA_APIS                                True)
```
joerchan, line 99: Why can this variable be deleted here?

nrf_security/src/core/CMakeLists.txt
joerchan: From what I can tell this is still correct.

The two source files were not guarded by COMPILE_PSA_APIS earlier, and files are still added to the variable src_crypto_core when COMPILE_PSA_APIS  is true.

However I don't think this commit improved a lot, as we intend to remove the builtin fairly quickly, and then we should ideally not split the building of the Oberon PSA core across two files.

SebastianBoe: "However I don't think this commit improved a lot, as we intend to remove the builtin fairly quickly, and then we should ideally not split the building of the Oberon PSA core across two files."

I was not aware we would remove the builtin fairly quickly.

Shall I revert?

SebastianBoe: deleted commit

nrf_security/src/core/CMakeLists.txt
SebastianBoe: commit no longer exists

joerchan (approved):

PR #1025, opened 2023-06-14
Test spec ignore doc change

Tschet1:
CI does not need to run on changes to docs only. Updated the test-spec to reflect that.


Tschet1:
Not sure how to verify this change.

guwa (approved):

grho (approved):

PR #1035, opened 2023-06-28
doc: Moving oberon changeloge to doc

mswarowsky:
Part of centraliting the place for the changelogs of the crypto libaries


crypto/README.rst
```diff
@@ -11,8 +11,8 @@ Crypto Libraries
    doc/nrf_cc310_platform
    doc/nrf_cc310_mbedcrypto
    doc/nrf_oberon
-   nrf_cc310_platform/CHANGELOG.rst
-   nrf_cc310_bl/CHANGELOG
-   nrf_cc310_mbedcrypto/CHANGELOG
-   nrf_oberon/CHANGELOG
+   doc/CHANGELOG_cc310_bl
```
melwee01, line 14: Does it matter that the previous link specified the RST file extension and the new one doesn't?

mswarowsky: I don't think so as all the other files don't have it I aligned it with the rest.

SebastianBoe (approved):

mia-ko (approved):

joerchan:
This looks good, but I on looking at the changelog titles I'm noticing a discrepancy in naming and I'm wondering of you could fix that also as part of this PR.
image

Could you update to these?
Changelog - nrf_cc3xx_bl
Changelog - nrf_oberon

mia-ko:
> This looks good, but I on looking at the changelog titles I'm noticing a discrepancy in naming and I'm wondering of you could fix that also as part of this PR. image
> 
> Could you update to these? Changelog - nrf_cc3xx_bl Changelog - nrf_oberon

Ah, I agree with Joakim here, it should definitely be aligned. But it has to be in a different PR it seems as this was just merged.
Joakim, it says nrf_cc310_bl crypto library, do you want the changelog to be changed to Changelog - nrf_cc3xx_bl and not nrf_cc310_bl? And if nrf_cc3xx_bl, does that mean that the library should be renamed as well? Just checking :)

joerchan:
> do you want the changelog to be changed to Changelog - nrf_cc3xx_bl

No, that was just a mistake, no renaming.

PR #1038, opened 2023-06-30
nrf_802154: Fix kconfiguration for simulation

aescolar:
The 802154 SL code now requires the gppi helper code for the PPI, which is only built if NRFX_PPI.
Select it also when building for simulated targets.


nrf_802154/zephyr/Kconfig.nrfxlib
```diff
@@ -32,7 +32,7 @@ config NRF_802154_SL
 	select DYNAMIC_INTERRUPTS
 	select SENSOR if NRF_802154_TEMPERATURE_UPDATE
 	select TEMP_NRF5_MPSL if NRF_802154_TEMPERATURE_UPDATE
-	select NRFX_PPI if SOC_SERIES_NRF52X
+	select NRFX_PPI if SOC_COMPATIBLE_NRF52X
```
ankuns, line 35: This file is deployed from
https://github.com/nrfconnect/sdk-nrf-802154/blob/main/zephyr/Kconfig.nrfxlib#L35
could you please make the change there?

ankuns (changes requested):

nrf_802154/zephyr/Kconfig.nrfxlib
aescolar: Thanks @ankuns Moved to  https://github.com/nrfconnect/sdk-nrf-802154/pull/375

PR #1059, opened 2023-07-31
openthread: Early support for PSA Crypto API

ArekBalysNordic:
Currently, Joiner, Commissioner, and Coaps functionalities are missing for PSA Crypto API. Disable them until full implementation, and allow testing of PSA Crypto functionalities.


canisLupus1313 (approved):

openthread/Kconfig
```diff
@@ -50,8 +50,8 @@ choice OPENTHREAD_NORDIC_LIBRARY_CONFIGURATION
 config OPENTHREAD_NORDIC_LIBRARY_MASTER
 	bool "Nordic OpenThread master configuration"
 	select OPENTHREAD_COAP
-	select OPENTHREAD_COAPS
-	select OPENTHREAD_COMMISSIONER
+	select OPENTHREAD_COAPS if !OPENTHREAD_NRF_SECURITY_PSA
```
Damian-Nordic, line 53: What happens if we enable these with PSA? Is it a build error or runtime failure?

ArekBalysNordic: If we enable these configs we need to enable TLS and DTLS and ecjpake from the MbedTLS library (without that there was a compilation error). We would like to remove MbedTLS dependency while using PSA Crypto API so for now we decided to disable coaps, commissioner, and joiner until we have support for the mentioned protocols from PSA crypto directly.

PR #1061, opened 2023-08-04
doc: remove usage of :members:

gmarull:
It's not supported by docleaf.


koffes (commented):
The changes which affects us in lc3 looks good, but I would need someone from docs to sign off on this. @greg-fer.

gmarull:
Closing as we're finally not moving to docleaf due to licensing issues.

PR #1089, opened 2023-10-11
nrf_modem: update build system files

lemrey:
Introduce a Kconfig choice for the binary version.

### DIFF
--- a/nrf_modem/Kconfig
+++ b/nrf_modem/Kconfig
@@ -6,7 +6,6 @@
 # Selected by NRF_MODEM_LIB in sdk-nrf
 config NRF_MODEM
 	bool
-	imply FPU_SHARING
 
 menu "nrf_modem (Modem library)"
 	visible if NRF_MODEM
--- a/nrf_modem/Kconfig
+++ b/nrf_modem/Kconfig
@@ -1,3 +1,8 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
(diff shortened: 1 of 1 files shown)


rlubos (approved):

nrf_modem/Kconfig
```diff
@@ -12,14 +15,24 @@ choice NRF_MODEM_BUILD_STRATEGY
...
+	default NRF_MODEM_BINARY_DEFAULT
+
+config NRF_MODEM_BINARY_DEFAULT
+	bool "default"
 	help
-	  Links the application with the library version capable of emitting logs.
-	  This increases the final size of the application.
+	  Link with the default binary.
+
+config NRF_MODEM_BINARY_LOG
```
eivindj-nordic, line 31: I'm wondering whether we should keep NRF_MODEM_LOG to enable modem library logs(?). This changes the Kconfig option to enable library logs and requires an update to the documentation.

eivindj-nordic: Would it makes sense to keep NRF_MODEM_LOG and have that select the binary version?

lemrey: Yes, it would. That's a good point, I will change that.

lemrey:
Have changed the commits, please revist.

nrf_modem/Kconfig
```diff
@@ -1,10 +1,15 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor
```
eivindj-nordic, line 2: nit: This file is older than this

eivindj-nordic (approved):

rlubos (approved):

PR #1090, opened 2023-10-11
nrf_modem: doc: Update GNSS 1PPS documentation

tokangas:
Updated GNSS documentation because 1PPS feature is not supported when LTE is enabled.

### DIFF
--- a/nrf_modem/doc/gnss_interface.rst
+++ b/nrf_modem/doc/gnss_interface.rst
@@ -446,7 +446,7 @@ In cases where GNSS is not running continuously, it may be beneficial to change 
 
 .. note::
 
-   1PPS accuracy is not guaranteed when LTE is active.
+   Do not use the 1PPS feature when LTE is enabled.
 
 Resolving the UTC time of 1PPS pulse occurrence
 ===============================================


nrf_modem/doc/gnss_interface.rst
```diff
@@ -446,7 +446,7 @@ In cases where GNSS is not running continuously, it may be beneficial to change
 
 .. note::
 
-   1PPS accuracy is not guaranteed when LTE is active.
+   1PPS feature must not be used when LTE is enabled.
```
peknis, line 449:
```suggestion
   Do not use the 1PPS feature when LTE is enabled.
```

peknis (changes requested):

nrf_modem/doc/gnss_interface.rst
peknis: I would have to disagree. In fact, "Do not use..." is actually a lot more aggressive way of putting this, but up to you. The original is not misleading by any means.

hiltunent (approved):

nrf_modem/doc/gnss_interface.rst
tokangas: Maybe it's just my background with IETF RFCs, where "must not" means "absolute prohibition" (https://datatracker.ietf.org/doc/html/rfc2119). 😄 

Ok, let's go with your suggestion.

peknis (approved):

PR #1107, opened 2023-11-02
Upmerge 20231010

de-nordic:
Upmerge commits.


eriksandgren (approved):
The change makes sense if these Kconfigs no longer exist, I have not checked how the removal of these Kconfigs affects building something that enables `NRF_DM` , like  https://github.com/nrfconnect/sdk-nrf/blob/main/subsys/dm/Kconfig#L7 with https://github.com/nrfconnect/sdk-nrf/blob/main/subsys/dm/Kconfig#L18 enabled

nordicjm (approved):

PR #1115, opened 2023-11-07
Fix/deinit crashes

krish2718:


nrf_wifi/hw_if/hal/src/hal_api.c
```diff
@@ -1044,6 +1044,39 @@ enum nrf_wifi_status hal_rpu_eventq_process(struct nrf_wifi_hal_dev_ctx *hal_dev
...
+
+		event = nrf_wifi_utils_q_dequeue(hal_dev_ctx->hpriv->opriv,
+						 hal_dev_ctx->event_q);
+
+		nrf_wifi_osal_spinlock_irq_rel(hal_dev_ctx->hpriv->opriv,
+					       hal_dev_ctx->lock_rx,
+					       &flags);
+
+		if (!event) {
+			goto out;
```
sachinthegreen, line 1066: Shouldn't status be updated to `FAIL` here?

nrf_wifi/hw_if/hal/src/hal_api.c
```diff
@@ -1256,22 +1289,11 @@ void nrf_wifi_hal_dev_rem(struct nrf_wifi_hal_dev_ctx *hal_dev_ctx)
...
-	for (i = 0; i < MAX_NUM_OF_RX_QUEUES; i++) {
-		nrf_wifi_osal_mem_free(hal_dev_ctx->hpriv->opriv,
-				       hal_dev_ctx->rx_buf_info[i]);
-		hal_dev_ctx->rx_buf_info[i] = NULL;
-	}
-
 	nrf_wifi_osal_tasklet_kill(hal_dev_ctx->hpriv->opriv,
 				   hal_dev_ctx->event_tasklet);
 
+	hal_rpu_eventq_drain(hal_dev_ctx);
```
sachinthegreen, line 1295: Use return value to log an error?

nrf_wifi/hw_if/hal/src/hal_api.c
```diff
@@ -1353,6 +1386,34 @@ void nrf_wifi_hal_dev_deinit(struct nrf_wifi_hal_dev_ctx *hal_dev_ctx)
 }
 
 
+enum nrf_wifi_status nrf_wifi_hal_lock_rx(struct nrf_wifi_hal_dev_ctx *hal_dev_ctx)
+{
+	enum nrf_wifi_status status = NRF_WIFI_STATUS_FAIL;
```
sachinthegreen, line 1391: Why do we need the status variable here? Maybe just make the function have a `void` return?

nrf_wifi/hw_if/hal/src/hal_api.c
```diff
@@ -1353,6 +1386,34 @@ void nrf_wifi_hal_dev_deinit(struct nrf_wifi_hal_dev_ctx *hal_dev_ctx)
...
+
+	nrf_wifi_osal_spinlock_irq_rel(hal_dev_ctx->hpriv->opriv,
+				       hal_dev_ctx->lock_rx,
+				       &flags);
+
+	status = NRF_WIFI_STATUS_SUCCESS;
+
+	return status;
+}
+
```
sachinthegreen, line 1416: Do we actually need the above two functions (since they are just doing irq take and rel?

sachinthegreen (changes requested):

nrf_wifi/hw_if/hal/src/hal_api.c
krish2718: No, not having any more events is a positive case.

nrf_wifi/hw_if/hal/src/hal_api.c
krish2718: There is error returned currently, may be I will just make it void.

nrf_wifi/hw_if/hal/src/hal_api.c
krish2718: This is the main fix in this PR, so, yes, we do need them, how else can FMAC access HAL's internal `lock_rx`?

sachinthegreen (approved):

VivekUppunda (approved):

PR #1117, opened 2023-11-07
ci: Add a workflow to check Wi-Fi FW patch version

krish2718:
This workflow checks the version between the C header and the firmware patch bins and leaves a comment.

### DIFF
--- a/.github/workflows/check_wifi_fw_patch.yml
+++ b/.github/workflows/check_wifi_fw_patch.yml
@@ -1,0 +1,49 @@
+name: Verify nRF70 Wi-Fi firmware patch version
+on: pull_request
+env:
+    PATCH_INFO_FILE: 'nrf_wifi/fw_if/umac_if/inc/fw/patch_info.h'
+    PATCH_BIN_BASE: 'nrf_wifi/fw_bins'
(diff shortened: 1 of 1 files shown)


.github/workflows/check_wifi_fw_patch.yml
```diff
@@ -0,0 +1,41 @@
...
+                  exit 1
+              else
+                  echo "Patch version verified for $patch" >> "$COMMENT"
+              fi
+          done
+      - name: Comment on PR
+        if: ${{ always() }}
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
thst-nordic, line 34: have you tested this?  

I can't remember if checks in this context have the correct permissions to write comments

krish2718: > Error: Unhandled error: HttpError: Resource not accessible by integration

Token doesn't have permissions, should I use a personal token or should we add `commenting` permissions? @thst-nordic

krish2718: or I have modified to use `pull_request_target` but that requires merging

PR #1125, opened 2023-11-14
wifi: Add storing of FW patching in external flash

krish2718:


rado17 (approved):

sachinthegreen (approved):

nrf_wifi/fw_if/umac_if/inc/fmac_structs_common.h
```diff
@@ -76,6 +76,21 @@ struct nrf_wifi_fmac_fw_info {
 	struct nrf_wifi_fw_info umac_patch_sec;
 };
 
+/**
+ * @brief Structure to hold FW patch chunk information.
+ *
+ */
+struct nrf_wifi_fmac_fw_chunk_info {
+	/** Pointer to the FW patch chunk ID string. */
+	char id_str[16];
```
sachinthegreen, line 85: Can we replace `16` with a macro defined in patch_info.h ?

sachinthegreen (changes requested):

nrf_wifi/fw_if/umac_if/inc/fmac_structs_common.h
krish2718: No, this is internal to driver, doesn't come from FW. I deliberately kept this hidden as the ID strings won't be changed, but if you want I can add a macro in the driver.

sachinthegreen (approved):

PR #1129, opened 2023-11-21
nrf_wifi: Fix maximum transmit buffer size check

krish2718:
The maximum buffer size should include MTU and headroom, fix the check.


krish2718:
Linux PR: https://github.com/NordicPlayground/nrf70-linux-driver/pull/35

P.S. I am working `manifest` action that automates this linking, but till that time we need to manually link.

rlubos (approved):

sachinthegreen (approved):

PR #1138, opened 2023-12-08
nrf_wifi: Fix null point dereference in driver raw tx path

VivekUppunda:
The code commit provides a fix for a NULL point de-reference in the tx path for RAW TX. Since, the error check is for an abnormal condition, it is not hit usually. This fix prevents a null dereference when this abnormal condition is hit.

### DIFF
--- a/nrf_wifi/fw_if/umac_if/src/tx.c
+++ b/nrf_wifi/fw_if/umac_if/src/tx.c
@@ -1830,8 +1830,7 @@ enum nrf_wifi_status nrf_wifi_fmac_start_rawpkt_xmit(void *dev_ctx,
 		 * return failure as network buffer and device
 		 * context are NULL
 		 */
-		def_dev_ctx->raw_pkt_stats.raw_pkt_send_failure += 1;
-		return NRF_WIFI_STATUS_FAIL;
+		goto fail;
 	}
 
 	fmac_dev_ctx = (struct nrf_wifi_fmac_dev_ctx *)dev_ctx;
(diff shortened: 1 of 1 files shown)


sachinthegreen (commented):
This needs a manifest PR in sdk-nrf?

nrf_wifi/fw_if/umac_if/src/tx.c
```diff
@@ -1830,8 +1830,7 @@ enum nrf_wifi_status nrf_wifi_fmac_start_rawpkt_xmit(void *dev_ctx,
 		 * return failure as network buffer and device
 		 * context are NULL
 		 */
-		def_dev_ctx->raw_pkt_stats.raw_pkt_send_failure += 1;
-		return NRF_WIFI_STATUS_FAIL;
+		goto fail;
```
sachinthegreen, line 1833: Why do we need a `fail` condition? We could just remove the `def_dev_ctx->raw_pkt_stats.raw_pkt_send_failure += 1;` and return `NRF_WIFI_STATUS_FAIL` from here.

VivekUppunda: I wished to have a FAIL condition as there was an earlier comment in my previous review that we should be returning at the end of the function. hence the goto. I see tx_cmd_prepare, tx_process functions doing the same. Now, if the comment is the opposite of what we are supposed to follow. 

Do let me know what you think and what we follow. I will make the changes accordingly.

Thanks and regards
Vivek

sachinthegreen (approved):

rado17 (approved):

krga2022 (approved):

PR #1164, opened 2024-01-16
nrf_wifi: Fix synchronization b/w down vs. RX

krish2718:
During interface down we de-initialize the FMAC FW (different from FMAC de-init, which is never called in Zephyr), we have a lock to synchronize the states with interrupt (RX or TX_DONE), but this doesn't avoid the interrupt handlers to be executed later causing crashes as the code doesn't have proper NULL checks.

Add a HAL status that can be used to silently ignore the interrupts if the HAL is disabled to fix crashes. This status is also protected by the same lock used to synchronize these two events.

Fixes SHEL-2359.


sachinthegreen (approved):

rado17 (approved):

krish2718:
Updated the fix, @sachinthegreen @rado17  please re-review.

rado17 (approved):

rlubos (approved):

PR #1168, opened 2024-01-19
nrf_wifi: Add regulatory channel attributes

ajayparida:
Regulatory channel attributes to be provided along with regulatory country code.

### DIFF
--- a/nrf_wifi/fw_if/umac_if/inc/fmac_structs_common.h
+++ b/nrf_wifi/fw_if/umac_if/inc/fmac_structs_common.h
@@ -105,6 +105,16 @@ struct nrf_wifi_fmac_otp_info {
 	unsigned int flags;
 };
 
+/* Maximum number of channels supported in a regulatory
+ * currently set to 42 as hardware supports 2.4GHz and 5GHz.
+ * In 2.4 GHz band maximum 14 channels and
+ * in 5 GHz band maximum 28 channels are supported.
+ * Maximum number of channels will change if more number
+ * of channels are enabled in different bands or
+ * different bands are supported in hardware.
+ */
+#define MAX_NUM_REG_CHANELS 42
+
 /**
  * @brief Structure to hold Regulatory parameter data.
  *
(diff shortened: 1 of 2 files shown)


sachinthegreen (approved):

nrf_wifi/fw_if/umac_if/inc/fmac_structs_common.h
```diff
@@ -105,6 +105,7 @@ struct nrf_wifi_fmac_otp_info {
 	unsigned int flags;
 };
 
+#define MAX_NUM_REG_CHANELS 42
```
krish2718, line 108: A comment that explains how did you arrive at this or when will it change would be good.

nrf_wifi/fw_if/umac_if/inc/fmac_structs_common.h
```diff
@@ -114,6 +115,10 @@ struct nrf_wifi_fmac_reg_info {
 	unsigned char alpha2[NRF_WIFI_COUNTRY_CODE_LEN];
 	 /** Forcefully set regulatory. */
 	bool force;
+	/** Regulatory channels count*/
+	unsigned int reg_chan_count;
+	/** Regulatory channel attributes */
+	struct nrf_wifi_get_reg_chn_info reg_chan_info[MAX_NUM_REG_CHANELS] __NRF_WIFI_ALIGN_4;
```
krish2718, line 121: What about the size of `nrf_wifi_get_reg_chn_info`, it should also be multiple of 4, is that handled?

nrf_wifi/fw_if/umac_if/inc/fmac_structs_common.h
ajayparida: Added comment

nrf_wifi/fw_if/umac_if/inc/fmac_structs_common.h
ajayparida: struct nrf_wifi_get_reg_chn_info reg_chan_info[MAX_NUM_REG_CHANELS] __NRF_WIFI_ALIGN_4";" is required other wise the image itself crash on boot because of alignment issue. No other issue was observed realated to alignment.

krish2718: Sure, it just might be working because it's already aligned to 4, but both the start_address and length should be 4 bytes aligned, so, need to explicitly make sure they are aligned.

ajayparida: The issue is being observed no more even without alignment for "reg_chan_info". So removed __NRF_WIFI_ALIGN_4 for reg_chan_info element.

rado17 (approved):

PR #1196, opened 2024-02-02
nrf_modem: doc: porting_guide: Semaphore update

eivindj-nordic:
Semaphores are required also for non-multithreaded access. Remove the statement that "Semaphores are not required if multithreaded access to modem functionality is not needed. In this case, the function can blindly return 0."


divipillai (approved):

lemrey (approved):

MirkoCovizzi (approved):

anhmolt (approved):

eivindj-nordic:
Added to #1227.

PR #1203, opened 2024-02-08
test-spect: remove mpsl&sdc from some CI labels

guwa:
remove few CI from mpsl&sdc changes


.github/test-spec.yml
```diff
@@ -37,10 +37,6 @@
   - "mpsl/lib/**/*"
 
 "CI-mesh-test":
-  - "softdevice_controller/include/**/*"
-  - "softdevice_controller/lib/**/*"
-  - "mpsl/include/**/*"
-  - "mpsl/lib/**/*"
```
PerMac, line 43: is leaving empty entries accepted?

thst-nordic: according to yml spec empty should get convert to null safely.  
but you could add it to be safe. 
`"CI-mesh-test": null`

miha-nordic (approved):

doublemis1 (approved):
LGTM

mariusz-nordicsemi (approved):
Matter tests disabled as changes are verified by lower layers (Radio Software and Thread tests) and BLE itself. Matter verification would be performed in nightly plan.
In case of any regression the idea is to increase test coverage in BLE plan. And if that wouldn't work, the Matter plan could be enabled again.

PR #1206, opened 2024-02-12
test-spec: remove some CI labels from mpsl&sdc change

guwa:
Few CIs are removed from mpsl&sdc test scope, due to the fact that many ble scenarios are already covered in ble CI


thst-nordic (approved):

guwa:
Please merge this PR, 
All CI has passed in https://github.com/nrfconnect/sdk-nrf/pull/13959, except for thread failure, which is not relevant for this change and also exists on their master branch.

PR #1214, opened 2024-02-15
nrf_wifi: Add Promiscuous mode support

VivekUppunda:
This set of changes adds Promiscuous mode support to nrf_wifi

### DIFF
--- a/nrf_wifi/fw_if/umac_if/inc/default/fmac_structs.h
+++ b/nrf_wifi/fw_if/umac_if/inc/default/fmac_structs.h
@@ -264,11 +264,12 @@ struct nrf_wifi_fmac_callbk_fns {
 	void (*process_rssi_from_rx)(void *os_vif_ctx,
 				     signed short signal);
 #endif /* CONFIG_NRF700X_STA_MODE */
-#ifdef CONFIG_NRF700X_RAW_DATA_RX
+#if defined(CONFIG_NRF700X_RAW_DATA_RX) || defined(CONFIG_NRF700X_PROMISC_DATA_RX)
 	void (*rx_sniffer_frm_callbk_fn)(void *os_vif_ctx,
 					 void *frm,
-					 struct raw_rx_pkt_header *);
-#endif
+					 struct raw_rx_pkt_header *,
+					 bool pkt_free);
+#endif /* CONFIG_NRF700X_RAW_DATA_RX || CONFIG_NRF700X_PROMISC_DATA_RX */
 	void (*reg_change_callbk_fn)(void *os_vif_ctx,
 				     struct nrf_wifi_event_regulatory_change *reg_change,
 				     unsigned int event_len);
@@ -515,6 +516,10 @@ struct nrf_wifi_fmac_vif_ctx {
 	/** Filter setting for Monitor and Promiscuous modes */
 	unsigned char packet_filter;
 #endif /* CONFIG_NRF700X_RAW_DATA_RX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+	/** Promiscuous mode setting */
+	bool promisc_mode;
+#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
 };
 
 /**
--- a/nrf_wifi/fw_if/umac_if/src/event.c
+++ b/nrf_wifi/fw_if/umac_if/src/event.c
@@ -868,6 +868,101 @@ out:
 	return status;
 }
 
+#ifdef CONFIG_NRF700X_SYSTEM_MODE
+static enum nrf_wifi_status
+nrf_wifi_fmac_if_mode_set_event_proc(struct nrf_wifi_fmac_dev_ctx *fmac_dev_ctx,
+				     struct nrf_wifi_event_raw_config_mode *mode_event)
+{
+	enum nrf_wifi_status status = NRF_WIFI_STATUS_SUCCESS;
+	struct nrf_wifi_fmac_dev_ctx_def *def_dev_ctx;
+	struct tx_config *config;
+	struct nrf_wifi_fmac_vif_ctx *vif;
+	unsigned char if_idx = mode_event->if_index;
+
+	def_dev_ctx = wifi_dev_priv(fmac_dev_ctx);
+	config = &def_dev_ctx->tx_config;
+	vif = def_dev_ctx->vif_ctx[if_idx];
+
+	if (!mode_event->status) {
+		vif->mode = mode_event->op_mode;
+#ifdef CONFIG_NRF700X_RAW_DATA_TX
+		vif->txinjection_mode = false;
+#endif /* CONFIG_NRF700X_RAW_DATA_TX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+		vif->promisc_mode = false;
+#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
+		if ((mode_event->op_mode & NRF_WIFI_STA_MODE)
+			== NRF_WIFI_STA_MODE) {
+			mode_event->op_mode ^= NRF_WIFI_STA_MODE;
+			vif->if_type = NRF_WIFI_STA_MODE;
+			config->peers[MAX_PEERS].peer_id = -1;
+			config->peers[MAX_PEERS].if_idx = -1;
+
+#if defined(CONFIG_NRF700X_RAW_DATA_TX) && defined (CONFIG_NRF700X_PROMISC_DATA_RX)
+			if ((mode_event->op_mode ^
+			    (NRF_WIFI_PROMISCUOUS_MODE |
+			     NRF_WIFI_TX_INJECTION_MODE)) == 0) {
+				vif->if_type
+					= NRF_WIFI_STA_PROMISC_TX_INJECTOR;
+				config->peers[MAX_PEERS].peer_id
+					= MAX_PEERS;
+				config->peers[MAX_PEERS].if_idx
+					= if_idx;
+				vif->txinjection_mode = true;
+				vif->promisc_mode = true;
+			}
+#endif /* CONFIG_NRF700X_RAW_DATA_TX && defined CONFIG_NRF700X_PROMISC_DATA_RX */
+#ifdef CONFIG_NRF700X_RAW_DATA_TX
+			if ((mode_event->op_mode ^
+			     NRF_WIFI_TX_INJECTION_MODE) == 0) {
+				config->peers[MAX_PEERS].peer_id
+					= MAX_PEERS;
+				config->peers[MAX_PEERS].if_idx
+					= if_idx;
+				vif->if_type = NRF_WIFI_STA_TX_INJECTOR;
+				vif->txinjection_mode = true;
+			}
+#endif /* CONFIG_NRF700X_RAW_DATA_TX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+			if ((mode_event->op_mode ^
+			     NRF_WIFI_PROMISCUOUS_MODE) == 0) {
+				vif->if_type = NRF_WIFI_STA_PROMISC;
+				vif->promisc_mode = true;
+			}
+#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
+			goto out;
+		}
+#ifdef CONFIG_NRF700X_RAW_DATA_RX
+		if ((mode_event->op_mode & NRF_WIFI_MONITOR_MODE)
+			== NRF_WIFI_MONITOR_MODE) {
+			mode_event->op_mode ^= NRF_WIFI_MONITOR_MODE;
+			vif->if_type = NRF_WIFI_IFTYPE_MONITOR;
+			config->peers[MAX_PEERS].peer_id = -1;
+			config->peers[MAX_PEERS].if_idx = -1;
+#ifdef CONFIG_NRF700X_RAW_DATA_TX
+			if ((mode_event->op_mode ^
+			     NRF_WIFI_TX_INJECTION_MODE) == 0) {
+				config->peers[MAX_PEERS].peer_id
+					= MAX_PEERS;
+				config->peers[MAX_PEERS].if_idx
+					= if_idx;
+				vif->if_type = NRF_WIFI_MONITOR_TX_INJECTOR;
+				vif->txinjection_mode = true;
+			}
+#endif
+			goto out;
+		}
+#endif
+	} else {
+		nrf_wifi_osal_log_err(fmac_dev_ctx->fpriv->opriv,
+				      "%s: Set mode failed!",
+				      __func__);
+		status = NRF_WIFI_STATUS_FAIL;
+	}
+out:
+	return status;
+}
+#endif
 
 static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx *fmac_dev_ctx,
 						    struct host_rpu_msg *rpu_msg)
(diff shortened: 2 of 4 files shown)


nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -970,8 +978,40 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
+									= false;
+#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
 				}
 #endif /* CONFIG_NRF700X_RAW_DATA_RX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+				else if (mode_event->op_mode == (NRF_WIFI_STA_MODE |
+								 NRF_WIFI_PROMISCUOUS_MODE |
+								 NRF_WIFI_TX_INJECTION_MODE)) {
+					def_dev_ctx->vif_ctx[mode_event->if_index]->if_type =
+								NRF_WIFI_STA_PROMISC_TX_INJECTOR;
```
sachinthegreen, line 992: Shouldn't this also be protected by `#ifdef CONFIG_NRF700X_RAW_DATA_TX`

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -970,8 +978,40 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+					def_dev_ctx->vif_ctx[mode_event->if_index]->promisc_mode
+									= false;
+#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
 				}
 #endif /* CONFIG_NRF700X_RAW_DATA_RX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+				else if (mode_event->op_mode == (NRF_WIFI_STA_MODE |
+								 NRF_WIFI_PROMISCUOUS_MODE |
+								 NRF_WIFI_TX_INJECTION_MODE)) {
```
sachinthegreen, line 990: Shouldn't `NRF_WIFI_TX_INJECTION_MODE` also be protected by `#ifdef CONFIG_NRF700X_RAW_DATA_TX`

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -970,8 +978,40 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
 									= true;
 #endif /* CONFIG_NRF700X_RAW_DATA_TX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+					def_dev_ctx->vif_ctx[mode_event->if_index]->promisc_mode
+									= false;
+#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
 				}
 #endif /* CONFIG_NRF700X_RAW_DATA_RX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+				else if (mode_event->op_mode == (NRF_WIFI_STA_MODE |
```
sachinthegreen, line 988: Instead of having different `else if` cases to check for combinations of `NRF_WIFI_STA_MODE`, `NRF_WIFI_PROMISCUOUS_MODE`, `NRF_WIFI_TX_INJECTION_MODE` would it be better to just check for:

`else if (mode_event->op_mode & NRF_WIFI_STA_MODE == NRF_WIFI_STA_MODE )` and then have checks internally for :

`if (mode_event->op_mode & NRF_WIFI_TX_INJECTION_MODE == NRF_WIFI_TX_INJECTION_MODE)` and,

`if (mode_event->op_mode & NRF_WIFI_PROMISCUOUS_MODE == NRF_WIFI_PROMISCUOUS_MODE )`

separately so that relevant code can be protected by the compile time flags appropriately ?

nrf_wifi/fw_if/umac_if/src/fmac_util.c
```diff
@@ -363,7 +363,8 @@ enum nrf_wifi_status nrf_wifi_check_mode_validity(unsigned char mode)
 bool nrf_wifi_util_is_rawpktmode_enabled(struct nrf_wifi_fmac_vif_ctx *vif)
 {
 	if ((vif->if_type == NRF_WIFI_STA_TX_INJECTOR) ||
-	    (vif->if_type == NRF_WIFI_MONITOR_TX_INJECTOR)) {
+	    (vif->if_type == NRF_WIFI_MONITOR_TX_INJECTOR) ||
+	    (vif->if_type == NRF_WIFI_STA_PROMISC_TX_INJECTOR)) {
```
sachinthegreen, line 367: Shouldn't these be protected by #if defined(CONFIG_NRF700X_RAW_DATA_RX) || defined(CONFIG_NRF700X_PROMISC_DATA_RX)

sachinthegreen (changes requested):

nrf_wifi/fw_if/umac_if/src/fmac_util.c
VivekUppunda: No ...this only checks for TX injection and only comes in when config block TX injection is enabled, be it with Promiscuous or Monitor. So this is fine. No check is required

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: No, these come from the UMAC header files and no checks are needed for those.
Thanks
Vivek

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: I need to think about this. I will get back to you on this.

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: No, the interface types and modes come from UMAC header file. Setting mode to Station + Promiscuous + TX injection has a checks. This block does not need that check as it will only be hit if the mode is able to be set to this.

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: I have moved the code into specific mode blocks. Kindly check

nrf_wifi/fw_if/umac_if/inc/default/fmac_structs.h
```diff
@@ -264,10 +264,11 @@ struct nrf_wifi_fmac_callbk_fns {
...
 				     signed short signal);
 #endif /* CONFIG_NRF700X_STA_MODE */
-#ifdef CONFIG_NRF700X_RAW_DATA_RX
+#if defined(CONFIG_NRF700X_RAW_DATA_RX) || defined(CONFIG_NRF700X_PROMISC_DATA_RX)
 	void (*rx_sniffer_frm_callbk_fn)(void *os_vif_ctx,
 					 void *frm,
-					 struct raw_rx_pkt_header *);
+					 struct raw_rx_pkt_header *,
+					 bool pkt_free);
 #endif
```
sachinthegreen, line 272:
```suggestion
#endif /* CONFIG_NRF700X_RAW_DATA_RX || CONFIG_NRF700X_PROMISC_DATA_RX */
```

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -927,62 +927,93 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
-									NRF_WIFI_STA_TX_INJECTOR;
-					def_dev_ctx->vif_ctx[mode_event->if_index]->txinjection_mode
-									= true;
-				} else
-#endif /* CONFIG_NRF700X_RAW_DATA_TX */
-				if (mode_event->op_mode == NRF_WIFI_STA_MODE) {
-					def_dev_ctx->vif_ctx[mode_event->if_index]->if_type =
-									NRF_WIFI_IFTYPE_STATION;
+				vif->txinjection_mode = false;
+#endif
```
sachinthegreen, line 939:
```suggestion
#endif /* CONFIG_NRF700X_RAW_DATA_TX */
```

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -927,62 +927,93 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+				vif->promisc_mode = false;
+#endif
```
sachinthegreen, line 942:
```suggestion
#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
```

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -927,62 +927,93 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
+				if ((mode_event->op_mode & NRF_WIFI_STA_MODE)
+								== NRF_WIFI_STA_MODE) {
+					mode_event->op_mode =
+						(mode_event->op_mode ^ NRF_WIFI_STA_MODE);
+					vif->if_type = NRF_WIFI_STA_MODE;
+					config->peers[MAX_PEERS].peer_id = -1;
+					config->peers[MAX_PEERS].if_idx = -1;
+
+#if defined(CONFIG_NRF700X_RAW_DATA_TX) && defined (CONFIG_NRF700X_PROMISC_DATA_RX)
+					if (mode_event->op_mode) {
```
sachinthegreen, line 952: Why is this check necessary (when we have the inner checks in place) ?

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -927,62 +927,93 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
-									NRF_WIFI_IFTYPE_STATION;
+				vif->txinjection_mode = false;
+#endif
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+				vif->promisc_mode = false;
+#endif
+				if ((mode_event->op_mode & NRF_WIFI_STA_MODE)
+								== NRF_WIFI_STA_MODE) {
+					mode_event->op_mode =
+						(mode_event->op_mode ^ NRF_WIFI_STA_MODE);
```
sachinthegreen, line 946: Is this operation necessary? If yes can change this to `mode_event->op_mode ^= NRF_WIFI_STA_MODE;`

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -927,62 +927,93 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
 				}
 #ifdef CONFIG_NRF700X_RAW_DATA_RX
-				else if (mode_event->op_mode == NRF_WIFI_MONITOR_MODE) {
-					def_dev_ctx->vif_ctx[mode_event->if_index]->if_type =
-									NRF_WIFI_IFTYPE_MONITOR;
-					def_dev_ctx->tx_config.peers[MAX_PEERS].peer_id = -1;
+				if ((mode_event->op_mode & NRF_WIFI_MONITOR_MODE)
+							== NRF_WIFI_MONITOR_MODE) {
+					mode_event->op_mode =
+						(mode_event->op_mode ^ NRF_WIFI_MONITOR_MODE);
```
sachinthegreen, line 993: Is this operation necessary? If yes can change this to `mode_event->op_mode ^= NRF_WIFI_MONITOR_MODE;`

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -927,62 +927,93 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
+					vif->if_type = NRF_WIFI_IFTYPE_MONITOR;
+					config->peers[MAX_PEERS].peer_id = -1;
+					config->peers[MAX_PEERS].if_idx = -1;
+
+					if (mode_event->op_mode) {
```
sachinthegreen, line 998: Why is this check necessary (when we have the inner checks in place) ?

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -927,62 +927,93 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
+						if ((mode_event->op_mode ^
+						     NRF_WIFI_TX_INJECTION_MODE) == 0) {
 #ifdef CONFIG_NRF700X_RAW_DATA_TX
```
sachinthegreen, line 1001: I guess this #ifdef needs to also protect the above `if` condition.

sachinthegreen (changes requested):

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: If the mode setting is only Station or Monitor, we can skip all the other checks. In this case, the code will not do unnecessary `if checks` .

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: This operation is to prevent going through each if check. If the mode is set to zero after the XOR operation, we can use that to prevent further `mode if checks`

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: The operation as explained previously in the previous comment. I will make the change to `mode_event->op_mode ^= NRF_WIFI_MONITOR_MODE;`

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: No, the check is not needed as 'NRF_WIFI_TX_INJECTION_MODE` is part of the UMAC header file and UMAC header files do not have any CONFIG checks. The code within the if block has to be protected.

nrf_wifi/fw_if/umac_if/src/event.c
VivekUppunda: Kindly refer previous comment and note for why this check is placed

sachinthegreen (approved):

kapbh (approved):

nrf_wifi/fw_if/umac_if/src/event.c
```diff
@@ -927,62 +927,94 @@ static enum nrf_wifi_status umac_process_sys_events(struct nrf_wifi_fmac_dev_ctx
...
-					def_dev_ctx->tx_config.peers[MAX_PEERS].peer_id = -1;
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+						if ((mode_event->op_mode ^
+						     NRF_WIFI_PROMISCUOUS_MODE) == 0) {
+							vif->if_type = NRF_WIFI_STA_PROMISC;
+							vif->promisc_mode = true;
+							break;
+						}
+					}
+#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
```
jfischer-no, line 986: This requires some code refactoring, and would like to be moved to a dedicated routine even before another indentation in this commit.

VivekUppunda: I have moved the functionality to a dedicated routine. The Code has multiple CONFIG checks which are also refactored as much as possible to make sure that the functionality is not affected.

PR #1218, opened 2024-02-21
nrf_wifi: Add CONFIG_NRF700X_RAW_DATA_RX check in validity check API

VivekUppunda:
This change adds the CONFIG_NRF700X_RAW_DATA_RX check for monitor mode to prevent monitor mode from being configured when driver is not compiled for the same.

It addresses SHEL-2461


sachinthegreen (approved):

jfischer-no:
@krish2718 @rado17 ptal

krish2718 (approved):

PR #1221, opened 2024-02-22
doc: crypto: Add missing ocrypto APIs to doxygen

mswarowsky:
Several API docs were missing for the ocrypto library added them

Ref: NCSDK-21222

### DIFF
--- a/crypto/doc/api.rst
+++ b/crypto/doc/api.rst
@@ -100,7 +100,7 @@ KMU/KDR APIs
 
 
 .. _crypto_api_nrf_oberon:
-   
+
 nrf_oberon crypto library
 *************************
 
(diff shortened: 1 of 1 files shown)


SebastianBoe (approved):

crypto/doc/api.rst
```diff
@@ -115,6 +115,20 @@ AES - Advanced Encryption Standard APIs
    :project: nrfxlib
    :members:
 
+AES-CBC - AES Cipher Block Chaining Mode
```
mia-ko, line 118: Most of the sections in the file have APIs in the end, maybe we could add it to all new ones as well for consistency? Except perhaps the ocrypto internal types...

mia-ko: Another question is if it should be API and not APIs when only one doxygengroup is presented in the section. But this is relevant for the entire file and does not have to be addressed in this PR.

mia-ko (changes requested):

crypto/doc/api.rst
mswarowsky: Then this is in the API section anyway so maybe it would be better to just remove APIs/API as it should be obvious

mia-ko (approved):

crypto/doc/api.rst
mia-ko: True

mswarowsky: Ok removed all the API(s) in the titles

PR #1239, opened 2024-03-04
Commad status check phase1

ajayparida:
First PR from a series or PR to handle command status check for all the commands.
In this PR, only those commands for which no information is passed either to supplicant/user space is handled. 
On sending the command, host will wait for the command status to be returned  by the firmare. 
Depending upon the status/timeout it will return the result to userspace/supplicant.


ajayparida:
closing it. If required will be picked up later.

PR #1302, opened 2024-04-19
mpsl: Use kconfig to determine library path

hermabe:
Instead of using SOC and ARCH configs to determine the library paths, use mpsl-specific kconfigs. We will then be able to change the paths independently of which value various upstream configs have.

This also makes it simpler to use the same library for different socs, or different libraries for the same cpu core variant.

### DIFF
--- a/mpsl/Kconfig
+++ b/mpsl/Kconfig
@@ -27,4 +27,24 @@ config MPSL_BUILD_TYPE_LIB
 
 endchoice
 
-endif
+if MPSL_BUILD_TYPE_LIB
+config MPSL_LIB_DIR
+	string
+	default "cortex-m4" if SOC_SERIES_NRF52X
+	default "cortex-m33+nodsp" if SOC_COMPATIBLE_NRF5340_CPUNET
+	default "nrf54h20_cpurad" if SOC_NRF54H20_CPURAD
+	default "nrf54l15_cpuapp" if SOC_SERIES_NRF54LX
+	default "bsim_nrfxx" if SOC_SERIES_BSIM_NRFXX
+	help
+	  Hidden helper option to calculate the library path
+
+config MPSL_LIB_FLOAT_ABI_DIR
+	string
+	default "hard-float" if FPU && FP_HARDABI
+	default "softfp-float" if FPU && FP_SOFTABI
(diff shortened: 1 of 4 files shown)


mpsl/Kconfig
```diff
@@ -27,4 +27,24 @@ config MPSL_BUILD_TYPE_LIB
...
+	string
+	default "cortex-m4" if SOC_SERIES_NRF52X
+	default "cortex-m33+nodsp" if SOC_COMPATIBLE_NRF5340_CPUNET
+	default "nrf54h20_cpurad" if SOC_NRF54H20_CPURAD
+	default "nrf54l15_cpuapp" if SOC_SERIES_NRF54LX
+	default "bsim_nrfxx" if SOC_SERIES_BSIM_NRFXX
+	help
+	  Hidden helper option to calculate the library path
+
+config MPSL_LIB_FLOAT_DIR
```
rugeGerritsen, line 41: Maybe ?
```suggestion
config MPSL_LIB_FLOAT_ABI_DIR
```

rugeGerritsen (approved):

mpsl/Kconfig
hermabe: Yes. I am also not a fan of the naming of `MPSL_LIB_DIR`, but there is no real connection between the different values. It is a combination of soc, cortex variant, and a bsim hardcoded value that I dont think is used anywhere.

rugeGerritsen: Another option would be to move more of the logic found in https://github.com/nrfconnect/sdk-nrfxlib/blob/main/common.cmake#L30 to Kconfig and add the possibility to override it somewhere else. I'm not sure if that really solves what we are trying to solve

hermabe: The logic there is used by most (all?) of the libraries in nrfxlib. I think the other libraries can do something similar or same as this, but I didnt want to touch the other libraries for now.

hermabe: I also dont know if we want to force the same paths to the libraries for everyone. Other libraries may have the same implementation for different hw combinations than sdc/mpsl.

vegu-nordic (approved):

PR #1317, opened 2024-05-02
crypto: update to nrf_oberon 3.0.15

Vge0rge:


vili-nordic (approved):

mswarowsky (approved):

crypto/doc/CHANGELOG_oberon.rst
```diff
@@ -9,6 +9,96 @@ Changelog - nrf_oberon
...
 
+nrf_oberon - 3.0.15
+*******************
+
+New version of the nrf_oberon library with the following changes.
+
+Added
+=====
+
+* Added missing ocrypto functions for SECP160 curves which broke some configurations.
```
melwee01, line 20: The grammatical fix here depends on the intended meaning.

There should always be a comma after the word before `which`, which in this case would read, "SECP160  curves, which broke some configurations." 

However, I read this construction as meaning that adding the missing functions broke some configurations.

If this is not the intended meaning, use:

"SECP160 curves that broke..."

I read this as: missing functions broke some configurations, and adding them fixes it.

crypto/doc/CHANGELOG_oberon.rst
```diff
@@ -9,6 +9,96 @@ Changelog - nrf_oberon
...
+* Fixed a bug which resulted in AEAD encryption/decryption with AES-CCM producing incorrect results for large buffers (>1MB data with 12 byte nonce).
```
melwee01, line 21:
```suggestion
* Fixed a bug that caused AEAD encryption/decryption with AES-CCM to produce incorrect results for large buffers (>1MB data with 12 byte nonce).
```

crypto/nrf_oberon/include/ocrypto_constant_time.h
```diff
@@ -38,7 +38,7 @@ extern "C" {
 int ocrypto_constant_time_equal(const void *x, const void *y, size_t length);
 
 /**
- * Variable length compared to zero.
+ * Variable length compare to zero.
```
melwee01, line 41: comparison?

Vge0rge: This is an external file that we get from Oberon so if this is not a blocker I would propose to leave it as is.

crypto/doc/CHANGELOG_oberon.rst
Vge0rge: I was intended to write that the fact that we had the missing functions broke some configurations. Now that I added them back they should be fixed. So I updated the text based on your second suggestion.

melwee01 (approved):

PR #1319, opened 2024-05-03
nrf_wifi: Create dedicated memory pools for Wi-Fi

rado17:
Create dedicated memory pools for Wi-Fi data and
management paths. Introduce corresponding memory
allocation/deallocation calls.

### DIFF
--- a/nrf_wifi/fw_if/umac_if/src/default/fmac_api.c
+++ b/nrf_wifi/fw_if/umac_if/src/default/fmac_api.c
@@ -86,7 +86,7 @@ static void nrf_wifi_fmac_deinit_tx(struct nrf_wifi_fmac_dev_ctx *fmac_dev_ctx)
 
 	tx_deinit(fmac_dev_ctx);
 
-	nrf_wifi_osal_mem_free(def_dev_ctx->tx_buf_info);
+	nrf_wifi_osal_data_mem_free(def_dev_ctx->tx_buf_info);
 }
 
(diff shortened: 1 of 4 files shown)


nrf_wifi/fw_if/umac_if/src/default/fmac_api.c
```diff
@@ -60,7 +60,7 @@ static enum nrf_wifi_status nrf_wifi_fmac_init_tx(struct nrf_wifi_fmac_dev_ctx *
 		def_priv->data_config.max_tx_aggregation *
 		sizeof(struct nrf_wifi_fmac_buf_map_info));
 
-	def_dev_ctx->tx_buf_info = nrf_wifi_osal_mem_zalloc(fmac_dev_ctx->fpriv->opriv,
+	def_dev_ctx->tx_buf_info = nrf_wifi_osal_pkt_mem_zalloc(fmac_dev_ctx->fpriv->opriv,
```
krish2718, line 63: My preference would be to use `data` instead of `pkt`, but not a requirement.

sachinthegreen (commented):
@rado17 : As discussed please move this PR to Zephyr upstream.

nrf_wifi/fw_if/umac_if/src/default/fmac_api.c
sachinthegreen: Also shouldn't there be a corresponding `ctrl` variant now?

PR #1346, opened 2024-05-27
lc3: add dependency to symbols

JordanYates:
Add a dependency on `SW_CODEC_LC3_T2_SOFTWARE` for all the sub-options. This prevents every `default y` symbol from appearing in `autoconf.h` when this module is not used.


koffes (approved):
Nice PR. Thank you.

PR #1372, opened 2024-06-06
kconfig: don't 'select' ZCBOR

hakonfam:
When an option depends on another option being set, its definition should use 'depends on' - not 'select'.

Using 'select' introduces a two way dependency which makes it impossible to express proper kconfig dependencies in some cases.

See 'select pitfalls' chapter in zephyr documentation for details.

Ref: NCSDK-27818


KAGA164 (commented):
Looks good buy you need also align NCS repository to this change

PR #1374, opened 2024-06-07
doc improvement: Update nrf_modem arch diagram

divipillai:
Update the nrf_modem architecture diagram
with DECT NR+ changes.


lemrey (approved):

lemrey:
I wonder if the diagram as such makes sense, we could have a table instead. The relation between the library and the OS glue may be a dedicated diagram somewhere else...

divipillai:
> I wonder if the diagram as such makes sense, we could have a table instead. The relation between the library and the OS glue may be a dedicated diagram somewhere else...

I would say, it is good to keep the architecture diagram to showcase the interaction between different components of Modem library with application and nRF91 Series firmware.

anhmolt (approved):

eivindj-nordic (approved):

PR #1384, opened 2024-06-12
wifi: Fix watchdog re-arming

krish2718:


udaynordic (approved):

krish2718:
Driver is now upstream.

PR #1404, opened 2024-07-08
nrf_modem: update Kconfig

lemrey:
Split the Kconfig that control the variant and build strategy.

### DIFF
--- a/nrf_modem/Kconfig
+++ b/nrf_modem/Kconfig
@@ -11,30 +11,41 @@ config NRF_MODEM
 	select NRFX_IPC if (SOC_SERIES_NRF91X || ZTEST)
 	select REQUIRES_FULL_LIBC
 
+if NRF_MODEM
+
 choice NRF_MODEM_BUILD_STRATEGY
 	prompt "Build strategy"
-	depends on NRF_MODEM
-	default NRF_MODEM_LINK_BINARY_CELLULAR
+
+config NRF_MODEM_LINK_BINARY
+	bool "Link binary"
+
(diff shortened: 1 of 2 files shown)


nrf_modem/Kconfig
```diff
@@ -11,30 +11,41 @@ config NRF_MODEM
...
 	prompt "Build strategy"
-	depends on NRF_MODEM
-	default NRF_MODEM_LINK_BINARY_CELLULAR
+
+config NRF_MODEM_LINK_BINARY
+	bool "Link binary"
+
+endchoice # NRF_MODEM_BUILD_STRATEGY
+
+choice
```
eivindj-nordic, line 24: Should we give the choice a name?

nrf_modem/Kconfig
```diff
@@ -11,30 +11,41 @@ config NRF_MODEM
...
+	prompt "Variant"
+	help
+	  Choose which binary variant to link with: cellular or DECT PHY.
 
 config NRF_MODEM_LINK_BINARY_CELLULAR
```
eivindj-nordic, line 29: Not sure how backwards-compatible we want to be, though these should be renamed.

lemrey: It's good to not break these; plus these are still correct from an NCS perspective.
Elsewhere, where the library can be built from source, these will automatically select the correct flavour.

eivindj-nordic (approved):

MirkoCovizzi (approved):

anhmolt (approved):

PR #1406, opened 2024-07-11
nrf_wifi: Update RPU patch to rev#612647283a5

srinivas1201:
Lmac:
[NRF7X-200]: Process TIM for all received beacons instead of
             DTIM beacons.
[SHEL-2919]: Null frame based legacy power save support in LMAC.

UMAC:
[SHEL-2816]: Protected keep-alive subfield is Handled based on the
             assoc type coming from host.
[SHEL-2817]: New command is added for configuring quiet period to rpu. [SHEL-2798]: Added option to disable WMM mode.
[SHEL-2918]: Add support to configure Null frame based power save.


srinivas1201:
All checks passed. Please review and approve accordingly.

udaynordic (approved):

ajayparida (approved):

PR #1411, opened 2024-07-19
test: CI-suit-dfu-test in test-spec.yml

jnsgsr:
Add triggers for SUIT DFU tests.


ahasztag (approved):
Seems ok, I wondered if adding crypto wouldn't be useful, however it does not seem that there is anythin in nrfxlib connected to the backend we are using.

kszromek-nordic (approved):

PR #1414, opened 2024-07-25
nrf_wifi: Add as a standlone library

krish2718:
Add cmake and kconfig files to make this module as a standlone library.


krish2718:
ported to nrf_wifi

PR #1428, opened 2024-08-08
openthread: add soft float fallback for lib path calculation

maciejbaczmanski:
Add soft float fallback to enable building from libs with softfp ABI enabled.

Code built with `soft` and `softfp` `float-abi` options can be linked together as they both use the same soft-float ABI and generate functions to pass floating-point arguments to integer registers. This is contrary to the `hard` option where floating-point arguments are passed directly to FPU registers. `softfp` additionally allows using hardware floating-point instructions instead of emulated ones.

### DIFF
--- a/openthread/cmake/extensions.cmake
+++ b/openthread/cmake/extensions.cmake
@@ -7,6 +7,8 @@
 # The Build the path of the library taking in nrf_security backend type,
 # OpenThread version and feature set
 function(openthread_calculate_lib_path ot_version lib_path)
+  set(OPENTHREAD_LIB_BASE_DIR "${ZEPHYR_NRFXLIB_MODULE_DIR}/openthread")
+
   if(CONFIG_OPENTHREAD_NORDIC_LIBRARY_MASTER)
     set(ot_feature_set "master")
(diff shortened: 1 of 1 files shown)


canisLupus1313 (approved):

openthread/cmake/extensions.cmake
```diff
@@ -7,7 +7,9 @@
 # The Build the path of the library taking in nrf_security backend type,
 # OpenThread version and feature set
 function(openthread_calculate_lib_path ot_version lib_path)
-  if(CONFIG_OPENTHREAD_NORDIC_LIBRARY_MASTER)
+set(OPENTHREAD_LIB_BASE_DIR "${ZEPHYR_NRFXLIB_MODULE_DIR}/openthread")
+
+if(CONFIG_OPENTHREAD_NORDIC_LIBRARY_MASTER)
```
edmont, line 12: Alignment is off.

edmont (approved):

openthread/cmake/extensions.cmake
maciejbaczmanski: right, thanks!

PR #1431, opened 2024-08-08
nrf_802154: nRF54L15 use non-secure SL library for builds with `CONFIG_TRUSTED_EXECUTION_NONSECURE`

ankuns:
For builds with Kconfig option `CONFIG_TRUSTED_EXECUTION_NONSECURE` the non-secure binaries of nRF 802.15.4 SL library must be used.

To allow this with current `nrfxlib_calculate_lib_path` cmake function, the function is given additional parameter NS_PROVIDED.
This approach does not force other users to provide non-secure versions of their libraries (for example if their libs are secure-agnostic).
The nrfxlib_calculate_lib_path adds `_ns` suffix to the returned library path when the SoC is nRF54L15 and `CONFIG_TRUSTED_EXECUTION_NONSECURE` is enabled.
Other SoCs are out of scope now.

The nRF 802.15.4 SL library for nRF54L non-secure is deployed already with the PR:
https://github.com/nrfconnect/sdk-nrfxlib/pull/1430

### DIFF
--- a/common.cmake
+++ b/common.cmake
@@ -20,6 +20,10 @@ include_guard(GLOBAL)
 #
 # SOC_MODE:            Use the SoC name instead of the SoC architecture when calculating the path.
 #                      For example 'nrf52840' will be used for the path instead of 'cortex-m4'
+# NS_PROVIDED          Use name identifying non-secure library version for builds with
+#                      'CONFIG_TRUSTED_EXECUTION_NONSECURE' on supported SoCs.
+#                      Libraries that do not distinguish between secure and non-secure version
+#                      should not use this parameter.
 # BASE_DIR:            Base path from where the calculated path should start from.
 #                      When specifying BASE_DIR the path returned will be absolute, or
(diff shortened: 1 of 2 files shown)


common.cmake
```diff
@@ -46,6 +50,9 @@ function(nrfxlib_calculate_lib_path lib_path)
       set(arch_soc_dir ${arch_soc_dir}_cpunet)
     elseif(DEFINED CONFIG_SOC_NRF54L15_ENGA_CPUAPP)
       set(arch_soc_dir ${arch_soc_dir}_cpuapp)
+      if(DEFINED CONFIG_TRUSTED_EXECUTION_NONSECURE AND ${CALC_LIB_PATH_NS_PROVIDED})
+        set(arch_soc_dir ${arch_soc_dir}_ns)
```
e-rk, line 54: Shouldn't this be `${arch_soc_dir}_cpuapp_ns`?

e-rk (approved):

dawidprzybylo (approved):

PR #1447, opened 2024-08-27
bluetooth: Use different arch directory per bsim platform

ilhanates:
A seperate arch directory is defined per bsim platform to be able to use different plarforms in the same build process


common.cmake
```diff
@@ -56,8 +56,12 @@ function(nrfxlib_calculate_lib_path lib_path)
     elseif(DEFINED CONFIG_SOC_NRF54H20_CPURAD)
       set(arch_soc_dir ${arch_soc_dir}_cpurad)
     endif()
-  elseif(CONFIG_SOC_SERIES_BSIM_NRFXX)
-    set(arch_soc_dir "bsim_nrfxx")
+  elseif(CONFIG_SOC_SERIES_BSIM_NRF52X)
```
hermabe, line 59: We dont use this function in mpsl/sdc anymore but I guess it is ok to add it here as well.

hermabe (approved):

jori-nordic (approved):

PR #1473, opened 2024-09-18
suit: Add initial version of the SUIT manifest starter

carlescufi:
This is intended for Zephyr upstream use, and allows the nRF54H20 to start the application and radio core images.


PerMac:
As I understand, to add a blob to a module it is required to also link a doc. Maybe adding some file with this minimalistic description would be worth within this PR?

robertstypa (approved):

carlescufi:
> As I understand, to add a blob to a module it is required to also link a doc. Maybe adding some file with this minimalistic description would be worth within this PR?

I will find a suitable link for that in the NCS doc.

PR #1488, opened 2024-09-26
nrf_wifi: Update RPU patch to rev#c96a720c9e4

imapa:
[SHEL-3087]: Determine the Chip package and modify the optimized setting value for PALDOTRIM to 3.4V for CSP and leave it unchanged at 3.3V for QFN.

[SHEL-2768]: Develop patch for Enhanced TX DC algorithm.


kspraveeen:
PR is raised to Upstream.

PR #1512, opened 2024-10-10
kconfig: openthread: add `OPENTHREAD_NORDIC_LIBRARY_COMMON`

maciejbaczmanski:
add `OPENTHREAD_NORDIC_LIBRARY_COMMON` for configs that are selected in all library configurations, and add Thread 1.4 mandatory selections


edmont (approved):

openthread/Kconfig
```diff
@@ -79,32 +89,14 @@ config OPENTHREAD_NORDIC_LIBRARY_MASTER
...
-	select OPENTHREAD_DNS_CLIENT
-	select OPENTHREAD_ECDSA
-	select OPENTHREAD_EXTERNAL_HEAP
-	select OPENTHREAD_IP6_FRAGM
-	select OPENTHREAD_NETDIAG_CLIENT
-	select OPENTHREAD_SLAAC
-	select OPENTHREAD_SRP_CLIENT
-	select OPENTHREAD_DUA if !OPENTHREAD_THREAD_VERSION_1_1
-	select OPENTHREAD_MLR if !OPENTHREAD_THREAD_VERSION_1_1
+	select OPENTHREAD_NORDIC_LIBRARY_COMMON
```
LuDuda, line 99: We don't want to enable TCP in the FTD and MTD libraries. As far as i understand TCP is required by component, but it's not required for end-products. Thus i would keep it enabled only for Master variant.

LuDuda (dismissed):

openthread/Kconfig
maciejbaczmanski: right! fixed

maciejbaczmanski:
@LuDuda is out of office today. I've moved TCP related configs from common feature set to master according to the comment and offline discussion. Dismissing Łukasz's review as we need this merged today to start round of certification tests over the weekend

PR #1514, opened 2024-10-10
nrf_rpc: fix build error for C++

Damian-Nordic:
In C++, conversion from a function pointer to void* is either undefined (prior to C++11) or not implicit.

Move the static inline functions to the C file for better portability. The flash impact imposed by this is negligible.

### DIFF
--- a/nrf_rpc/include/nrf_rpc.h
+++ b/nrf_rpc/include/nrf_rpc.h
@@ -372,9 +372,8 @@ int nrf_rpc_init(nrf_rpc_err_handler_t err_handler);
  * @return             0 on success or negative error code if a transport layer
  *                     reported a sending error.
  */
-static inline int nrf_rpc_cmd(const struct nrf_rpc_group *group, uint8_t cmd,
-			      uint8_t *packet, size_t len,
-			      nrf_rpc_handler_t handler, void *handler_data);
+int nrf_rpc_cmd(const struct nrf_rpc_group *group, uint8_t cmd, uint8_t *packet,
+		size_t len, nrf_rpc_handler_t handler, void *handler_data);
 
 /** @brief Send a command and get response as an output parameter.
  *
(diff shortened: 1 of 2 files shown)


Damian-Nordic:
> Why is C code built by C++ compiler?

Because Matter firmware needs to include `nrf_rpc.h` (to call `nrf_rpc_init`), which includes the c++-prohibited casts.

alxelax:
> > Why is C code built by C++ compiler?
> 
> Because Matter firmware needs to include `nrf_rpc.h` (to call `nrf_rpc_init`), which includes the c++-prohibited casts.

I do not think that `inline` functions are a correct implementation for headers, which are used in both C and C++.
I see problem not in `void *` usage in general, but in inlining of C code within C++ without explicit letting know about that to compiler.

I do not like forced typecasting since it might dig in deep the type mismatching. Would it be more meaningful to move inline functions into C file and remove `inline` modificator?

Damian-Nordic:
> I do not think that `inline` functions are a correct implementation for headers, which are used in both C and C++. I see problem not in `void *` usage in general, but in inlining of C code within C++ without explicit letting know about that to compiler.
> 
> I do not like forced typecasting since it might dig in deep the type mismatching. Would it be more meaningful to move inline functions into C file and remove `inline` modificator?

As long as the inline functions don't use any C-only features like compound literals, there's no issue with that, no?

Moving the functions to C file will not remove the casts. To make it 100% type-safe I would probably need to duplicate some code, which contradicts the authors' intention to optimize the code size of this fragment.

alxelax:
Discussed IRL. Let it be but for future to be accurate with potential miscasting in such functions.

alxelax (approved):

Damian-Nordic:
@doki-nordic Could you please review?

doki-nordic (approved):

PR #1521, opened 2024-10-17
cmake: openthread: change crypto check for prebuilt libs

maciejbaczmanski:
Change approach on crypto checks for prebuilt libraries to compare generated configs instead of mbedtls header file.


openthread/lib/nrf52840/hard-float/v1.3/ftd/openthread_lib_configuration.txt
```diff
@@ -213,3 +219,75 @@ OT_VENDOR_MODEL=
 OT_VENDOR_NAME=
 OT_VENDOR_SW_VERSION=
 OT_VERHOEFF_CHECKSUM=OFF
+
+###################### Crypto configuration ##################
+
+CONFIG_MBEDTLS_AES_DECRYPT_ALT=y
```
ArekBalysNordic, line 225: We can think about it as a follow-up to this PR, but I wonder whether we can have the common configuration which is visible in all variants and versions, and have only the missing ones in each library. What do you think? Having it in this approach increases maintenance costs.

ArekBalysNordic (approved):

openthread/lib/nrf52840/hard-float/v1.3/ftd/openthread_lib_configuration.txt
maciejbaczmanski: I think that's a good idea! I'll create a follow up task for that as I doubt we will make this in time for 2.8.0

kkasperczyk-no (approved):

PR #1545, opened 2024-10-29
nrf_rpc: remove NULL pointer referenece

tomchy:
This reverts half of the changes introduced in the following PR:
https://github.com/nrfconnect/sdk-nrfxlib/pull/1493

Ref: NCSDK-29976

### DIFF
--- a/nrf_rpc/nrf_rpc.c
+++ b/nrf_rpc/nrf_rpc.c
@@ -801,13 +801,7 @@ cleanup_and_exit:
 	}
 
 	if (err < 0) {
-		internal_data.type = NRF_RPC_ERROR;
-		internal_data.group = group;
-		internal_data.err = err;
-		internal_data.hdr_id = hdr.id;
-		internal_data.hdr_type = hdr.type;
-		nrf_rpc_os_thread_pool_send((const uint8_t *)&internal_data, sizeof(internal_data));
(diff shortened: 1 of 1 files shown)


alxelax:
FYI @Damian-Nordic

ahasztag (changes requested):
This does not seem to be enough - the SUIT update still fails, although it passed after reverting the commits which introduced the change reverted here

alxelax:
> This does not seem to be enough - the SUIT update still fails, although it passed after reverting the commits which introduced the change reverted here

Then it should be fixed correctly, not by reverting this commit. If changes are reverted, it breaks protocol serialization functionality.

alxelax (changes requested):
The cause of issue should be investigated and fixed.

tomchy:
> The cause of issue should be investigated and fixed.

Changed back to the partial revert. We are trying to fix the order of messages by adjusting the main thread priority.

alxelax:
Hi @tomchy, please apply this patch: https://github.com/nrfconnect/sdk-nrfxlib/commit/47fdbee5fd83ed4a650b3cce7c1552a472bc570b
This is more correct fix rather than partial reverting.

alxelax:
Alternate PR with fix: https://github.com/nrfconnect/sdk-nrfxlib/pull/1548

PR #1548, opened 2024-10-30
nrf_rpc: fix error handling

alxelax:
Commit fixes null dereference when error handling
is triggered but group does not exist.

### DIFF
--- a/nrf_rpc/nrf_rpc.c
+++ b/nrf_rpc/nrf_rpc.c
@@ -102,6 +102,7 @@ struct internal_pool_data {
 static struct nrf_rpc_cmd_ctx cmd_ctx_pool[CONFIG_NRF_RPC_CMD_CTX_POOL_SIZE];
 
 static struct nrf_rpc_os_event groups_init_event;
+static struct nrf_rpc_os_event error_event;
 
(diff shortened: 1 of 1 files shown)


nrf_rpc/nrf_rpc.c
```diff
@@ -102,6 +102,7 @@ struct internal_pool_data {
 static struct nrf_rpc_cmd_ctx cmd_ctx_pool[CONFIG_NRF_RPC_CMD_CTX_POOL_SIZE];
 
 static struct nrf_rpc_os_event groups_init_event;
+static struct nrf_rpc_os_event error_event;
```
Damian-Nordic, line 105: I would call it as `internal_data_consumed` and use it for all types of internal packets for consistency.

Damian-Nordic (approved):

nrf_rpc/nrf_rpc.c
alxelax: Kind of agree, but let's do improvements as separate activity. I guess this PR is mostly hotfix.

ahasztag (approved):

tomchy (approved):

PR #1568, opened 2024-11-10
wifi: Fix memleak in regulatory processing

krish2718:
Fixes a leak in regulatory processing.


rado17 (approved):

sachinthegreen (approved):

krish2718:
will revisit when there is a release from v2.7-branch.

PR #1581, opened 2024-11-20
nrf_rpc: detect reboot of the remote

Damian-Nordic:
Raise an error when the remote sends an init packet with unknown destination group ID after the group has been bound, which indicates a reboot of the peer.

### DIFF
--- a/nrf_rpc/nrf_rpc.c
+++ b/nrf_rpc/nrf_rpc.c
@@ -98,6 +98,7 @@ struct internal_task {
 			const struct nrf_rpc_group *group;
 			bool send_reply;
 			bool signal_groups_init_event;
+			bool signal_reboot;
 		} group_init;
 
 		struct {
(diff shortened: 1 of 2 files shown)


Damian-Nordic:
Converting to draft, as after second thought I'm not sure if reporting a reboot as an error won't cause some issues on existing deployments. Maybe we should add another callback to `nrf_rpc_init` instead?

nrf_rpc/nrf_rpc.c
```diff
@@ -379,6 +380,11 @@ static void internal_tx_handler(void)
 	case NRF_RPC_TASK_GROUP_INIT: {
 		const struct nrf_rpc_group *group = task.group_init.group;
 
+		if (task.group_init.signal_reboot) {
```
alxelax, line 383: Hmm...  Why should device send error instead of notifying application about rebooted remote and reply to remote?

Damian-Nordic: `nrf_rpc_err` does notify the application about the error, and only sends the error in certain cases. But I would say, you probably will like https://github.com/nrfconnect/sdk-nrfxlib/pull/1583 approach more than this one :)

PR #1583, opened 2024-11-20
nrf_rpc: detect reboot of the remote

Damian-Nordic:
Add nrf_rpc_set_bound_handler() that allows to register a global handler that is called when the remote binds to a local nRF RPC group (either for the first time, or after a reboot.

manifest-pr-skip


konradderda (approved):
LGTM

alxelax (commented):
I like this solution more in comparison with the first one.

alxelax (approved):

PR #1593, opened 2024-11-27
nrf_rpc: Add function for single group initialization

Vge0rge:
Adds the function nrf_rpc_init_group to allow initializing an nrf_rpc group individually. This is useful when groups need to be initialized in different boot stages.

nrf_rpc_init now takes into account that a group could could be initialized using the nrf_rpc_init_group and avoids initializiting again.

### DIFF
--- a/nrf_rpc/nrf_rpc.c
+++ b/nrf_rpc/nrf_rpc.c
@@ -411,49 +411,51 @@ static void internal_tx_handler(void)
 	}
 }
 
-static int transport_init(nrf_rpc_tr_receive_handler_t receive_cb)
+static int transport_init_single(nrf_rpc_tr_receive_handler_t receive_cb, const struct nrf_rpc_group *group)
 {
-	int err = 0;
-	void *iter;
-	const struct nrf_rpc_group *group;
+	const struct nrf_rpc_tr *transport = group->transport;
+	struct nrf_rpc_group_data *data = group->data;
+	int err;
 
-	for (NRF_RPC_AUTO_ARR_FOR(iter, group, &nrf_rpc_groups_array,
-				 const struct nrf_rpc_group)) {
-		const struct nrf_rpc_tr *transport = group->transport;
-		struct nrf_rpc_group_data *data = group->data;
+	NRF_RPC_ASSERT(transport != NULL);
 
-		NRF_RPC_ASSERT(transport != NULL);
+	if (group->data->transport_initialized) {
+		return 0;
+	}
 
-		/* Initialize all dependencies of `receive_handler` before calling the transport
-		 * init to avoid possible data race if `receive_handler` was invoked before this
-		 * function was completed. */
-		if (auto_free_rx_buf(transport)) {
-			err = nrf_rpc_os_event_init(&data->decode_done_event);
-			if (err < 0) {
-				continue;
-			}
-		}
-
-		err = transport->api->init(transport, receive_cb, NULL);
-		if (err) {
-			NRF_RPC_ERR("Failed to initialize transport, err: %d", err);
-			continue;
-		}
-
-		group->data->transport_initialized = true;
-
-		if (group->flags & NRF_RPC_FLAGS_INITIATOR) {
-			err = group_init_send(group);
-			if (err) {
-				NRF_RPC_ERR("Failed to send group init packet for group id: %d strid: %s err: %d",
-					data->src_group_id, group->strid, err);
-				continue;
-			}
+	/* Initialize all dependencies of `receive_handler` before calling the transport
+	 * init to avoid possible data race if `receive_handler` was invoked before this
+	 * function was completed. */
+	if (auto_free_rx_buf(transport)) {
+		err = nrf_rpc_os_event_init(&data->decode_done_event);
+		if (err < 0) {
+			return err;
 		}
 	}
 
-	/* Group initialization errors are not propagated to the caller. */
-	err = 0;
+	err = transport->api->init(transport, receive_cb, NULL);
+	if (err) {
+		NRF_RPC_ERR("Failed to initialize transport, err: %d", err);
+		return err;
+	}
+
+	group->data->transport_initialized = true;
+
+	if (group->flags & NRF_RPC_FLAGS_INITIATOR) {
+		err = group_init_send(group);
+		if (err) {
+			NRF_RPC_ERR("Failed to send group init packet for group id: %d strid: %s err: %d",
+				data->src_group_id, group->strid, err);
+			return err;
+		}
+	}
+
+	return 0;
+}
+
+static int groups_init_event_wait(void)
+{
+	int err = 0;
 
 	if (waiting_group_count > 0) {
 		err = nrf_rpc_os_event_wait(&groups_init_event, CONFIG_NRF_RPC_GROUP_INIT_WAIT_TIME);
@@ -463,6 +465,27 @@ static int transport_init(nrf_rpc_tr_receive_handler_t receive_cb)
 	}
 
 	return err;
+}
+
+static int transport_init_all(nrf_rpc_tr_receive_handler_t receive_cb)
+{
+	void *iter;
+	const struct nrf_rpc_group *group;
+
+	for (NRF_RPC_AUTO_ARR_FOR(iter, group, &nrf_rpc_groups_array,
+				 const struct nrf_rpc_group)) {
+
+		transport_init_single(receive_cb, group);
+	}
+
+	/* Group initialization errors are not propagated to the caller. */
+	return groups_init_event_wait();
+}
+
+static void default_err_handler(const struct nrf_rpc_err_report *report)
+{
+	NRF_RPC_ERR("nRF RPC error %d ocurred. See nRF RPC logs for more details", report->code);
+	nrf_rpc_os_fatal_error();
 }
 
 /* ======================== Receiving Packets ======================== */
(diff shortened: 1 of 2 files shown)


doki-nordic (changes requested):

nrf_rpc/nrf_rpc.c
```diff
@@ -465,6 +467,27 @@ static int transport_init(nrf_rpc_tr_receive_handler_t receive_cb)
...
+	}
+
+	/* Group initialization errors are not propagated to the caller. */
+	return groups_init_event_wait();
+}
+
+static void default_err_handler(const struct nrf_rpc_err_report *report)
+{
+	NRF_RPC_ERR("nRF RPC error %d ocurred. See nRF RPC logs for more details", report->code);
+	k_oops();
```
doki-nordic, line 488: This should be OS-independent code. k_oops() is a Zephyr function. Please pass it through OS porting layer.

doki-nordic (changes requested):

nrf_rpc/nrf_rpc.c
```diff
@@ -1144,20 +1162,63 @@ int nrf_rpc_init(nrf_rpc_err_handler_t err_handler)
...
+
+	NRF_RPC_ASSERT(group_count > 0);
+	if (initialized_group_count == group_count - 1) {
+		return groups_init_event_wait();
+	}
+
+	return 0;
+}
+
+int nrf_rpc_init(nrf_rpc_err_handler_t err_handler)
```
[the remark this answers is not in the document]
Vge0rge: Good catch, I changed that.

nrf_rpc/nrf_rpc.c
Vge0rge: Regarding this, do you refer to this OS porting layer do you refer to the nrf_rpc one?
So what you are proposing is to add another function here:
https://github.com/nrfconnect/sdk-nrf/blob/main/subsys/nrf_rpc/nrf_rpc_os.c

Which handles the error case?

nrf_rpc/nrf_rpc.c
```diff
@@ -411,49 +411,51 @@ static void internal_tx_handler(void)
...
+	int err;
 
-	for (NRF_RPC_AUTO_ARR_FOR(iter, group, &nrf_rpc_groups_array,
-				 const struct nrf_rpc_group)) {
-		const struct nrf_rpc_tr *transport = group->transport;
-		struct nrf_rpc_group_data *data = group->data;
+	NRF_RPC_ASSERT(transport != NULL);
 
-		NRF_RPC_ASSERT(transport != NULL);
+	if(group->data->transport_initialized) {
```
pdunaj, line 422: space missing
`s/if(/if (`

nrf_rpc/nrf_rpc.c
```diff
@@ -465,6 +467,27 @@ static int transport_init(nrf_rpc_tr_receive_handler_t receive_cb)
...
 
+static int transport_init_all(nrf_rpc_tr_receive_handler_t receive_cb)
+{
+	void *iter;
+	const struct nrf_rpc_group *group;
+
+	for (NRF_RPC_AUTO_ARR_FOR(iter, group, &nrf_rpc_groups_array,
+				 const struct nrf_rpc_group)) {
+
+		transport_init_single(receive_cb, group);
```
pdunaj, line 478: error is not checked

nrf_rpc/nrf_rpc.c
```diff
@@ -1144,20 +1162,67 @@ int nrf_rpc_init(nrf_rpc_err_handler_t err_handler)
 		return err;
 	}
 
-	for (i = 0; i < CONFIG_NRF_RPC_CMD_CTX_POOL_SIZE; i++) {
+	for (int i = 0; i < CONFIG_NRF_RPC_CMD_CTX_POOL_SIZE; i++) {
```
pdunaj, line 1165: s/int/size_t

nrf_rpc/nrf_rpc.c
```diff
@@ -1144,20 +1162,67 @@ int nrf_rpc_init(nrf_rpc_err_handler_t err_handler)
...
+
+	is_initialized = true;
+
+	return 0;
+}
+
+int nrf_rpc_init_group(const struct nrf_rpc_group *group)
+{
+	int err = nrf_rpc_prepare_init();
+	if(err < 0) {
```
pdunaj, line 1183: missing space `if (`

nrf_rpc/nrf_rpc.c
```diff
@@ -1144,20 +1162,67 @@ int nrf_rpc_init(nrf_rpc_err_handler_t err_handler)
...
+
+	/* Everything is initialized, nothing to do here */
+	if (group_count > 0 && group_count == initialized_group_count) {
+		return 0;
+	}
+
+	NRF_RPC_DBG("Initializing nRF RPC module");
+
+	err = nrf_rpc_prepare_init();
+	if(err < 0) {
```
pdunaj, line 1213: `if (`

pdunaj (changes requested):

pdunaj:
Hi @Vge0rge , please just fix the error handling and I will approve and merge.

nrf_rpc/nrf_rpc.c
Vge0rge: This is intentional though, because I saw in the previous implementation they didn't return an error when a single initialization failed.
The logic here:
    https://github.com/nrfconnect/sdk-nrfxlib/blob/49299379e19fb3b2d7bc8cc1b0bd2803c93a6878/nrf_rpc/nrf_rpc.c#L420

Should be the same with the logic in the function transport_init_single basically. 

The error that I return in this function is the logic which exists here:
https://github.com/nrfconnect/sdk-nrfxlib/blob/49299379e19fb3b2d7bc8cc1b0bd2803c93a6878/nrf_rpc/nrf_rpc.c#L458

nrf_rpc/nrf_rpc.c
Vge0rge: Done

nrf_rpc/nrf_rpc.c
Vge0rge: Done

nrf_rpc/nrf_rpc.c
Vge0rge: Done

nrf_rpc/nrf_rpc.c
Vge0rge: Done

nrf_rpc/nrf_rpc.c
Vge0rge: Do you have an opinion on this Pawel? I can  do what I am describing here if you have the same understanding as me

pdunaj (approved):

nrf_rpc/nrf_rpc.c
```diff
@@ -465,6 +467,27 @@ static int transport_init(nrf_rpc_tr_receive_handler_t receive_cb)
...
+				 const struct nrf_rpc_group)) {
+
+		transport_init_single(receive_cb, group);
+	}
+
+	/* Group initialization errors are not propagated to the caller. */
+	return groups_init_event_wait();
+}
+
+static void default_err_handler(const struct nrf_rpc_err_report *report)
```
Damian-Nordic, line 485: IMHO, this is a bit inconsistent that when you want to have groups auto-initialized then you can register a custom error handler and if you want to init groups manually then you have to use the default error handler.

Would it make sense to create a public API such as `nrf_rpc_init_core` which would only initialize the nRF PRC without any groups (and which would take the error handler). `nrf_rpc_init_group` would then validate that the core has already been initialized.

nrf_rpc/nrf_rpc.c
```diff
@@ -1144,20 +1162,66 @@ int nrf_rpc_init(nrf_rpc_err_handler_t err_handler)
...
+		return err;
+	}
+
+	err = transport_init_single(receive_handler, group);
+	if (err < 0) {
+		return err;
+	}
+
+	NRF_RPC_ASSERT(group_count > 0);
+	if (initialized_group_count == group_count - 1) {
```
Damian-Nordic, line 1193: Shouldn't you use `waiting_group_count`?  If there are groups without `NRF_RPC_FLAGS_WAIT_ON_INIT` flag, the wait below will hang indefinitely. By the way, it's non-obvious that the function waits only if there's only one group that has not been initialized so far, so this should be documented or there should be a follow up filed to make the group initialization less coupled (so that each group can only wait for its own initialization completion).

nrf_rpc/nrf_rpc.c
```diff
@@ -1144,20 +1162,66 @@ int nrf_rpc_init(nrf_rpc_err_handler_t err_handler)
...
+	global_err_handler = default_err_handler;
+
+	is_initialized = true;
+
+	return 0;
+}
+
+int nrf_rpc_init_group(const struct nrf_rpc_group *group)
+{
+	int err = nrf_rpc_prepare_init();
```
Damian-Nordic, line 1182: It seems that this code is not thread-safe. Is there an assumption that `nrf_rpc_init_group` must be called from the main thread only? If so, I think we should document it. Otherwise, `nrf_rpc_prepare_init` may end up being executed from two threads.

Vge0rge: You are right that this  is not thread safe, but is it not the case that the whole nrf_pc is not thread safe? The variable `is_initialized` was static before, I didn't change that. I am not too familiar with the library but it I see many static objects being manipulated so I thought that it is not thread safe. Am I missing something?

nrf_rpc/nrf_rpc.c
Vge0rge: What I was trying to do here to here is to call the event_wait function only for the last group which was getting initialized. The function `groups_init_event_wait` checks for waiting threads so it will actually not hang. But on the other hand now that I look at it again there is a bug in this indeed. The `initialized_group_count` only increases for the threads which have the `NRF_RPC_FLAGS_WAIT_ON_INIT` flag enabled, so if there is one thread that doesn't then the wait will never get called. I will have to refactor this.

Vge0rge:
Converted to draft until I finish the refactoring to signal that this should get merged as is.

Vge0rge:
Another approach will be followed for this so closing this.

PR #1628, opened 2025-01-07
nrf_modem: doc: Change GNSS interface name

tokangas:
Changed GNSS interface name in documentation from `Global Satellite Navigation System` to `GNSS`. The name of the interface in modemlib is `GNSS` and in general all documentation refers to `GNSS interface` or `GNSS API`.

### DIFF
--- a/nrf_modem/doc/gnss_interface.rst
+++ b/nrf_modem/doc/gnss_interface.rst
@@ -1,8 +1,8 @@
 .. _nrf_modem_gnss:
 .. _gnss_interface:
 
-Global Navigation Satellite System
-##################################
+GNSS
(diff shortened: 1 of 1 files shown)


eivindj-nordic:
Looking at where it is used in documentation I think it should either be changed to `GNSS interface`, or we should update the documentation where it is used to add `interface` after the reference. I don't have a strong opinion of what is best.

rlubos (approved):

tokangas:
We had a call with Emanuele and agreed that we'll go with "GNSS". I'll update the references in the SDK.

divipillai (approved):

PR #1660, opened 2025-02-21
mpsl|softdevice_controller: add SoC-specific (D)PPI defines

masz-nordic:


gordonklaus (approved):

softdevice_controller/include/sdc_soc.h
```diff
@@ -29,22 +29,33 @@ extern "C" {
 #include <stdint.h>
 #include <stdbool.h>
 
+#define SDC_NRF52_PPI_CHANNELS_USED_MASK       (0xfffe0000)
```
seko-nordic, line 32: I think this change will be overwritten by the next deploy of SDC. It should be done here instead https://projecttools.nordicsemi.no/bitbucket/projects/DRGN/repos/dragoon/browse/stack/projects/sdc/include/sdc_soc.h

seko-nordic (changes requested):

PR #1681, opened 2025-03-18
nrf_rpc: fix another case where transport_initialized is false

Damian-Nordic:
Fix one missed scenario in which transport_initialized flag may be false because of the thread running nrf_rpc_init() being switched out by the scheduler before setting this flag.

In such a case, if there's an init packet received, responding to this packet might trigger an error.

manifest-pr-skip


Damian-Nordic:
@doki-nordic Could you please review?

Damian-Nordic:
@doki-nordic ping

doki-nordic (approved):

bpienk (approved):

maciejbaczmanski (approved):

PR #1732, opened 2025-04-17
openthread: Remove dependency to Zephyr NET L2 OpenThread

ArekBalysNordic:
Rely on CONFIG_OPENTHREAD instead of CONFIG_NET_L2_OPENTHREAD to support a different networking approach than Zephyr's L2 layer.

This allows to use of the nrf5 radio driver directly in the OpenThread platform as a simple alternative to Zephyr L2 networking.


adigie:
Done in #1751.

PR #1734, opened 2025-04-25
nrf_modem: add invalid_sec_tag value

maxd-nordic:
Define a placeholder value for the sec tag type.

### DIFF
--- a/nrf_modem/include/nrf_socket.h
+++ b/nrf_modem/include/nrf_socket.h
@@ -780,6 +780,9 @@ struct nrf_ifaddrs {
  */
 typedef uint32_t nrf_sec_tag_t;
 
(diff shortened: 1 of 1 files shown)


rlubos (approved):

MirkoCovizzi (changes requested):
This should be added from within `nrf_modem` and then exported here.

MirkoCovizzi:
@maxd-nordic Have you seen this already? https://github.com/nrfconnect/sdk-nrfxlib/blob/b7f26d1b1fe2205a8e207aa34c07ff9dcf86a5f5/nrf_modem/include/nrf_socket.h#L568 The accepted values have a documented range.

PR #1764, opened 2025-06-06
nrf_rpc: implement timeout for response

Damian-Nordic:
Introduce necessary changes to support timing out waiting for a command response:
1. Handle nrf_rpc_os_msg_get() returning null data pointer, indicating that the response has not arrived within the required maximum time.
2. Introduce context mutex to protect all context members. This is necessary because now the response may arrive after the initiating thread has already released the context.

Note that 2 may not sufficient solution to address another scenario where the context is freed by the initiator but is then reused before a delayed response arrives - to solve this, we likely need to introduce the concept of session ID or conversation ID to detect outdated packets.

manifest-pr-skip

### DIFF
--- a/nrf_rpc/CHANGELOG.rst
+++ b/nrf_rpc/CHANGELOG.rst
@@ -9,6 +9,18 @@ Changelog
 
 All the notable changes to this project are documented on this page.
 
+nRF Connect SDK v3.0.99
+***********************
+
+Added
+=====
+
+* Added a timeout feature for awaiting command responses.
+  This change required making the following changes to the nRF RPC OS abstraction layer:
+
+  * Introduced the :c:struct:`nrf_rpc_os_mutex` structure to protect the nRF RPC thread-local context.
+  * Updated the :c:func:`nrf_rpc_os_msg_get` function to unlock the thread-local context mutex and to allow returning no data from the function if a timeout occurs.
+
 nRF Connect SDK v2.8.0
 **********************
 
--- a/nrf_rpc/nrf_rpc.c
+++ b/nrf_rpc/nrf_rpc.c
@@ -52,6 +52,8 @@ struct nrf_rpc_cmd_ctx {
 				    */
 	nrf_rpc_handler_t handler; /* Response handler provided be the user. */
 	void *handler_data;	   /* Pointer for the response handler. */
+	struct nrf_rpc_os_mutex mutex; /* Mutex for protecting all context
+					* members. */
 	struct nrf_rpc_os_msg recv_msg;
 				   /* Message passing between transport
 				    * receive callback and a thread that waits
@@ -187,6 +189,7 @@ static struct nrf_rpc_cmd_ctx *cmd_ctx_alloc(void)
 	NRF_RPC_ASSERT(index < CONFIG_NRF_RPC_CMD_CTX_POOL_SIZE);
 
 	ctx = &cmd_ctx_pool[index];
+	nrf_rpc_os_mutex_lock(&ctx->mutex);
 	ctx->handler = NULL;
 	ctx->remote_id = NRF_RPC_ID_UNKNOWN;
 	ctx->use_count = 1;
@@ -200,6 +203,7 @@ static struct nrf_rpc_cmd_ctx *cmd_ctx_alloc(void)
 
 static void cmd_ctx_free(struct nrf_rpc_cmd_ctx *ctx)
 {
+	nrf_rpc_os_mutex_unlock(&ctx->mutex);
 	nrf_rpc_os_tls_set(NULL);
 	nrf_rpc_os_ctx_pool_release(ctx->id);
 }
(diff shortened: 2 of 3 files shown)


Damian-Nordic:
@doki-nordic @grochu Friendly reminder to take a look.

nrf_rpc/template/nrf_rpc_os_tmpl.h
```diff
@@ -94,6 +94,26 @@ void nrf_rpc_os_event_set(struct nrf_rpc_os_event *event);
  */
 int nrf_rpc_os_event_wait(struct nrf_rpc_os_event *event, int32_t timeout);
 
+/** @brief Initialize mutex structure.
+ *
+ * @param mutex Pointer to mutex structure.
+ *
+ * @return    0 on success or negative error code.
+ */
+int nrf_rpc_os_mutex_init(struct nrf_rpc_os_mutex *mutex);
```
grochu, line 106: Please add a provisional structure definition for `nrf_rpc_os_mutex` in the template header file, like it is done for e.g. `nrf_rpc_os_event`.

nrf_rpc/nrf_rpc.c
```diff
@@ -907,12 +923,14 @@ static void wait_for_response(const struct nrf_rpc_group *group, struct nrf_rpc_
 	NRF_RPC_DBG("Waiting for a response");
 
 	do {
-		nrf_rpc_os_msg_get(&cmd_ctx->recv_msg, &packet, &len);
+		nrf_rpc_os_msg_get(&cmd_ctx->recv_msg, &cmd_ctx->mutex, &packet, &len);
 
-		NRF_RPC_ASSERT(packet != NULL);
+		if (packet == NULL) {
```
grochu, line 928: Could you clarify in the `nrf_rpc_os_msg_get` documentation, that returning NULL is used for signaling the response timeout?

doki-nordic:
I started reviewing this PR, but since I have limited time to review it, I did not finalized it. I have concerns about mutex locking logic. Do we really need to lock it in context alloc function? When locking and unlocking logic is spread across multiple functions and multiple files, there is higher risk of edge cases or time races that may lead to dead lock. I am concern about context reusing functionality. Can you do simple context reusing test on this PR? See docs for some details about it: https://docs.nordicsemi.com/bundle/ncs-latest/page/nrfxlib/nrf_rpc/doc/architecture.html#id5 . This will happen, for example, when the local side sends a command and waits for a response, next the remote send another command back to local within the command handler context. In this case, context on local side should be reused.

Damian-Nordic:
> I started reviewing this PR, but since I have limited time to review it, I did not finalized it. I have concerns about mutex locking logic. Do we really need to lock it in context alloc function? When locking and unlocking logic is spread across multiple functions and multiple files, there is higher risk of edge cases or time races that may lead to dead lock. I am concern about context reusing functionality. Can you do simple context reusing test on this PR? See docs for some details about it: https://docs.nordicsemi.com/bundle/ncs-latest/page/nrfxlib/nrf_rpc/doc/architecture.html#id5 . This will happen, for example, when the local side sends a command and waits for a response, next the remote send another command back to local within the command handler context. In this case, context on local side should be reused.

`cmd_ctx_alloc` may seem counterintuitive but I didn't find a better place to do the locking:
1. `cmd_ctx_alloc` already modifies the context to initialize it, so it must lock the mutex in order to prevent another thread from using a partially initialized context when e.g. a delayed response arrives. If we wanted to do lock/unlock every time in a single function, we would need to do this in many functions, such as `cmd_ctx_alloc`, `cmd_ctx_reserve`, `nrf_rpc_cmd_common`. I thought this would be less efficient.
2. `cmd_ctx_alloc` is called only when a new context is allocated - in the case of context reusing, this won't be called again. If we moved locking to e.g. `nrf_rpc_cmd_common`, we might be going to lock an already locked mutex, so we would need to assume the mutex is reentrant or add an extra check.

I'm happy to document it better, or rework this if you prefer locking the mutex in individual functions, but I thought doing this the current way actually simplifies the design - you always run nRF RPC code with the context mutex held and unlock it only when waiting for the peer or when releasing the context.

And yeah, I can run a test like that :)

Damian-Nordic:
@grochu I addressed your comments, please take another look.
@doki-nordic I tested this code with the context reusing (implemented UDP callback which retrieves extra information about the received UDP packet from the remote device) and it works as expected.
Let me know if you have time for a call so I can provide the rationale for the current design.

grochu (approved):
Approving to unblock development. Let's discuss the context locking with mutex later in another context.

bpienk (approved):

nrf_rpc/CHANGELOG.rst
```diff
@@ -9,6 +9,18 @@ Changelog
...
+***********************
+
+Changes
+=======
+
+* Added support for timing out waiting for a command response.
+  This required making the following changes to the nRF RPC OS abstraction layer:
+
+  * Added :c:struct:`nrf_rpc_os_mutex` structure to protect the nRF RPC thread-local context.
+  * Modified :c:func:`nrf_rpc_os_msg_get` function to unlock the thread-local context mutex, and to allow returning no data from the function if a timeout occurs.
```
annwoj: First time looking at this changelog, but seeing other examples the content probably matches the "added" section better:
```suggestion
Added
=====

* Added a timeout feature for awaiting command responses.
  This change required making the following changes to the nRF RPC OS abstraction layer:

  * Introduced the :c:struct:`nrf_rpc_os_mutex` structure to protect the nRF RPC thread-local context.
  * Updated the :c:func:`nrf_rpc_os_msg_get` function to unlock the thread-local context mutex and to allow returning no data from the function if a timeout occurs.
```

carlescufi (approved):

PR #1765, opened 2025-06-06
nfc: align the T2T and T4T Kconfig options with DTS node status

kapi-no:
Aligned the availability of the NFC_T2T_NRFXLIB and NFC_T4T_NRFXLIB Kconfig options with the status of the nfct DTS node.

### DIFF
--- a/nfc/Kconfig
+++ b/nfc/Kconfig
@@ -1,11 +1,13 @@
 #
-# Copyright (c) 2018 Nordic Semiconductor
+# Copyright (c) 2018-2025 Nordic Semiconductor
 #
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
 config NFC_T2T_NRFXLIB
 	bool "Enable NFC Type 2 Tag library"
+	depends on HAS_HW_NRF_NFCT
 
 config NFC_T4T_NRFXLIB
 	bool "Enable NFC Type 4 Tag library"
+	depends on HAS_HW_NRF_NFCT


grochu:
Thanks for providing this PR! We still haven't managed to resolve it. Just a though from me -  how about moving the dependency to the CONFIG_NFC_PLATFORM option in sdk-nrf? Both T2T and T4T require some platform implementation for the runtime environment they are used in. Like that we'd avoid dependency expansion in the nrfxlib, which in theory could be used outside of nrf-sdk.

kapi-no:
> Thanks for providing this PR! We still haven't managed to resolve it. Just a though from me - how about moving the dependency to the CONFIG_NFC_PLATFORM option in sdk-nrf? Both T2T and T4T require some platform implementation for the runtime environment they are used in. Like that we'd avoid dependency expansion in the nrfxlib, which in theory could be used outside of nrf-sdk.

I wanted to put this `depends on` statement initially in the NFC platform Kconfig:

https://github.com/nrfconnect/sdk-nrf/blob/main/subsys/nfc/lib/Kconfig

However, the current KConfig definition is written in such a way that it will fail silently for the user:

```
config NFC_PLATFORM
	bool "Common NFC configuration"
	default y if NFC_T2T_NRFXLIB || NFC_T4T_NRFXLIB
	select NRFX_NFCT
	select CLOCK_CONTROL
	help
	  Enable common configuration for the NFC
```

If you were to add `depends on HAS_HW_NRF_NFCT` in the `NFC_PLATFORM` Kconfig, you won't get a warning about unmet dependency (HAS_HW_NRF_NFCT) here because the NFC platform is not explicitly selected but is rather enabled by default. I think this approach would confuse users. 

If you want to set it up like this, I can do that. The most important thing is that the build process for a project will fail with either approach.

Also, there is nothing stopping us from moving the NFC T2T/T4T library Kconfig to sdk-nrf. You can follow the approach used by the SDC team for the SDC library.

grochu:
> > Thanks for providing this PR! We still haven't managed to resolve it. Just a though from me - how about moving the dependency to the CONFIG_NFC_PLATFORM option in sdk-nrf? Both T2T and T4T require some platform implementation for the runtime environment they are used in. Like that we'd avoid dependency expansion in the nrfxlib, which in theory could be used outside of nrf-sdk.
> 
> I wanted to put this `depends on` statement initially in the NFC platform Kconfig:
> 
> https://github.com/nrfconnect/sdk-nrf/blob/main/subsys/nfc/lib/Kconfig
> 
> However, the current KConfig definition is written in such a way that it will fail silently for the user:
> 
> ```
> config NFC_PLATFORM
> 	bool "Common NFC configuration"
> 	default y if NFC_T2T_NRFXLIB || NFC_T4T_NRFXLIB
> 	select NRFX_NFCT
> 	select CLOCK_CONTROL
> 	help
> 	  Enable common configuration for the NFC
> ```
> 
> If you were to add `depends on HAS_HW_NRF_NFCT` in the `NFC_PLATFORM` Kconfig, you won't get a warning about unmet dependency (HAS_HW_NRF_NFCT) here because the NFC platform is not explicitly selected but is rather enabled by default. I think this approach would confuse users.
> 
> If you want to set it up like this, I can do that. The most important thing is that the build process for a project will fail with either approach.
> 
> Also, there is nothing stopping us from moving the NFC T2T/T4T library Kconfig to sdk-nrf. You can follow the approach used by the SDC team for the SDC library.

I see, I didn't take this into account. NFC configurations need some general refactoring, we'll do this at some point, when there is capacity. For now, your approach looks good.

grochu (approved):

maje-emb (approved):

kapi-no:
Pure rebase. The PR file diff has not changed

PR #1829, opened 2025-08-11
.github: test-spec: add LL label

nordic-piks:
Run LL tests for changes at softperipheral.


nordic-piks:
My mistake, LL downstream does not have required HW setup to test this at integration level (still will be tested at night)

PR #1831, opened 2025-08-12
Update sdc_hci_data_put documentation

asilz:
The documentation currently states that the first byte of the buffer is the handle when the first 12 bits are the handle. This caused me a lot of confusion as I assumed there is a different handle that needs to be placed before the ACL data packet. The documentation also does not reference which section of the spec the format is specified.


seko-nordic (changes requested):
This need to be updated in dragoon repo, otherwise the next push here will overwrite it

PR #1866, opened 2025-09-30
Add nRF71 Wi-Fi interface files

krish2718:
These will be used to communicate with nRF71 Wi-Fi domain.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,9 +6,8 @@
 include(common.cmake)
 
 set(NRFXLIB_DIR ${CMAKE_CURRENT_LIST_DIR} CACHE PATH "NRFXLIB root directory")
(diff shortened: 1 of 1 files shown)


sachinthegreen (approved):

nrf71_wifi/Kconfig
```diff
@@ -0,0 +1,12 @@
+#
+# Copyright (c) 2025 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+menu "nrf71_Wi-Fi firmware interface library"
```
tejlmand: why do you have a menu here when the only entry inside that menu is a promptless symbol ?

krish2718: It's a leftover when there were more symbols, will remove it.

PR #2066, opened 2026-04-07
doc: Update doc versions in nrfxlib for 3.3.0

richabp:
Updated doc versions in nrfxlib for 3.3.0

NCSDK-38562

### DIFF
--- a/mpsl/CHANGELOG.rst
+++ b/mpsl/CHANGELOG.rst
@@ -10,8 +10,8 @@ Changelog
 
 All the notable changes to this project are documented on this page.
 
-Main branch
-***********
+nRF Connect SDK v3.3.0
+**********************
 
 Added
 =====
--- a/nrf_802154/doc/CHANGELOG.rst
+++ b/nrf_802154/doc/CHANGELOG.rst
@@ -10,8 +10,8 @@ Changelog
 All notable changes to this project are documented in this file.
 See also :ref:`nrf_802154_limitations` for permanent limitations.
 
-Main branch - nRF 802.15.4 Radio Driver
-***************************************
+nRF Connect SDK v3.3.0 - nRF 802.15.4 Radio Driver
+**************************************************
 
 Added
 =====
--- a/nrf_fuel_gauge/CHANGELOG.rst
+++ b/nrf_fuel_gauge/CHANGELOG.rst
@@ -9,8 +9,8 @@ Changelog
 
 All notable changes to this project are documented on this page.
 
-Main branch
-***********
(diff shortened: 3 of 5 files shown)


rlubos (approved):

rugeGerritsen (approved):

umapraseeda (approved):

nrf_modem/doc/architecture.rst
```diff
@@ -19,6 +19,10 @@ The following figure shows a simplified Modem library architecture:
 Shared memory configuration
 ***************************
 
+.. note::
+
+   |pm_deprecation_note|
```
nordic-mare: I don't think this note makes sense here for this release. For NCS 3.3.0 libmodem will not work without Partition Manager so this will probably only confuse the customers. What do you think @lucasvdp @kacperradoszewski ?

FrancescoSer (approved):

nrf_modem/doc/architecture.rst
richabp: cc: @divipillai

divipillai: We have similar deprecation note for nrf_modem_lib in the NCS repo - https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/libraries/modem/nrf_modem_lib/index.html, which mentions that PM is in the process of being deprecated. Since this will be deprecated and removed in the future, shouldn't we keep this note?

kacperradoszewski: Yeah I agree, we should not encourage users to use the Modem Library without PM in this release.

nordic-mare: Adding this note to main makes sense, adding it to this release does not. PM will not be deprecated for this release and quite a few features won't work without it, the Modem Library being one of them.

nordic-mare (approved):

divipillai (approved):

nrf_modem/doc/architecture.rst
lucasvdp: We can still announce that it will be deprecated at some point even if we cannot tell them yet how to do it.
If it is needed for consistency or other reasons, I am oké with adding it now, otherwise we can add it when we update the documentation on how to run without PM.

MarGasiorek (approved):

nordic-auko (approved):

nrf_modem/doc/architecture.rst
divipillai: Could leave it as is for above reasons.

kacperradoszewski (approved):

PR #2095, opened 2026-04-29
[DNM] mpsl: include mdk_config in hwres header

mstasiaknordic:
Target symbol definition has been move from CMake to MDK config header. Include the header to get access to defined symbols.

### DIFF
--- a/mpsl/include/mpsl_hwres.h
+++ b/mpsl/include/mpsl_hwres.h
@@ -17,6 +17,9 @@
 #ifndef MPSL_HWRES_H__
 #define MPSL_HWRES_H__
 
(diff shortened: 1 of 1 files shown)


mpsl/include/mpsl_hwres.h
```diff
@@ -17,6 +17,9 @@
 #ifndef MPSL_HWRES_H__
 #define MPSL_HWRES_H__
 
+/* Symbols passed to nrfx-based MDK. */
+#include <mdk_config.h>
```
ppryga-nordic, line 21: We can't incude this file. It will break our internal builds. Also customers using the MPSL without the NCS/Zephyr will be affected.

PR #2159, opened 2026-06-12
nrf_rpc: Reset decode_done_event so event_wait always works

inductivekickback:
If the transport is auto_free_rx_buf then the receive_handler must not return until after the packet has been decoded. Without this patch some code path sets the decode_done_event without a corresponding nrf_rpc_os_event_wait (not necessarily a bug) and then the event_wait stops working as intended.

Depends on nrfconnect/sdk-nrf/pull/29541


MarekPorwisz (approved):

kderda (approved):
LGTM, thanks

PR #2194, opened 2026-07-14
zephyr: Move files from sdk-nrf

nordicjm:
Moves files from sdk-nrf to here so this can be used as an actual zephyr module

manifest-pr-skip

### DIFF
--- a/zephyr/module.yml
+++ b/zephyr/module.yml
@@ -1,5 +1,5 @@
 build:
-  cmake-ext: True
+  cmake: zephyr
   kconfig: Kconfig.nrfxlib
   depends:
   - hal_nordic
--- a/nrf_802154/sl/platform/zephyr/nrf_802154_hp_timer.c
+++ b/nrf_802154/sl/platform/zephyr/nrf_802154_hp_timer.c
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2018 - 2022 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/**
+ * @file
+ *   This file contains implementation of the nRF 802.15.4 high precision timer abstraction.
+ *
+ * This implementation is built on top of the TIMER peripheral.
+ * If SoftDevice RAAL is in use the TIMER peripheral is shared between RAAL and this module.
+ *
+ */
+
+#include "platform/nrf_802154_hp_timer.h"
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "nrfx.h"
+#include "hal/nrf_timer.h"
+
+#include "nrf_802154_sl_periphs.h"
+
+/**@brief Timer instance. */
+#define TIMER                 NRF_802154_HIGH_PRECISION_TIMER_INSTANCE
+
+/**@brief Timer compare channel definitions. */
+#define TIMER_CC_CAPTURE      NRF_TIMER_CC_CHANNEL1
+#define TIMER_CC_CAPTURE_TASK NRF_TIMER_TASK_CAPTURE1
+
+#define TIMER_CC_SYNC         NRF_TIMER_CC_CHANNEL2
+#define TIMER_CC_SYNC_TASK    NRF_TIMER_TASK_CAPTURE2
+#define TIMER_CC_SYNC_EVENT   NRF_TIMER_EVENT_COMPARE2
+#define TIMER_CC_SYNC_INT     NRF_TIMER_INT_COMPARE2_MASK
+
+#define TIMER_CC_EVT          NRF_TIMER_CC_CHANNEL3
+#define TIMER_CC_EVT_TASK     NRF_TIMER_TASK_CAPTURE3
+#define TIMER_CC_EVT_INT      NRF_TIMER_INT_COMPARE3_MASK
+
+/**@brief Unexpected value in the sync compare channel. */
+static uint32_t m_unexpected_sync;
+
+/**@brief Get current time on the Timer. */
+static inline uint32_t timer_time_get(void)
+{
+	nrf_timer_task_trigger(TIMER, TIMER_CC_CAPTURE_TASK);
+	return nrf_timer_cc_get(TIMER, TIMER_CC_CAPTURE);
+}
+
+void nrf_802154_hp_timer_init(void)
+{
+	nrf_timer_bit_width_set(TIMER, NRF_TIMER_BIT_WIDTH_32);
+	nrf_timer_prescaler_set(TIMER, NRF_TIMER_FREQ_1MHz);
+	nrf_timer_mode_set(TIMER, NRF_TIMER_MODE_TIMER);
+}
+
+void nrf_802154_hp_timer_deinit(void)
+{
+	nrf_timer_task_trigger(TIMER, NRF_TIMER_TASK_SHUTDOWN);
+}
+
+void nrf_802154_hp_timer_start(void)
+{
+	nrf_timer_task_trigger(TIMER, NRF_TIMER_TASK_START);
+}
+
+void nrf_802154_hp_timer_stop(void)
+{
+	nrf_timer_task_trigger(TIMER, NRF_TIMER_TASK_SHUTDOWN);
+}
+
+uint32_t nrf_802154_hp_timer_sync_task_get(void)
+{
+	return nrf_timer_task_address_get(TIMER, TIMER_CC_SYNC_TASK);
+}
+
+void nrf_802154_hp_timer_sync_prepare(void)
+{
+	uint32_t past_time = timer_time_get() - 1;
+
+	m_unexpected_sync = past_time;
+	nrf_timer_cc_set(TIMER, TIMER_CC_SYNC, past_time);
+}
+
+bool nrf_802154_hp_timer_sync_time_get(uint32_t *p_timestamp)
+{
+	bool result = false;
+	uint32_t sync_time = nrf_timer_cc_get(TIMER, TIMER_CC_SYNC);
+
+	assert(p_timestamp != NULL);
+
+	if (sync_time != m_unexpected_sync) {
+		*p_timestamp = sync_time;
+		result = true;
+	}
+
+	return result;
+}
+
+uint32_t nrf_802154_hp_timer_timestamp_task_get(void)
+{
+	return nrf_timer_task_address_get(TIMER, TIMER_CC_EVT_TASK);
+}
+
+uint32_t nrf_802154_hp_timer_timestamp_get(void)
+{
+	return nrf_timer_cc_get(TIMER, TIMER_CC_EVT);
+}
+
+uint32_t nrf_802154_hp_timer_current_time_get(void)
+{
+	return timer_time_get();
+}
--- a/nrf_802154/sl/platform/zephyr/nrf_802154_platform_sl_lptimer_grtc.c
+++ b/nrf_802154/sl/platform/zephyr/nrf_802154_platform_sl_lptimer_grtc.c
@@ -1,0 +1,233 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include "platform/nrf_802154_platform_sl_lptimer.h"
+#include "nrf_802154_platform_sl_lptimer_grtc_hw_task.h"
+
+#include <assert.h>
+#include <zephyr/sys/atomic.h>
+#include <zephyr/drivers/timer/nrf_grtc_timer.h>
+
+#include <haly/nrfy_dppi.h>
+#include <haly/nrfy_grtc.h>
+
+#include "nrf_802154_sl_config.h"
+#include "nrf_802154_sl_atomics.h"
+#include "nrf_802154_sl_utils.h"
+#include "timer/nrf_802154_timer_coord.h"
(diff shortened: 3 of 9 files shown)


ppryga-nordic (approved):

MarGasiorek (commented):
Could you elaborate little more about the changes?
Are the files moved without modification?
What is the purposes of the changes? In order to make sdk-nrfxlib work as a standalone Zephyr module? Or to
keeping required module files in sdk-nrf creates coupling and prevents clean reuse of sdk-nrfxlib as an independent module?

ppryga-nordic:
> Could you elaborate little more about the changes? What is the purposes of the changes? In order to make sdk-nrfxlib work as a standalone Zephyr module? Or to keeping required module files in sdk-nrf creates coupling and prevents clean reuse of sdk-nrfxlib as an independent module?

The PR moved the nRF 802.15.4 SL Zephyr platform glue from sdk-nrf and switches the module from cmake-ext: True to cmake: zephyr. 

Previously sdk-nrf had to provide the nrfxlib CMake integration and host these sources, because nrfxlib only shipped the driver/SL and platform API headers. This makes nrfxlib self-contained: it ships both the sources and the Zephyr build entry point (nrf_802154/zephyr/CMakeLists.txt). Once west points at this revision, sdk-nrf can drop its nrfxlib module overlay entirely.

Outcome of the module.yml change:
- Before: nrfxlib declared itself as a cmake-ext module, so sdk-nrf had to provide the actual CMake entry point (ZEPHYR_NRFXLIB_CMAKE_DIR) and host the platform sources.
- After: nrfxlib is a standard Zephyr module. Zephyr loads nrfxlib/zephyr/CMakeLists.txt directly, which pulls in the 802.15.4
  zephyr integration from nrf_802154/zephyr/CMakeLists.txt.

With `cmake: zephyr`, nrfxlib becomes a standard Zephyr module with its own build entry point. Downstream manifests can consume it directly, or provide a thin integration layer that selects which nrfxlib components to use and which to replace — the same composability model already used for other nrfxlib libraries such as MPSL and the SoftDevice Controller.

>Are the files moved without modification?

The source files are relocated unchanged — no modifications to the platform implementation code. Only the location and CMake wiring differ.

BTW @nordicjm such information, in my humble opinion, should be in commit message, so that people not involved in background communication know the reason, why the change from `cmake-ext` like module to current approach, was done.

nordicjm:
```
zephyr: Move files from sdk-nrf

Moves files from sdk-nrf to here so this can be used as an actual zephyr module
```
it's short and to the point, with `move` implying that the files are not changed, otherwise it would be `update` or similar

ppryga-nordic:
> it's short and to the point, with move implying that the files are not changed, otherwise it would be update or similar

In my opinion the commit message say what was done not why it was done. Or putting it differently why the former approach was not good enough.

MarGasiorek:
Some of these files include Zephyr headers (e.g. #include <zephyr/drivers/timer/nrf_grtc_timer.h>), which makes me question whether moving them to `nrfxlib` is appropriate. I thought that the `nrfxlib` should be OS agnostic and glue layer should be placed in the `ncs`. @tejlmand, could you share your thoughts?

nordicjm:
> Some of these files include Zephyr headers (e.g. #include <zephyr/drivers/timer/nrf_grtc_timer.h>), which makes me question whether moving them to `nrfxlib` is appropriate. I thought that the `nrfxlib` should be OS agnostic and glue layer should be placed in the `ncs`. @tejlmand, could you share your thoughts?

I would say that has not been true for a long time, giving a quick grep of sdk-nrfxlib of `main` right now:
```
nrf_802154/sl/sl_opensource/src/nrf_802154_sl_timer.c:#include <zephyr/kernel.h>
lc3/src/sw_codec_lc3.c:#include <zephyr/kernel.h>
lc3/src/sw_codec_lc3.c:#include <zephyr/logging/log.h>
lc3/include/sw_codec_lc3.h:#include <zephyr/types.h>
crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_no_mutex_zephyr.c:#include <zephyr/kernel.h>
crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_mutex_zephyr.c:#include <zephyr/kernel.h>
crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_abort_zephyr.c:#include <zephyr/kernel.h>
crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_abort_zephyr.c:#include <zephyr/sys/reboot.h>
crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_abort_zephyr.c://#include <zephyr/logging/log.h>
crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_no_mutex_zephyr.c:#include <zephyr/kernel.h>
crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_mutex_zephyr.c:#include <zephyr/kernel.h>
crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_abort_zephyr.c:#include <zephyr/kernel.h>
crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_abort_zephyr.c:#include <zephyr/sys/reboot.h>
crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_abort_zephyr.c://#include <zephyr/logging/log.h>
```
And sdk-nrfxlib is a _zephyr module_ (for NCS) with code that can only be used with zephyr, it is not OS independent for a lot of it anyhow

MarGasiorek:
> > Some of these files include Zephyr headers (e.g. #include <zephyr/drivers/timer/nrf_grtc_timer.h>), which makes me question whether moving them to `nrfxlib` is appropriate. I thought that the `nrfxlib` should be OS agnostic and glue layer should be placed in the `ncs`. @tejlmand, could you share your thoughts?
> 
> I would say that has not been true for a long time, giving a quick grep of sdk-nrfxlib of `main` right now:
> 
> ```
> nrf_802154/sl/sl_opensource/src/nrf_802154_sl_timer.c:#include <zephyr/kernel.h>
> lc3/src/sw_codec_lc3.c:#include <zephyr/kernel.h>
> lc3/src/sw_codec_lc3.c:#include <zephyr/logging/log.h>
> lc3/include/sw_codec_lc3.h:#include <zephyr/types.h>
> crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_no_mutex_zephyr.c:#include <zephyr/kernel.h>
> crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_mutex_zephyr.c:#include <zephyr/kernel.h>
> crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_abort_zephyr.c:#include <zephyr/kernel.h>
> crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_abort_zephyr.c:#include <zephyr/sys/reboot.h>
> crypto/nrf_cc312_platform/src/nrf_cc3xx_platform_abort_zephyr.c://#include <zephyr/logging/log.h>
> crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_no_mutex_zephyr.c:#include <zephyr/kernel.h>
> crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_mutex_zephyr.c:#include <zephyr/kernel.h>
> crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_abort_zephyr.c:#include <zephyr/kernel.h>
> crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_abort_zephyr.c:#include <zephyr/sys/reboot.h>
> crypto/nrf_cc310_platform/src/nrf_cc3xx_platform_abort_zephyr.c://#include <zephyr/logging/log.h>
> ```
> 
> And sdk-nrfxlib is a _zephyr module_ (for NCS) with code that can only be used with zephyr, it is not OS independent for a lot of it anyhow

I'm aware of this. However I see that there are simple headers related to `kernel/system...`. This PR contains files with headers for `drivers`, so I want to get an independent opinion about this change.

zephyr/module.yml
```diff
@@ -1,5 +1,5 @@
 build:
-  cmake-ext: True
+  cmake: zephyr
```
tejlmand, line 2: why not use `cmake: .` and thereby source the root CMakeLists.txt file directly ?

That file can then source `nrf_802154/zephyr` conditionally.

tejlmand:
> Some of these files include Zephyr headers (e.g. #include <zephyr/drivers/timer/nrf_grtc_timer.h>), which makes me question whether moving them to nrfxlib is appropriate. 

@MarGasiorek it's true that one of the reasons, among many, for moving having glue code in sdk-nrf.
https://github.com/nrfconnect/sdk-nrf/pull/4146

but if some areas are having Zephyr (glue) code properly in sdk-nrf, while other areas have everything in nrfxlib, then nrfxlib itself becomes a mess and confusing to our users regarding what code belongs where.

So something has to be done, either have nrfxlib as a full self-contained Zephyr module (like this PR), or cleanup and do a proper separation (again), with the risk of ending back here again in a couple of years.

I would like @e-rk to also take a look, cause I iirc then another reason was to facilitate some pre-work / internal glue code handling for development, and would like to ensure this PR doesn't break currently supported use-cases, or that we at least continue to support those in some form.

dawidprzybylo:
For verification purpose I also checked if build is ok with `CONFIG_NRF_802154_SOURCE_INTERNAL=y`, and sample build with 802.15.4 works :). 
I have another concern here related to established workflow in my RS team -  by default all related 802.15.4 code that is put in nrfxlib is following our deploy procedure from sdk-nrf-802154 repository (CI/scripts). And moving files from nrf to nrfxlib break this requirement. So base on decision here where to maintain this code we will need to plan some work to align it or modified deploy requirements - its need refinement from our site that will land for now in our backlog.

zephyr/module.yml
nordicjm: I added it this way as I was assuming that the base directory CMakeLists.txt was meant so it was "RTOS independent", but can move the changes there if we aren't considering that anymore?

tejlmand (approved):

e-rk:
Back in the day, we wanted to avoid having NCS specific code, because the nrfxlib was meant to be OS agnostic. Other components didn't follow that guidance, but `nrf-802154` as delivered through nrfxlib is perfectly OS agnostic.

The development lifecycle of the platform integration is tied to a particular revision of `sdk-nrf`. During sdk-nrf development, upmerges, and what not, any change in nrf-802154 dependencies is contained within zephyr/sdk-nrf and do not reach nrfxlib.

MPSL and SDC also have integration code in `sdk-nrf` and the code in nrfxlib is OS agnostic (although the build system scripts aren't). Moving nrf-802154 integration code here makes things inconsistent for no benefit.

Another thing is that the content in nrf-802154 is the only source of truth for the rendered content in nrfxlib. IIRC the deployment scripts work by deleting the entire `nrf-802154` directory and re-constructing the entire directory tree. This PR will break this.

nordicjm:
> The development lifecycle of the platform integration is tied to a particular revision of sdk-nrf. During sdk-nrf development, upmerges, and what not, any change in nrf-802154 dependencies is contained within zephyr/sdk-nrf and do not reach nrfxlib.

I don't follow, you need to adapt after an upmerge and the location of the files makes no difference to that. Actually the changes should be done _during_ the upmerge, doing them after is too late.

> MPSL and SDC also have integration code in sdk-nrf and the code in nrfxlib is OS agnostic (although the build system scripts aren't). Moving nrf-802154 integration code here makes things inconsistent for no benefit.

Neither of those are zephyr modules

> Another thing is that the content in nrf-802154 is the only source of truth for the rendered content in nrfxlib. IIRC the deployment scripts work by deleting the entire nrf-802154 directory and re-constructing the entire directory tree. This PR will break this.

then update the script to delete the directory here and reconstruct it? It's the exact same contents as sdk-nrf had

MarGasiorek:
> > Another thing is that the content in nrf-802154 is the only source of truth for the rendered content in nrfxlib. IIRC the deployment scripts work by deleting the entire nrf-802154 directory and re-constructing the entire directory tree. This PR will break this.
> 
> then update the script to delete the directory here and reconstruct it? It's the exact same contents as sdk-nrf had

The files from NCS are not included in the CI scripts, hence they need to be added and the process must be verified - @dawidprzybylo correct me if I'm wrong.

nordicjm:
> > > Another thing is that the content in nrf-802154 is the only source of truth for the rendered content in nrfxlib. IIRC the deployment scripts work by deleting the entire nrf-802154 directory and re-constructing the entire directory tree. This PR will break this.
> > 
> > 
> > then update the script to delete the directory here and reconstruct it? It's the exact same contents as sdk-nrf had
> 
> The files from NCS are not included in the CI scripts, hence they need to be added and the process must be verified - @dawidprzybylo correct me if I'm wrong.

Given that this PR has been open for a month and is needed by other teams, can you give a timeframe on when that can be done?

bpienk:
> > > > Another thing is that the content in nrf-802154 is the only source of truth for the rendered content in nrfxlib. IIRC the deployment scripts work by deleting the entire nrf-802154 directory and re-constructing the entire directory tree. This PR will break this.
> > > 
> > > 
> > > then update the script to delete the directory here and reconstruct it? It's the exact same contents as sdk-nrf had
> > 
> > 
> > The files from NCS are not included in the CI scripts, hence they need to be added and the process must be verified - @dawidprzybylo correct me if I'm wrong.
> 
> Given that this PR has been open for a month and is needed by other teams, can you give a timeframe on when that can be done?

Considering the required effort, current team load and available resources, we can pick this up not earlier than early September.

MarGasiorek (dismissed):

MarGasiorek (approved):

PR #2195, opened 2026-07-15
cmake: Do path calculation in CMake as opposed to Kconfig

nordicjm:
These Kconfigs should not have been added to begin with. This is not optimal and can be improved by making a common function in a future task

manifest-pr-skip

### DIFF
--- a/softdevice_controller/CMakeLists.txt
+++ b/softdevice_controller/CMakeLists.txt
@@ -4,30 +4,71 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
-if (CONFIG_BT_LL_SOFTDEVICE_HEADERS_INCLUDE AND NOT CONFIG_BT_LL_SOFTDEVICE_HEADERS_INCLUDE_PATH_OVERRIDE)
+if(CONFIG_BT_LL_SOFTDEVICE_HEADERS_INCLUDE AND NOT CONFIG_BT_LL_SOFTDEVICE_HEADERS_INCLUDE_PATH_OVERRIDE)
   zephyr_include_directories(include)
 endif()
 
 if(CONFIG_BT_LL_SOFTDEVICE AND CONFIG_BT_LL_SOFTDEVICE_BUILD_TYPE_LIB)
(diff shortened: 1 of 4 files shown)


softdevice_controller/CMakeLists.txt
```diff
@@ -4,30 +4,69 @@
...
 
-  set(SOFTDEVICE_CONTROLLER_LIB
-    ${SOFTDEVICE_CONTROLLER_LIB_DIR}/libsoftdevice_controller_${softdevice_controller_variant}.a)
-
-  if(NOT EXISTS ${SOFTDEVICE_CONTROLLER_LIB})
-    message(FATAL_ERROR "This combination of SoC and floating point ABI is not"
-      "supported by the SoftDevice Controller lib.")
+  if(EXISTS ${lib_dir_path})
+    zephyr_link_libraries(${lib_dir_path})
+  elseif(CONFIG_SOC_SERIES_BSIM_NRFXX OR CONFIG_SOC_SERIES_NRF52 OR
```
rugeGerritsen: Use HAS_BT_LL_SOFTDEVICE_SUPPORT instead?

softdevice_controller/CMakeLists.txt
```diff
@@ -4,30 +4,69 @@
...
+    CONFIG_SOC_COMPATIBLE_NRF5340_CPUNET OR CONFIG_SOC_NRF54H20_CPURAD OR
+    CONFIG_SOC_SERIES_NRF54L OR CONFIG_SOC_SERIES_NRF71
+  )
+    message(FATAL_ERROR "This combination of SoC and floating point ABI is not "
+      "supported by the SoftDevice Controller lib."
+    )
   endif()
```
rugeGerritsen, line 73: Make error condition more verbose?
```suggestion
  else()
     message(FATAL_ERROR "Missing library file")
  endif()
```

nordicjm: done

rugeGerritsen (approved):
Since the commit message mentions "These Kconfigs should not have been added to begin with", it would be nice if you could point towards where a style guide or similar mentions this

nordicjm:
> Since the commit message mentions "These Kconfigs should not have been added to begin with", it would be nice if you could point towards where a style guide or similar mentions this

They are not configurable, so there was no reason to put them into Kconfig, why would a user change this (they can't but the point still stands)

carlescufi (approved):

PR #2231, opened 2026-08-12
mpsl: enable legacy implementation of clock drivers

nika-nordic:
Enable clock drivers legacy implementation for MPSL. To be removed once MPSL becomes aligned to new clock scheme.

manifest-pr-skip


nika-nordic:
closing as analogous PR was already merged

PR #2247, opened 2026-08-21
nrf_rpc: fix race condition in disable/enable functionality

MarekPorwisz:
A solution was implemented, to complete all the handlers first, prevent registering handlers when disabled, avoid enabling during disabling, etc.

### DIFF
--- a/nrf_rpc/include/nrf_rpc.h
+++ b/nrf_rpc/include/nrf_rpc.h
@@ -447,20 +447,33 @@ void nrf_rpc_register_cleanup_handler(struct nrf_rpc_cleanup_handler *handler);
  * Calling this function automatically fails all communication until @ref nrf_rpc_resume is called.
  * It also causes all pending commands to fail immediately.
  *
+ * @warning This function is experimental and subject to change or removal without prior notice.
+ *
  * @param cleanup set to true to also invoke all the custom cleanup handlers.
  *
  * @note If the cleanup parameter is set to true it may be needed to reset the peer before invoking
  *       @ref nrf_rpc_resume.
  *
+ * @retval 0             on success.
+ * @retval -NRF_EALREADY when RPC is already stopped or stopping.
+ * @retval -NRF_ETIMEDOUT when waiting for in-flight RPC processing to complete times out
+ *                        and assertions are disabled.
+ *
  */
-void nrf_rpc_stop(bool cleanup);
+int nrf_rpc_stop(bool cleanup);
 
 /** @brief resumes RPC communication
  *
  * Calling this function reverts the effects of @ref nrf_rpc_stop.
  *
+ * @warning This function is experimental and subject to change or removal without prior notice.
+ *
+ * @retval 0             on success.
+ * @retval -NRF_EINPROGRESS when RPC is currently stopping.
+ * @retval -NRF_EALREADY when RPC is already running.
+ *
  */
-void nrf_rpc_resume(void);
+int nrf_rpc_resume(void);
 
 
 /** @brief Send a command and provide callback to handle response.
--- a/nrf_rpc/include/nrf_rpc.h
+++ b/nrf_rpc/include/nrf_rpc.h
@@ -452,15 +452,24 @@ void nrf_rpc_register_cleanup_handler(struct nrf_rpc_cleanup_handler *handler);
  * @note If the cleanup parameter is set to true it may be needed to reset the peer before invoking
  *       @ref nrf_rpc_resume.
  *
+ * @retval 0             on success.
+ * @retval -NRF_EALREADY when RPC is already stopped or stopping.
+ * @retval -NRF_ETIMEDOUT when waiting for in-flight RPC processing to complete times out
+ *                        and assertions are disabled.
+ *
  */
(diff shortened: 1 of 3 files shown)


nrf_rpc/include/nrf_rpc.h
```diff
@@ -452,15 +452,24 @@ void nrf_rpc_register_cleanup_handler(struct nrf_rpc_cleanup_handler *handler);
  * @note If the cleanup parameter is set to true it may be needed to reset the peer before invoking
  *       @ref nrf_rpc_resume.
  *
+ * @retval 0             on success.
```
Damian-Nordic, line 457: This API is not documented as experimental so in theory its signature shouldn't be changed without deprecation period. Could you at least document it as experimental now, hoping nobody has started using it?

nrf_rpc/nrf_rpc.c
```diff
@@ -129,7 +130,20 @@ static uint8_t initialized_group_count;
...
+	NRF_RPC_STATE_RUNNING,
+	NRF_RPC_STATE_STOPPING,
+	NRF_RPC_STATE_STOPPED,
+};
+static enum nrf_rpc_runtime_state rpc_runtime_state = NRF_RPC_STATE_RUNNING;
+
+static bool nrf_rpc_record_processing_start(void);
+static void nrf_rpc_record_processing_end(void);
+
+static bool nrf_rpc_state_is_stopped(void)
```
Damian-Nordic: It's a bit confusing that "stopped" returns true also for "stopping" case. Why not implement "is_running" instead?

nrf_rpc/nrf_rpc.c
```diff
@@ -1301,30 +1345,121 @@ void nrf_rpc_register_cleanup_handler(struct nrf_rpc_cleanup_handler *handler)
...
+}
+
+static bool nrf_rpc_record_processing_start(void)
+{
+	bool stopped;
+
+	nrf_rpc_os_mutex_lock(&rpc_state_mutex);
+	assert(processing_in_progress_count < UINT8_MAX);
+
+	nrf_rpc_os_event_wait(&processing_complete_event,
```
Damian-Nordic, line 1425: What's the reason for this?

nrf_rpc/nrf_rpc.c
```diff
@@ -1301,30 +1345,121 @@ void nrf_rpc_register_cleanup_handler(struct nrf_rpc_cleanup_handler *handler)
 		cleanup_handlers = handler;
 	}
 
-	nrf_rpc_os_mutex_unlock(&cleanup_mutex);
+	nrf_rpc_os_mutex_unlock(&rpc_state_mutex);
+}
+
+static bool nrf_rpc_record_processing_start(void)
```
Damian-Nordic: It's kind of counterintuitive that "true" means "failure" (that processing shouldn't start). It would re-think the return value.

nrf_rpc/nrf_rpc.c
MarekPorwisz: when the processing starts, the event needs to be consumed in order from preventing immediate RPC stop operation. The event is consumed only by the first handler and does not block remaining ones - thus no timeout. Event is set back again when the last handler completes (processing_in_progress_count is 0). 
the original approach was trying to do this in parallel, unfortunately some cases were causing issues so we decided to handle "stop" operation synchronously i.e prevent new handlers from spawning, wait (with timeout) for all pending handlers to complete and do the stop operation after.

nrf_rpc/nrf_rpc.c
```diff
@@ -1301,30 +1349,121 @@ void nrf_rpc_register_cleanup_handler(struct nrf_rpc_cleanup_handler *handler)
...
+int nrf_rpc_stop(bool cleanup)
+{
+	int err;
+	bool requested = nrf_rpc_request_completion();
+	if (requested) {
+		err = nrf_rpc_os_event_wait(&processing_complete_event, CONFIG_NRF_RPC_STOP_WAIT_TIME);
+		if (err) {
+			assert(err != -NRF_ETIMEDOUT);
+			nrf_rpc_os_mutex_lock(&rpc_state_mutex);
+			rpc_runtime_state = NRF_RPC_STATE_RUNNING;
```
kderda, line 1496: What will happen when asserts are disabled in production environment and a timeout occurs? Are there any side effects of such scenario like freed buffers, context, etc.?

MarekPorwisz: Side effects in that timeout path:
No stop finalization happens (nrf_rpc_finalize_stop() is not called)
so no abort_all_ops(), no cleanup-handler execution, no forced teardown
no extra/freeing of TX buffers/contexts is done by this branch itself
command contexts and buffers stay under normal ownership/lifecycle
So memory-safety-wise (double-free / premature free), this path is fine.

Practical/runtime consequence
During the temporary STOPPING window before timeout, packet gating can drop traffic, so some in-flight RPC activity may fail/timeout. After rollback to RUNNING, traffic resumes, but those dropped exchanges are already lost.

kderda (approved):

Damian-Nordic (approved):

kenr (approved):

PR #2249, opened 2026-08-24
nrf_rpc: reset init event before wait

MarekPorwisz:
It is possible for the init event to remain set after earlier bind call. To address this, event is reset during before bind.

### DIFF
--- a/nrf_rpc/nrf_rpc.c
+++ b/nrf_rpc/nrf_rpc.c
@@ -1216,6 +1216,9 @@ int nrf_rpc_bind(void)
 	void *iter;
 	const struct nrf_rpc_group *group;
 
+	/* Clear any stale signal from a previous bind cycle before waiting again. */
+	nrf_rpc_os_event_reset(&groups_init_event);
+
 	for (NRF_RPC_AUTO_ARR_FOR(iter, group, &nrf_rpc_groups_array, const struct nrf_rpc_group)) {
 		const struct nrf_rpc_tr *transport = group->transport;
 		struct nrf_rpc_group_data *data = group->data;
@@ -1261,6 +1264,7 @@ void nrf_rpc_unbind(void)
 	const struct nrf_rpc_group *group;
 
 	initialized_group_count = 0;
+	nrf_rpc_os_event_reset(&groups_init_event);
 
 	for (NRF_RPC_AUTO_ARR_FOR(iter, group, &nrf_rpc_groups_array, const struct nrf_rpc_group)) {
 		group->data->dst_group_id = NRF_RPC_ID_UNKNOWN;


nrf_rpc/nrf_rpc.c
```diff
@@ -1216,6 +1216,9 @@ int nrf_rpc_bind(void)
 	void *iter;
 	const struct nrf_rpc_group *group;
 
+	/* Clear any stale signal from a previous bind cycle before waiting again. */
+	nrf_rpc_os_event_reset(&groups_init_event);
```
Damian-Nordic, line 1220: Isn't the change in `nrf_rpc_unbind` enough? It seems resetting the event without resetting `initialized_group_count` may lead to other unexpected behavior so I think `nrf_rpc_unbind` should be mandatory if re-bind is needed.

Damian-Nordic (approved):

nrf_rpc/nrf_rpc.c
MarekPorwisz: This fix is needed for the case with late event arrival. Event may be set after nrf_rpc_os_event_wait has already failed.  another call to bind would continue without waiting in such case.

Damian-Nordic: Yeah, but my question is if the API doesn't require that you call `nrf_rpc_unbind` in such a case before `nrf_rpc_bind`. I mean, if you reset the event without resetting the `initialized_group_count`, does it mean that you allow a user to call `nrf_rpc_bind` multiple times, incrementally, until all groups are bound? I think this needlessly complicates the implementation. If not, and `nrf_rpc_unbind` is always necessary before retrying `nrf_rpc_bind`, then clearing this event should already be done in the former, no?

MarekPorwisz: TBH I do not quite get the concept of this API... If I am not mistaken, it currently expects to call this multiple time. clearing the event before waiting on it seems natural to me, however it seems this API despite reporting bind failure, leaves the the device in some intermediate state. I would not expect the need to call unbind in case the bind has failed. I think we need to discuss this in person and decide the best approach. Please let me know when you are back at the office.

kderda (approved):

kderda:
Isn't there a chance for two scenarios to happen?

1. nrf_rpc_unbind() is called -> Initalization packet arrives from the other party -> nrf_rpc_bind() resets initialization event -> false timeout
2. first bind() succeeds -> second bind can be called as there is no hard requirement to unbind first -> first_init stays false

MarekPorwisz:
> Isn't there a chance for two scenarios to happen?
> 
> 1. nrf_rpc_unbind() is called -> Initalization packet arrives from the other party -> nrf_rpc_bind() resets initialization event -> false timeout
> 2. first bind() succeeds -> second bind can be called as there is no hard requirement to unbind first -> first_init stays false

calling new bind now unbinds all bound

Damian-Nordic (approved):

kderda (approved):

